PR #29, opened 2017-07-31
Make sure we've got a proper IP address from DO

onorua:
In some cases DigitalOcean doesn't provide valid values, which had broken our automation several time. This happen really rare, mostly when DO guys do some updates to API, that is why it is really hard to reproduce.

### DIFF
--- a/digitalocean/resource_digitalocean_droplet.go
+++ b/digitalocean/resource_digitalocean_droplet.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net"
 	"strconv"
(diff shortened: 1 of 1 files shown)


radeksimko:
This bugfix looks sensible to me, but do you mind sharing some details about what exactly happened when it broke? What value did you get back from the API?

onorua:
I do not have exact message at the moment, but it looks like this:
```
<xml blalbabldfsf > more blalslls </xml> 
```
instead of IP address.

radeksimko:
I see, thanks, that's helpful.

PR #37, opened 2017-08-31
Added importability

jonnydford:


jamietanna:
Watching this with intent as it's currently making it difficult to bring over a number of domains + DNS without tearing it down and building it.

jonnydford:
@radeksimko This should be all of the changes made now.

jonnydford:
@radeksimko Can we get this merged please?

eddiezane:
@danrabinowitz to review

andrewsomething:
Hi @jonnydford. Thanks again for the work you put in here. I'm sorry that it has taken so long to get this reviewed and merged.

Reviewing this PR, there were still a number of issues. None of the tests added were passing. Since the review cycle has been so long already, rather than pass this back to you, I've resolved the issues for both the Certificate and Load Balancer issues in #104. Another PR (#64) has already been opened for the Record imports that is a bit further along than the work here.

You can see the additional changes I needed to make in these commits:

* The DigitalOcean API was rejecting the certificates that you included as invalid. Rather than including the certificates as test fixtures, I have refactored the code already in `TestAccDigitalOceanCertificate_Basic` to make it reusable here. See: https://github.com/terraform-providers/terraform-provider-digitalocean/pull/104/commits/2bc90c62a579feb7c2fd4ec489fd01f5d91c0b4f

* The Load Balancer import test was failing as the `droplet_ids` were not being correctly imported. This exposed a few small bugs in the existing Load Balancer resource. We did not check for an error when calling `flattenDropletIds`. If we had, we'd see that the IDs needed to be converted to strings. Instead, we silently ignored them. See: https://github.com/terraform-providers/terraform-provider-digitalocean/pull/104/commits/e315cdbacd4be08a0318b6e0535643f6192eaa37

Like I said, there is already another PR for Record imports (#64). Here the test initially fails as `testAccCheckDigitalOceanRecordConfig_basic` is set for the config but that requires a domain name like so:

```
fmt.Sprintf(testAccCheckDigitalOceanRecordConfig_basic, domain),
```

Test result:

```
$ make testacc TESTARGS='-run=TestAccDigitalOceanRecord_importBasic'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccDigitalOceanRecord_importBasic -timeout 120m
=== RUN   TestAccDigitalOceanRecord_importBasic
--- FAIL: TestAccDigitalOceanRecord_importBasic (0.97s)
	testing.go:427: Step 0 error: Error applying: 1 error(s) occurred:
		
		* digitalocean_domain.foobar: 1 error(s) occurred:
		
		* digitalocean_domain.foobar: Error creating Domain: POST https://api.digitalocean.com/v2/domains: 422 Name can not be just a TLD.
FAIL
FAIL	github.com/terraform-providers/terraform-provider-digitalocean/digitalocean	0.978s
GNUmakefile:15: recipe for target 'testacc' failed
make: *** [testacc] Error 1
```

Even after fixing that, there is still an issue with needing to pass the domain in as an argument for the import as well. See #64 for one approach to that.

Thanks again for helping to improve this Terraform provider!

PR #38, opened 2017-08-31
Added monitoring boolean

jonnydford:
Added for enhancement: https://github.com/terraform-providers/terraform-provider-digitalocean/issues/12


lfarnell:
@jonnydford I would suggest moving over the code Josh wrote on https://github.com/hashicorp/terraform/pull/13494. Also, I would read the comments on that PR to handle some of the concerns not addressed yet.

PR #43, opened 2017-09-23
Configure Terraform version as part of the godo User-Agent

joonas:
This configures the calling Terraform version as part of the `godo` client so that we know when calls are coming from Terraform and which version of Terraform.

cc @radeksimko

### DIFF
--- a/digitalocean/config.go
+++ b/digitalocean/config.go
@@ -2,6 +2,7 @@ package digitalocean
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"net/http"
 	"net/http/httputil"
@@ -10,6 +11,7 @@ import (
 	"github.com/digitalocean/godo"
 	"github.com/hashicorp/terraform/helper/logging"
 	"github.com/hashicorp/terraform/helper/resource"
+	"github.com/hashicorp/terraform/terraform"
 	"golang.org/x/oauth2"
 )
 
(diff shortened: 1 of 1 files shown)


joonas:
> Just FYI digitalocean/godo#151 will require version bump of the vendored godo - probably best after you tag new release, so we're not pinning to a hash?

Yup, I was planning on doing a new release once I have a chance to check internally with folks to make sure nothing else should be included.

> Also the terraform.VersionString() will return version of vendored Terraform (0.10 at this point). It is a wide-spread issue across all providers we plan to address eventually, so that each provider can report both provider and core version (ideally the one talking to provider over RPC, instead of the vendored one which is used in helper/*).

I was wondering about that, it seemed like the `hashicorp/terraform` repo had a few mentions of this in various issues, do you know of any concrete plans to address this or would it be a potentially area for contributions?

> There is nothing to do on the provider side at this point - just be aware of this before you start drawing any conclusions from reported versions.

Not a problem, knowing the exact version of the provider is definitely something we'd like to have, but for now it's sufficient to know that requests are coming from Terraform at least.

Thanks!

radeksimko:
> do you know of any concrete plans to address this or would it be a potentially area for contributions?

Not in the nearest term, but I think the core team has some plans to address this along with making the experience of using 3rd party providers (outside of releases.hashicorp.com) more seamless.
It will require some changes in both `hashicorp/terraform` (core) and on the provider side.

I think we would be willing to review contributions in that area, but be aware that it _may_ turn out to be much bigger job than it seems to be 😄

PR #49, opened 2017-10-05
Correct the logic for resizing a droplet. Fixes 15

danrabinowitz:
I'm new to this codebase, and fairly new to golang, so please forgive my ignorance.

This issue https://github.com/terraform-providers/terraform-provider-digitalocean/issues/15 describes a bug. 

The current/old logic says to power off and resize iff `(the size changes OR the resize disk field changes) AND resizeDisk is true`. That's logic does not seem correct to me. Instead we should power off and resize iff `resizeDisk is true AND d.HasChange("size")`.


captn3m0:
Faced this issue issue while applying no-op on a imported resource.

PR #55, opened 2017-11-13
Handle missing floating IPs

aknuds1:
Handle updating state wrt. floating IPs that turn out to be missing. This fixes a bug for me with `terraform refresh` when I've deleted a floating IP manually since last applying.

Please let me know if I should add a test! I'm new to both Terraform plugin development and Go...


eddiezane:
@danrabinowitz to review

PR #57, opened 2017-11-13
Handle denial to unassign floating IP from droplet

aknuds1:
When trying to unassign a floating IP, the DigitalOcean API may return a status code 422 meaning that the request was denied. In my experience this can happen because the droplet in question has been deleted. I make the change here to handle these failures gracefully, as it shouldn't abort Terraform.

I've verified through manual testing that this change solves the issue I was experiencing, i.e. that during `terraform destroy`, the floating IP deletion would abort the whole process since it couldn't be unassigned from the droplet that was deleted prior.

I would appreciate if someone explained to me how to write a test for this case :)


eddiezane:
@danrabinowitz to review

andrewsomething:
Thanks for the PR! The DO API does indeed return a 422 if you attempt to unassign a Floating IP that is not assigned to a Droplet. So in the case where the Droplet has already been destroyed, we can safely ignore that error. In #109 I added some additional testing for reassigning and unassigning.

PR #63, opened 2018-01-27
Adding digitalocean_domain data source

slapula:
Hello,  not sure if others would find this useful but I've cobbled together a data source for domain names managed by a Digital Ocean account.  It seems to work well for me and I've done my best to bring it up to a point where it could be shared with others.  I'm no go programmer so any feedback would be much appreciated.  I ended up using both the `digitalocean_image` data source and the `digitalocean_domain` resource types as inspiration.  Thanks in advance.

### DIFF
--- a/digitalocean/datasource_digitalocean_domain.go
+++ b/digitalocean/datasource_digitalocean_domain.go
@@ -1,0 +1,74 @@
+package digitalocean
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/digitalocean/godo"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceDigitalOceanDomain() *schema.Resource {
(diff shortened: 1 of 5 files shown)


tritao:
This seems useful, especially as workaround for https://github.com/terraform-providers/terraform-provider-digitalocean/issues/53

What is needed to merge it?

andrewsomething:
Thanks for the PR @slapula! Sorry again for how long it's take to get these reviewed. There's one change additional change that I had to make before merging. The DO endpoints that return lists are paginated. So you must iterate over them if there are more resources than the default number returned. Check out https://github.com/terraform-providers/terraform-provider-digitalocean/pull/106/commits/2d817a423f3a9db1b4089663ced4c272092fbbe1 to see how I approached that.

PR #64, opened 2018-01-27
Adding digitalocean_record data source

slapula:
Similar to #63,  I've cobbled together another data source for DNS records managed by a Digital Ocean account. It also seems to work well for me. I'm no go programmer so any feedback would be much appreciated. I ended up using both the digitalocean_image data source and the digitalocean_record resource types as inspiration. Thanks in advance.

### DIFF
--- a/digitalocean/datasource_digitalocean_record.go
+++ b/digitalocean/datasource_digitalocean_record.go
@@ -1,0 +1,120 @@
+package digitalocean
+
+import (
+	"context"
+	"fmt"
+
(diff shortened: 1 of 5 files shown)


andrewsomething:
Merged via #107 The main change was that the name was being set as the ID and a few attributes we not being set at all (e.g. tag and flags). See https://github.com/terraform-providers/terraform-provider-digitalocean/pull/107/commits/3f135ddd8bb8df267ca5ad36755108c7b4246787

Thanks again for all these additions!

PR #66, opened 2018-02-02
Adding bucket resource for Digital Ocean Spaces

slapula:
This is my stab at adding a bucket resource for Digital Ocean Spaces.  Any feedback would be much appreciated.  One thing I'd like to do here that's missing is moving the Spaces access id and secret key to the provider level and make it an optional credential.  I'm not sure how to do that though since I'm a go newb.


slapula:
I suppose I should add my dependency in a different PR?  All of my tests are passing locally so Im not sure what else would be causing the Travis error.

lfarnell:
@slapula Thanks for the great work! The main concern I have with this PR is that I don't know if the Minio SDK is the way to go. For instance, the DO Spaces API allows you to enable bucket versioning, the Minio SDK does not have the ability to do this as Minio has said this is not a focus for the project. So as much as its a pain sometime to use sometimes I think the AWS SDK would be a better option. I'm open to anyone who has different thoughts on this.

slapula:
@lfarnell That makes sense.  I think I'll start working on a branch that uses the AWS SDK instead.  I just went with Minio because that's what I am currently comfortable with (in my limited golang experience).  This will be good practice for me regardless of which way we decide to go here.

slapula:
I submitted a new PR using the AWS Go SDK #77.  I think I prefer that one over this implementation.  I'm going to stew on it for a bit then I will likely close this.

PR #84, opened 2018-04-13
Add an override for the DigitalOcean API URL

protochron:
This allows you to change the default URL for accessing the DigitalOcean
API. It's similar to what you would do when overriding an endpoint in
the AWS provider, only with the DO API it's an all or nothing change.


andrewsomething:
@protochron We'll be looking to cut a new release of the provider sometime in the very near future. It would be great to land this. In addition to the docs update, it would be good to have a test that the default works as expected.

PR #119, opened 2018-08-28
Updated the terraform provider SDK with supporting dependencies and aligned the code base

TFaga:
This pull request brings the terraform provider SDK to its latest version in order to support terraforms latest features.  In addition, the PR brings in the latest version of the DigitalOcean SDK and other required updates to the supporting dependencies.

Due to the upgrades in the dependencies, the minimum `go` version required to build the provides is now `1.9` or above.


README.md
```diff
@@ -11,7 +11,7 @@ Requirements
 ------------
 
 -	[Terraform](https://www.terraform.io/downloads.html) 0.10.x
```
andrewsomething: Should this now be 0.11.x?

TFaga: We can leave it at 0.10.x. I've ran the acceptance tests with terraform 0.10.1 and it produced the same results as with 0.11. There should be no breaking changes between the terraform provider SDK versions.

andrewsomething:
:shipit:

PR #120, opened 2018-08-28
Migrated tags to use sets and centralised tag definitions/functionality

### LINKED ISSUE #7: Tags for digital ocean droplets keep changing order
_This issue was originally opened by @simaofreitas as hashicorp/terraform#9099. It was migrated here as part of the [provider split](https://www.hashicorp.com/blog/upcoming-provider-changes-in-terraform-0-10/). The original body of the issue is below._

<hr>

Hi,

I'm using terraform for managing a set of droplets in digital ocean.
Yesterday I introduced tags (using `digitalocean_tag`) and everything is working correctly.
The problem is that every time I run `terraform plan` (or `apply`) I get changes because the `tags` keep changing their order inside the `tfstate` file.

Terraform version: `0.7.3`

Example of tags definition:

```
resource "digitalocean_tag" "staging" {
    name = "staging"
}

resource "digitalocean_tag" "web" {
    name = "web"
}
```

Example of droplet definition

```
resource "digitalocean_droplet" "machine01" {
  image              = "ubuntu-14-04-x64"
  name               = "machine01"
  region             = "fra1"
  size               = "16gb"
  private_networking = true
  backups            = true
  tags               = ["${digitalocean_tag.staging.id}", "${digitalocean_tag.web.id}"]

  ssh_keys = [
     ...
  ]
}
```

When running `terraform plan` (after I already ran `terraform apply`) I get this:

```
~ digitalocean_droplet.machine01
    tags.0: "staging" => "web"
    tags.1: "web" => "staging"
```

If I `apply`, the next time I run `plan` I would get the reverse output. But it always changes.

Couldn't find any open issues about this.

Any ideas?

Thanks a lot

TFaga:
Moved tags to a separate space/files, so that they can be managed in a single place and used by any resource that can be tagged.

Changed the schema and definition of the tags from a list to a set, to resolve outstanding issues with changing orders and duplicates. Fixes #7 

The change is backward compatible. If you have an existing state file, the tags inside it will be automatically updated/refreshed on the next read.

### DIFF
--- a/digitalocean/resource_digitalocean_droplet.go
+++ b/digitalocean/resource_digitalocean_droplet.go
@@ -322,7 +322,9 @@ func resourceDigitalOceanDropletRead(d *schema.ResourceData, meta interface{}) e
 		"host": findIPv4AddrByType(droplet, "public"),
 	})
 
-	d.Set("tags", flattenTags(droplet.Tags))
+	if err := d.Set("tags", flattenTags(droplet.Tags)); err != nil {
+		return fmt.Errorf("Error setting `tags`: %+v", err)
+	}
 
 	return nil
 }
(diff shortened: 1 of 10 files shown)


andrewsomething:
:shipit:

PR #127, opened 2018-08-30
Added validations with minor optimizations to the load balancer resource

TFaga:
Added some critical validation to the load balancer properties. In addition, updated the documentation on how to support in-place load balancer certificate renewals and updates with the same existing resources.


andrewsomething:
:ship: Good stuff. Even found a few constraints that we're missing in the DO API documentation. Will get those added there as well.

PR #128, opened 2018-08-30
Updated the state representation of the `user_data` droplet property to a hashed one

### LINKED ISSUE #65: digitalocean_droplet resource should probably use hash for user_data
When using `plan` or `apply`, the `aws_instance` resource uses the hash of the `user_data` attribute. Instead the `digitalocean_droplet` resources displays the value itself. The value can become very and might contain sensitive data. 

Compare [resource_digitalocean_droplet.go](https://github.com/terraform-providers/terraform-provider-digitalocean/blob/310d4ec82ea7bda0315ae56788b0465db6fd879c/digitalocean/resource_digitalocean_droplet.go#L144) and [resource_aws_instance.go](https://github.com/terraform-providers/terraform-provider-aws/blob/b82247b1d1270097aaf1e746749d958525bbdff5/aws/resource_aws_instance.go#L106).

Unfortunately, I know too little Go to create a pull request myself.

TFaga:
The `user_data` property of the `digitalocean_droplet` resource was updated to use the hash of the value to save in the state and use for diffs instead of the full value, which can be quite large. This way the actual value won't be shown in the console when doing diffs. This increases visibility and clarity of the output, as well as prevents potential sensitive data from being shown.

Backward compatibility for existing state files with full `user_data` values saved is preserved.

In addition, the PR includes validations for the droplet properties and minor fixes and enhancements to the codebase of the resource.

Resolves #65

### DIFF
--- a/digitalocean/hash.go
+++ b/digitalocean/hash.go
@@ -1,0 +1,11 @@
+package digitalocean
+
+import (
+	"crypto/sha1"
+	"encoding/hex"
+)
+
+func HashString(s string) string {
+	hash := sha1.Sum([]byte(s))
+	return hex.EncodeToString(hash[:])
+}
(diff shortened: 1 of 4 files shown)


andrewsomething:
:shipit:

PR #129, opened 2018-08-30
Added support for LetsEncrypt issued certificates

### LINKED ISSUE #93: Free certificates with LetsEncrypt
The DO API now provides a way to automatically create certificates with LetsEncrypt which can be used in load balancers.

All that needs to be provided is `type: "lets_encrypt"` instead of `private_key`, `leaf_certificate` and `certificate_chain`. The change does not look very complicated and I was wondering if there are any plans in the near term to implement this. If not, I could have a stab at it.

https://developers.digitalocean.com/documentation/v2/#certificates

TFaga:
Resolves #93

### DIFF
--- a/digitalocean/resource_digitalocean_certificate.go
+++ b/digitalocean/resource_digitalocean_certificate.go
@@ -153,7 +153,7 @@ func resourceDigitalOceanCertificateCreate(d *schema.ResourceData, meta interfac
 
 	d.SetId(cert.ID)
 
-	log.Printf("[INFO] Waiting for certificate (%s) to have state 'pending", cert.ID)
+	log.Printf("[INFO] Waiting for certificate (%s) to have state 'verified'", cert.ID)
 	stateConf := &resource.StateChangeConf{
 		Pending:    []string{"pending"},
 		Target:     []string{"verified"},
--- a/digitalocean/resource_digitalocean_certificate.go
+++ b/digitalocean/resource_digitalocean_certificate.go
@@ -4,9 +4,12 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"time"
 
 	"github.com/digitalocean/godo"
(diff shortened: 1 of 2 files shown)


andrewsomething:
@TFaga It would be great to get some acceptance tests added here.

digitalocean/resource_digitalocean_certificate.go
```diff
@@ -85,22 +153,49 @@ func resourceDigitalOceanCertificateCreate(d *schema.ResourceData, meta interfac
 
 	d.SetId(cert.ID)
 
+	log.Printf("[INFO] Waiting for certificate (%s) to have state 'pending", cert.ID)
```
andrewsomething: `'verified'`?

TFaga:
@andrewsomething Agreed, but the problem I encountered was that you need a real domain managed in your account otherwise the API will return an error that the domain is not managed by DO. Is there a way around that restriction?

andrewsomething:
@TFaga - Ah, I see. We could work around that error in the test by just creating a domain on DO with a random name. The certificate resource would be created, but it end up in an `errored` state rather than `verified` as that is dependent on the response from Let's Encrypt.

TFaga:
@andrewsomething Sadly it doesn't get to that stage. When creating the certificate, the API returns the following:

```
Status: 422 Unprocessable Entity
{
    "id": "unprocessable_entity",
    "message": "one or more DigitalOcean Name Server not found for randomdomain.com domain.",
    "request_id": "5a609b68-3602-4a34-97c1-18b4f73da07d"
}
```

And as such is not created at all.

digitalocean/resource_digitalocean_certificate.go
TFaga: Updated.

andrewsomething:
Thanks for digging into it more.

PR #131, opened 2018-08-31
Fixed an issue with force recreating droplets when using an older terraform state

TFaga:
Fixes a force recreate of a droplet If you have a state file from an older version of the terraform provider and didn't specify the `monitoring` or `backups` property in the resource.


TFaga:
On second look, this might be more clean by doing a state migration. I'll update the PR.

TFaga:
Updated

andrewsomething:
:+1:

PR #135, opened 2018-09-05
Added additional tests for the load balancer resource

TFaga:
The PR adds additional tests for the load balancer resources to cover the default values for the `healthcheck` and `sticky_sessions` block.


andrewsomething:
:+1:

PR #136, opened 2018-09-05
Refactored support for setting the volumes filesystem type and label

TFaga:
The PR splits out the volume properties for the filesystem type/label that the volume is initially created with and the volumes current filesystem type/label. The reason behind the change is because the filesystem can not be changed through the API after the fact. However if the filesystem is changed by the droplet, the changes will be reflected in the API and consequentially the terraform state file on refresh.

The changes should be fully backward compatibly with a deprecation message for using the old property.


andrewsomething:
:+1:

PR #21, opened 2017-07-04
Add Snapshot datasource.

andrewsomething:
Finally got back around to this...

This adds a datasource for DigitalOcean snapshots similar to the aws_ami datasource. It allows for referencing user created snapshots when creating a new Droplet or volume.

```
$ make testacc TEST=./digitalocean TESTARGS='-run=TestAccDigitalOceanSnapshotDataSource*'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./digitalocean -v -run=TestAccDigitalOceanSnapshotDataSource* -timeout 120m
=== RUN   TestAccDigitalOceanSnapshotDataSource_droplet
--- PASS: TestAccDigitalOceanSnapshotDataSource_droplet (240.15s)
=== RUN   TestAccDigitalOceanSnapshotDataSource_volume
--- PASS: TestAccDigitalOceanSnapshotDataSource_volume (30.32s)
PASS
ok  	github.com/terraform-providers/terraform-provider-digitalocean/digitalocean	270.48
```

PR #139, opened 2018-09-07
Add Volume Snapshot resource and datasource

TFaga:
This PR is a continuation of the PR #21 . I've made a few changes, one of which was narrowing the focus of both the snapshot datasource and the resource in order to simplify the scope and terminology of the resource in the API.

There is significant overlap between images and snapshots that are created from droplets, in the sense that they are currently one and the same. In addition, I didn't find a way to reliably create a Droplet snapshot and get its ID through the API. As well as once you can get a reference to a droplet snapshot, there is nothing you can do with it that is not already handled by the image datasource. You can only use it as an `image` rather than a `snapshot`.

For this reason it might make more sense to handle the droplet snapshots as images through the `digitalocean_image` datasource and potentially later an `image` resource if possible.

This way the `digitalocean_snapshot` resource/datasource would represent explicitly a volume snapshot, which can be comfortably managed through the API and also be used to create a new volume from it. Lumping in droplet snapshot would make the snapshot datasource more complex, since you can't use them for anything and can already get them using the `image` datasource. 

That said if they are any plans in the future for converging or otherwise changing the functionality of the different types of snapshots, that might be worth considering now as well.

What are your thoughts on this?


TFaga:
@andrewsomething Alright, sounds good. This way there can potentially be different kind of snapshot in the future if needed. And it makes explicit and straightforward.

I'll update the PR.

Issue #92, opened 2018-05-30
Resource 'digitalocean_droplet.droplet' does not have attribute 'ipv4_address_private'


mr-yaky:
Hi,

I try use droplet module from github:
https://github.com/antarctica/terraform-module-digital-ocean-droplet

And always get an error about unset attribute.
I know that I can use my own custom module and just comment/remove following lines:
```
output "ip_v4_address_private" {
    value = "${digitalocean_droplet.droplet.ipv4_address_private}"
}
```
But I think it will be great to fix it on provider level because it will be useful for dynamic resource's configurations for using with modules even if `private_networking` is `false`.
Also may be you know some other nice workaround to solve this problem while this will be fixed on provider level ? 

### Terraform Version
Terraform v0.11.7
+ provider.digitalocean v0.1.3

### Affected Resource(s)
- digitalocean_droplet

### Terraform Configuration Files
```hcl
terraform {
  required_version = ">= 0.11.7"
}

provider "digitalocean" {
    version = "~> 0.1.3"
    token   = "${var.digitalocean_token}"
}

variable "digitalocean_token" {
  type = "string"
  default = ""
  description = "DigitalOcean API Token"
}

variable "ssh_fingerprints" {
    default = [
        "blah"
    ]
}

module "vm01" {
    source = "github.com/antarctica/terraform-module-digital-ocean-droplet?ref=v1.2.0"
#    source = "./droplet"
    hostname = "vm01"
    image = "centos-7-x64"
    region = "lon1"
    size = "1gb"
    private_networking = "false"
    ssh_fingerprints = "${var.ssh_fingerprints}"
}

```

### Error Output
```
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

digitalocean_droplet.droplet: Refreshing state... (ID: 12345678)

------------------------------------------------------------------------

Error: Error running plan: 1 error(s) occurred:

* module.pgsql01.output.ip_v4_address_private: Resource 'digitalocean_droplet.droplet' does not have attribute 'ipv4_address_private' for variable 'digitalocean_droplet.droplet.ipv4_address_private'
```

### Expected Behavior
ipv4_address_private should be set and should be a null even if `private_networking` is `false`.

### Actual Behavior
always get an error: 
```
does not have attribute 'ipv4_address_private' for variable 'digitalocean_droplet.droplet.ipv4_address_private'
```
[...]

PR #140, opened 2018-09-07
Ensure tests are consistent with the current status of the resources/datasources

TFaga:
This PR includes minor fixes to ensure all the acceptance tests pass.

### DIFF
--- a/digitalocean/resource_digitalocean_droplet.go
+++ b/digitalocean/resource_digitalocean_droplet.go
@@ -321,21 +321,10 @@ func resourceDigitalOceanDropletRead(d *schema.ResourceData, meta interface{}) e
 	d.Set("status", droplet.Status)
 	d.Set("locked", droplet.Locked)
 
-	if publicIPv4 := findIPv4AddrByType(droplet, "public"); publicIPv4 != "" {
-		d.Set("ipv4_address", publicIPv4)
-	}
-
-	if privateIPv4 := findIPv4AddrByType(droplet, "private"); privateIPv4 != "" {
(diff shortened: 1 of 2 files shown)


digitalocean/resource_digitalocean_droplet.go
```diff
@@ -321,27 +321,35 @@ func resourceDigitalOceanDropletRead(d *schema.ResourceData, meta interface{}) e
...
+	if privateIPv6 := findIPv6AddrByType(droplet, "private"); privateIPv6 != "" {
+		d.Set("ipv6_address_private", strings.ToLower(privateIPv6))
+	}
 
-	if privateIPv4 := findIPv4AddrByType(droplet, "private"); privateIPv4 != "" {
-		d.Set("private_networking", true)
-		d.Set("ipv4_address_private", privateIPv4)
-	} else {
-		d.Set("private_networking", false)
-		d.Set("ipv4_address_private", "")
```
andrewsomething: @TFaga Is defaulting the values for `ipv6_address` and `ipv4_address_private` to an empty string generally problematic or just this implementation? Here's a use case that seemed reasonable to me: https://github.com/terraform-providers/terraform-provider-digitalocean/issues/92

TFaga: In general no, I wouldn’t say it’s problamatic. I was aiming to simplify the code together with reading the enabled features of a droplet. But supporting the usecase in #92 for a computed property is probably fine and reasonable. Especially since that was the case now and it’d be better to not break existing functionality.

I’ll update the PR.

Issue #146, opened 2018-10-04
Cert Resource constantly updates


troywilson:
Hi there,

Thank you for opening an issue. Please note that we try to keep the Terraform issue tracker reserved for bug reports and feature requests. For general usage questions, please see: https://www.terraform.io/community.html.

### Terraform Version
Run `terraform -v` to show the version. If you are not running the latest version of Terraform, please upgrade because your issue may have already been fixed.

0.11.8

### Affected Resource(s)
Please list the resources as a list, for example:
- digitalocean_certificate

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl

resource "tls_private_key" "primary-cert-private_key" {
  algorithm = "RSA"
}

resource "tls_cert_request" "primary" {
  key_algorithm   = "RSA"
  private_key_pem = "${tls_private_key.primary-cert-private_key.private_key_pem}"
  dns_names       = ["${var.primary_dns_name}", "*.${var.primary_dns_name}"]

  subject {
    common_name = "${var.primary_dns_name}"
  }
}

resource "acme_certificate" "primary" {
  account_key_pem         = "${tls_private_key.reg_private_key.private_key_pem}"
  certificate_request_pem = "${tls_cert_request.primary.cert_request_pem}"
  depends_on              = ["aws_route53_record.primary-caa-wildcard"]
  min_days_remaining      = 31

  dns_challenge {
    provider = "route53"
  }
}

resource "digitalocean_certificate" "primary" {
  name              = "${var.primary_dns_name}"
  private_key       = "${tls_private_key.primary-cert-private_key.private_key_pem}"
  leaf_certificate  = "${acme_certificate.primary.certificate_pem}"
  certificate_chain = "${acme_certificate.primary.issuer_pem}"
}
```

### Expected Behavior
The cert is installed once and only re-installed when the underlining acme cert is changed.

### Actual Behavior
The cert is deleted and re-installed into DO every time. It was working fine before the 1.0 upgrade. Sample output from apply:

-/+ digitalocean_certificate.primary (new resource required)
      id:                 "xxxxxxxxxxxxxxxxxxxxxxxxx" => \<computed\> (forces new resource)
      certificate_chain:  "DATA CLIPPED"
      domains.#:          "2" => "0" (forces new resource)
      domains.xxxxxxx: "domain1.com" => "" (forces new resource)
[...]

PR #148, opened 2018-10-04
Suppress diff for dns names on custom certificate resources (Fixes: #146)

andrewsomething:
When creating a custom certificate, the DigitalOcean API does not accept the `domains` attribute. This is only used when creating a Let's Encrypt certificate. Though the API does return this attribute when reading an existing custom certificate as the information is inferred from the provided certificate. This leads to the situation seen in #146. This PR resolves this issue by suppressing the diff for the `domains` attribute when the `type` is `custom`.

Most of the code added here is in the test. Our previous test did not catch this issue as it used Terraform's `acctest.RandTLSCert` to generate the test data. This function does not set the DNS names for the certificate. The new code is based on it, but allows for passing DNS name.

### DIFF
--- a/digitalocean/resource_digitalocean_certificate.go
+++ b/digitalocean/resource_digitalocean_certificate.go
@@ -60,6 +60,7 @@ func resourceDigitalOceanCertificate() *schema.Resource {
 				Optional:      true,
 				ForceNew:      true,
 				ConflictsWith: []string{"private_key", "leaf_certificate", "certificate_chain"},
+				// The domains attribute is computed for custom certs and should be ignored in diffs.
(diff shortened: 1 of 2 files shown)


digitalocean/resource_digitalocean_certificate.go
```diff
@@ -60,6 +60,9 @@ func resourceDigitalOceanCertificate() *schema.Resource {
 				Optional:      true,
 				ForceNew:      true,
 				ConflictsWith: []string{"private_key", "leaf_certificate", "certificate_chain"},
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return d.Get("type") == "custom"
```
eddiezane: I'd add a comment as to why this is necessary.

PR #164, opened 2018-11-13
Update dead link

dramich:
The link for api docs on DO has been changed, update link to reflect that.


andrewsomething:
:+1: Good catch. Thanks @dramich!

PR #169, opened 2018-12-05
K8s

nicholasjackson:
Raising this to start some provisional chats about the new K8s resource.

This PR adds 2 new resources `kubernetes_cluster` and `kubernetes_node_pool` as well as one data source `kubernetes_cluster`.

To create a cluster it is required that you specify a `cluster` resource and define the default node pool.  Additional node pools are defined using the `node_pool` resource.  Initially I wanted to just allow multiple `node_pool` blocks on the cluster however this caused problems when performing updates.  The only way to ensure that the correct node pool is updated was to limit the pool in the cluster to 1 and to add a hidden tag so that it can be read correctly later.

It is possible to change the size on the default node pool, this will be a documented note that changing the size will destroy the node pool and then re-create it.  All other parameters can be performed by a non-destructive update.  Changing the size for a node_pool resource would starts Terraform`s destroy / create cycle as normal.

Once a cluster has been created the K8s config will be available as an attribute, this is currently read from the API.

```hcl
 1 resource "digitalocean_kubernetes_cluster" "foobar" {                                                                                                                                                                                                                                                                                                         
  2   name    = "testing"                                                                                                                                  
  3   region  = "lon1"                                                                                                                                     
  4   version = "1.12.1-do.2"                                                                                                                              
  5   tags    = ["foo", "bar"]                                                                                                                             
  6                                                                                                                                                        
  7   node_pool {                                                                                                                                          
  8     size       = "s-1vcpu-2gb3"                                                                                                                        
  9     node_count = 1                                                                                                                                     
 10     tags       = ["one", "two"] // Tags from cluster are automatically added to node pools                              
 11   }                                                                                                                                                    
 12 }                                                                                                                                                      
 13                                                                                                                                                        
 14 // Add a second node pool                                                                                                                              
 15 resource "digitalocean_kubernetes_node_pool" "barfoo" {                                                                                                
 16   cluster_id = "${digitalocean_kubernetes_cluster.foobar.id}"                                                                                          
 17                                                                                                                                                        
 18   size       = "s-1vcpu-2gb3"                                                                                                                          
 19   node_count = 1                                                                                                                                       
 20   tags       = ["three", "four"] // Tags from cluster are automatically added to node pools                             
 21 }
```

I have not added the documentation yet as just wanted to run the schema past you before I write the docs.

Kind regards,

Nic


aybabtme:
Having all the node pools on the `cluster` resource I think would be preferable.

> Additional node pools are defined using the node_pool resource. Initially I wanted to just allow multiple node_pool blocks on the cluster however this caused problems when performing updates. The only way to ensure that the correct node pool is updated was to limit the pool in the cluster to 1 and to add a hidden tag so that it can be read correctly later.

I think this could be done by requiring node pools to have a name, and then matching the updates on the name.

aybabtme:
I'll try something about my last comment, based on your PR. If anything comes up, I'll make a PR on your PR. :)

andrewsomething:
:books: Docs in this PR to your branch: https://github.com/nicholasjackson/terraform-provider-digitalocean/pull/1

andrewsomething:
Thanks for all your work on this Nic! :shipit:

PR #172, opened 2018-12-12
Correct Kubernetes example

swestcott:


andrewsomething:
:+1: Good catch. Thanks!

PR #175, opened 2018-12-14
Fix typo and format in docs

huikaihoo:


andrewsomething:
:+1:  Good catch. Thanks for the PR!

PR #185, opened 2019-02-02
Fix typo in node pool name

plantfansam:
Just a tiny fix to change `woker-pool` to `worker-pool`.


eddiezane:
@plantfansam thanks for the fix!

PR #190, opened 2019-02-22
provider: Refactor logic for logging API requests/responses

radeksimko:
This should be practically no-op change as even the logging format is the same. Using http Transport is however more common way of logging requests and responses which is compatible with most SDKs so I'm merely aligning this to match what we have elsewhere.


andrewsomething:
:+1: LGTM, thanks @radeksimko

PR #193, opened 2019-03-05
Rebuild droplet on image change

pysysops:
Hi 👋🏼

This PR makes use of the DigitalOcean rebuild API on changes to a droplet's image attribute.

Currently, when you change an image the entire droplet resource is destroyed in DigitalOcean releasing the droplets IP addresses, volume attachments and other resources that are related to a droplet's ID.

By using the rebuild API a droplet will maintain it's IP addresses and other related resources on image changes.

I welcome any feedback or thoughts and experiences on this.

Thanks,

Tim Birkett.

### DIFF
--- a/digitalocean/resource_digitalocean_droplet.go
+++ b/digitalocean/resource_digitalocean_droplet.go
@@ -31,7 +31,6 @@ func resourceDigitalOceanDroplet() *schema.Resource {
 			"image": {
 				Type:         schema.TypeString,
 				Required:     true,
-				ForceNew:     true,
 				ValidateFunc: validation.NoZeroValues,
 			},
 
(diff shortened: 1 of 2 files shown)


pysysops:
DigitalOcean dashboard output and history during and after testing:

screenshot 2019-03-05 at 18 10 33

screenshot 2019-03-05 at 18 12 00

pysysops:
I haven't checked how this interoperates with lifecycle configuration. It might be better off refactoring to be an optional flag (somehow) or just not adding this functionality as it could result in users being precious with droplet IP addresses rather than architecting solutions using Load Balancers / Floating IPs.

andrewsomething:
Hi @pysysops! Sorry for the delayed review. I agree that this would be better as an optional flag for two reasons. Besides the one you mentioned, this would also be a fairly big change from current behavior that may have some surprising consequences. In particular, rebuilding a Droplet from one distro to another is something we discourage as it can be fragile especially when things like cloudint scripts are involved.

pysysops:
Closing for now. Destroy and create feels like it's the right model for the cloud and helps avoid droplet IPs becoming precious. Was fun to learn a bit about Terraform provider internals.

PR #196, opened 2019-03-20
[MODULES] Updating digitalocean/godo to latest version

slapula:
I apologize in advance but I'm completely new to managing go modules so I'm not sure if this is the correct way to update this dependency.


appilon:
hello @slapula, you have done the necessary steps if we were completely switched to go modules. We have in fact only made a partial switch and still vendor all our dependencies (for now!). You just need to run
```
$ go mod tidy
$ go mod vendor
```
It's good to run `go mod tidy` anytime after changing a dependency as it can declutter the `go.mod` file if it's possible.

slapula:
@appilon Thanks for the feedback!  This makes more sense now that I look at it.  I've went ahead and updated this PR with the rest of the needed changes.

eddiezane:
@slapula thanks!

Issue #163, opened 2018-11-10
digitalocean_certificate not working with computed values


eb08a167:
### Terraform Version
Terraform v0.11.10
+ provider.acme v1.0.1
+ provider.digitalocean v1.0.2
+ provider.tls v1.2.0

### Affected Resource(s)
+ digitalocean_certificate

### Terraform Configuration Files
```hcl
resource "tls_private_key" "acme" {
  algorithm = "RSA"
}

resource "acme_registration" "primary" {
  account_key_pem = "${tls_private_key.acme.private_key_pem}"
  email_address   = "${var.acme_email_address}"
}

resource "acme_certificate" "primary" {
  account_key_pem = "${acme_registration.primary.account_key_pem}"
  common_name     = "*.${digitalocean_domain.primary.name}"

  dns_challenge {
    provider = "digitalocean"

    config {
      DO_AUTH_TOKEN = "${var.do_token}"
    }
  }
}

resource "digitalocean_certificate" "primary" {
  name              = "primary-certificate"
  private_key       = "${acme_certificate.primary.private_key_pem}"
  leaf_certificate  = "${acme_certificate.primary.certificate_pem}"
  certificate_chain = "${acme_certificate.primary.issuer_pem}"
}
```

### Debug Output

https://pastebin.com/gWH0iM0q

### Expected Behavior

digitalocean_certificate is created on apply

### Actual Behavior

Terraform fails with ```* digitalocean_certificate.primary: `private_key` is required for when type is `custom` or empty``` error

### Steps to Reproduce

1. `terraform apply`

### Important Factoids

It's possible to workaround the problem by performing the steps below:

1. Comment the `digitalocean_certificate` resource
2. Run `terraform apply`
3. Uncomment the `digitalocean_certificate` resource
4. Run `terraform apply`

After that the `digitalocean_certificate` resource is successfully created.

parruda:
I can confirm. Having the exact same problem here.

maxwedwards:
Is this not more likely to be the acme_certificate resource not providing the private key after a create and not a problem with the digital ocean provider? Have you tried an example without acme_certificate to eliminate it?

etienne-napoleone:
Same problem here:
```terraform
provider "acme" {
  server_url = "https://acme-staging-v02.api.letsencrypt.org/directory"
}

resource "tls_private_key" "private_key" {
  algorithm = "RSA"
}

resource "acme_registration" "reg" {
[...]

PR #202, opened 2019-03-28
Certificates: Move attribute validation from CustomizeDiff to Create (Fixes: #163).

andrewsomething:
When CustomizeDiff is run for a new resource, it does not have access to the
current state. In the digitalocean_certificate resource, we are using a
CustomizeDiff function to validate that certain attributes are used together
depending on the certificate type. When the value for one of those attributes
is interpolated, as in the case of generating a cert using the acme provider,
the validation fails since it depends on reading the value from state. As state
is not present, the value is interpreted as empty.

From the CustomizeDiff doc string:

> // The phases Terraform runs this in, and the state available via functions
> // like Get and GetChange, are as follows:
> //
> //  * New resource: One run with no state

https://godoc.org/github.com/hashicorp/terraform/helper/schema#Resource

This PR moves the validation from the CustomizeDiff function into the
Create function as well as adding acceptance tests:

```
$ make testacc TESTARGS='-run=TestAccDigitalOceanCertificate_ExpectedErrors'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccDigitalOceanCertificate_ExpectedE$
?       github.com/terraform-providers/terraform-provider-digitalocean  [no test files]
=== RUN   TestAccDigitalOceanCertificate_ExpectedErrors
--- PASS: TestAccDigitalOceanCertificate_ExpectedErrors (0.10s)
PASS
ok      github.com/terraform-providers/terraform-provider-digitalocean/digitalocean     (cached)
```

### DIFF
--- a/digitalocean/resource_digitalocean_certificate.go
+++ b/digitalocean/resource_digitalocean_certificate.go
@@ -92,27 +92,6 @@ func resourceDigitalOceanCertificate() *schema.Resource {
 				Computed: true,
 			},
 		},
-
-		CustomizeDiff: func(diff *schema.ResourceDiff, v interface{}) error {
-
-			certificateType := diff.Get("type").(string)
-			if certificateType == "custom" {
-				if _, ok := diff.GetOk("private_key"); !ok {
-					return fmt.Errorf("`private_key` is required for when type is `custom` or empty")
-				}
-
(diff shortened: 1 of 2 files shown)


eddiezane:
Nice job with this one

PR #209, opened 2019-04-14
Normalize case for region attribute across all resources

### LINKED ISSUE #208: Normalize case for region attribute across all resources
The DigitalOcean API expects region slugs to be provided in lower case. Currently we normalize the case in the Droplet, Kubernetes cluster, and database cluster resources using a `StateFunc`. We should consistently do this across all of the resources that accept a region attribute from a user.

Example `StateFunc` from the Droplet resource:

https://github.com/terraform-providers/terraform-provider-digitalocean/blob/9b3175dc9ba63d14815e6faf2dba79446958f8b5/digitalocean/resource_digitalocean_droplet.go#L58-L61

thojkooi:
Fix #208

### DIFF
--- a/digitalocean/datasource_digitalocean_droplet_snapshot.go
+++ b/digitalocean/datasource_digitalocean_droplet_snapshot.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"strings"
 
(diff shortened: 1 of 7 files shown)


andrewsomething:
:+1:  Thanks for the PR @thojkooi!

PR #210, opened 2019-04-15
Updates the spaces resource to expose the urn.

trentrosenbaum:
This PR Introduces the URN attribute into the spaces/bucket resource.  Unlike the other resources that use the structs from the godo client, the spaces resource does not do this in the read function.  As a result the URN is created in the read function.

**NOTE:**  DigitialOcean API is not allowing the use of nyc3 for spaces/bucket creation and so the use of ams3 as a region has been used to allow the tests to execute.


andrewsomething:
:+1:  - Good call on using AMS3.

PR #211, opened 2019-04-15
Updates the floating ip resource to expose the urn.

trentrosenbaum:
This PR introduces the URN attribute for the Floating ip resource.  This covers the datasource and the resource.  Alterations also include the website to document the new attribute.


andrewsomething:
:+1:

PR #214, opened 2019-04-15
Feature/introduce urn attribute to loadbalancer

trentrosenbaum:
This PR introduces the URN attribute for the load balancer resource.  This covers the datasource and the resource.  Alterations also include the website to document the new attribute.


andrewsomething:
:shipit:

PR #218, opened 2019-04-18
Feature/introduce projects resource

pegerto:
This PR introduces the DigitalOcean Project resource and supportive documentation/tests. The first tests cover the creation and updating of the basic fields in the project. Later tests add and update/remove the other resource associations with the project.

Notes:
To remove the resources from the project, the old resources are linked to the current default project and then the new resources are linked back to the managed project. The new list may include resources from the old list if they are still linked to the project.

The tests use the droplet and spaces resources, but there are other resources. The main requirement is the availability of the URN attribute on the resource and this was before this PR.

### DIFF
--- a/digitalocean/resource_digitalocean_project.go
+++ b/digitalocean/resource_digitalocean_project.go
@@ -1,0 +1,56 @@
+package digitalocean
+
+import (
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+)
+
+func resourceDigitalOceanProject() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceDigitalOceanProjectCreate,
+		Read:   resourceDigitalOceanProjectRead,
+		Update: resourceDigitalOceanProjectUpdate,
+		Delete: resourceDigitalOceanProjectDelete,
+		Importer: &schema.ResourceImporter{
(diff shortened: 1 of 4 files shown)


digitalocean/resource_digitalocean_project.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+	projectId := d.Id()
+	defaultProjectId := defaultProject.ID
+
+	log.Printf("[DEBUG] Default project located, ID: %s", defaultProjectId)
+
+	if v, ok := d.GetOk("resources"); ok {
+
+		_, err := assignResourcesToProject(client, defaultProjectId, v.(*schema.Set))
```
andrewsomething: Seem like we could use `assignResourcesToDefaultProject` to reduce some code duplication here.

website/docs/r/project.html.markdown
```diff
@@ -0,0 +1,81 @@
...
+}
+
+resource "digitalocean_project" "playground" {
+    name = "playground"
+    description = "A project to represent development resources."
+    purpose = "Web Application"
+    environment = "Development"
+    resources = ["${digitalocean_droplet.foobar.urn}"]
+}
+
```
andrewsomething: Code block needs to be closed.

andrewsomething:
:+1:  Great work! Merged after addressing the above.

PR #221, opened 2019-04-23
Update sidebar and fix typo.

AAbouZaid:
Hello,

The `droplet-snapshot` pages were missing in the docs sidebar so I added them (datasource and resource).

Also there was a typo in `kubernetes_node_pool` file where it was "kuber**en**tes" also the file wasn't follow the convention where it has "digital_ocean" instead "digitalocean".

Another resource is missing is `resource_digitalocean_droplet_migrate` but there is no docs for it at all (so it's not added because no docs).
Maybe @TFaga knows about it (the only name I see in git log related to that file).

Thanks.


andrewsomething:
:man_facepalming: Doh! Great catches.

Re: `resource_digitalocean_droplet_migrate` - That file is not actually a resource itself. It is a helper to automatically update the state file for a past change to the `digitalocean_droplet`  resource's schema file.

Thanks for the contribution!

Issue #181, opened 2019-01-17
digitalocean_droplet: ipv6_address_private doesn't exist


pysysops:
Good day!

Thanks for this fantastic provider! I'm currently in the process of creating some DigitalOcean modules and have potentially found a bug with `private_networking` and `ipv6`.

### Terraform Version
Terraform v0.11.11
+ provider.digitalocean v1.1.0

### Affected Resource(s)
- digitalocean_droplet

### Terraform Configuration Files
Example:
```hcl

resource "digitalocean_droplet" "droplet" {
  count = 1

  image = "<< image_id >>"
  name  = "droplet"

  region = "ams3"
  size   = "s-1vcpu-1gb"

  ipv6               = true
  private_networking = true
}

resource "digitalocean_record" "private_aaaa" {
  count = 1

  domain = "example.com"
  type   = "AAAA"
  name   = "${element(digitalocean_droplet.droplet.*.name, count.index)}"
  value  = "${element(digitalocean_droplet.droplet.*.ipv6_address_private, count.index)}"
}

```

### Debug Output
```
* digitalocean_record.private_aaaa.0: Failed to create record: POST https://api.digitalocean.com/v2/domains/private.stg.ghostfoundation.org/records: 422 Data can't be blank and IP address did not match IPv6 format (e.g. 2001:db8::ff00:42:8329).
* module.apps.module.droplets.digitalocean_record.private_aaaa[0]: 1 error(s) occurred:
```

### Expected Behavior
I'm not sure if DigitalOcean have supported IPv6 private networking and removed it, removed it from the API or never supported it?

The documentation here: [digitalocean_droplet](https://www.terraform.io/docs/providers/do/r/droplet.html#ipv6_address_private) suggests that it the provider supports private IPv6 addresses 🤔 

### Actual Behavior
No private IPv6 addresses available from the API or visible in the DigitalOcean dashboard.

andrewsomething:
DigitalOcean has never supported IPv6 private networking. Though it looks like this attribute has been in the code base going almost all the way to the beginning, pre-dating my involvement with the project: https://github.com/hashicorp/terraform/commit/872acb3740684bbc75899ad21e5bae0219920be0

pysysops:
I had a suspicion that may have been the case. Thanks for the quick update and clarification @andrewsomething 😄

stack72:
@andrewsomething is this correct? I see IPv6 related code in the godo SDK

this is why this is able to happen:

[...]

PR #252, opened 2019-06-09
resource/droplet: Removing the IPv6 private parameter

stack72:
Fixes: #181

DigitalOcean doesn't support IPv6 private networking. Therefore,
having a parameter in the schema for this isn't correct. It has
been marked as `removed` with an appropriate error message


andrewsomething:
:shipit: LGTM. Thanks @stack72!

PR #257, opened 2019-06-19
Fixed CAA record tag

danielsreichenbach:
According to https://tools.ietf.org/html/rfc6844 the valid tag is "issuewild", not "wildissue".


andrewsomething:
:+1:  Thanks @danielsreichenbach!

Issue #260, opened 2019-06-30
Question: Could we specify DIGITALOCEAN_ACCESS_TOKEN rather than DIGITALOCEAN_TOKEN? 


KadenLNelson:
Following the authentication strategies provided by `doctl`, would it be worth changing the token variable to `DIGITALOCEAN_ACCESS_TOKEN` instead of `DIGITALOCEAN_TOKEN`? To provide a universal experience when interacting with both `doctl` and the DigitalOcean Terraform provider.

PR #264, opened 2019-07-04
provider: Add the ability to set DIGITALOCEAN_ACCESS_TOKEN

stack72:
Fixes: #260

This follows the same usage as `doctl` so not having to reset env
vars would be a better experience. This DOT NOT remove DIGITALOCEAN_TOKEN,
this just adds the ability to specify one of the two and the
provider will continue to work as expected

### DIFF
--- a/digitalocean/provider.go
+++ b/digitalocean/provider.go
@@ -10,8 +10,8 @@ func Provider() terraform.ResourceProvider {
 	return &schema.Provider{
 		Schema: map[string]*schema.Schema{
 			"token": {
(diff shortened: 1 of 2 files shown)


stack72:
@andrewsomething I'd love your thoughts on this PR - I think this is a good strategy to continue existing doctl user experience

andrewsomething:
:+1:  Agreed. Thanks @stack72! 

(Pushed a commit running `gofmt` to make CI happy.)

KadenLNelson:
Thank you.

PR #265, opened 2019-07-09
Add resource `digitalocean_database_connection_pool`

Photonios:
This is a very early version. Any feedback would be appreciated. Especially on how to add proper tests for this.


digitalocean/resource_digitialocean_database_connection_pool.go
```diff
@@ -0,0 +1,192 @@
...
+		Create: resourceDigitalOceanDatabaseConnectionPoolCreate,
+		Read:   resourceDigitalOceanDatabaseConnectionPoolRead,
+		// Update: resourceDigitalOceanDatabaseConnectionPoolUpdate,
+		Delete: resourceDigitalOceanDatabaseConnectionPoolDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"db_id": {
```
andrewsomething: How about `database_cluster`?

digitalocean/resource_digitialocean_database_connection_pool.go
```diff
@@ -0,0 +1,192 @@
...
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+
+			"pool_uri": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"pool_database": {
```
andrewsomething: `pool_database` and `pool_user` should always be the same as the `database` and `user`. I don't think there is a need to export them as separate attributes.

Photonios: Absolutely! I added the prefix to avoid a conflict between the computed `database` and the argument `database`. However, as you pointed out, they're always the same. It was late in the evening.

I'll patch it up :-)

digitalocean/resource_digitialocean_database_connection_pool.go
```diff
@@ -0,0 +1,192 @@
...
+		return fmt.Errorf("Error creating DatabaseConnectionPool: %s", err)
+	}
+
+	d.SetId(createConnectionPoolID(dbID, pool.Name))
+	log.Printf("[INFO] DatabaseConnectionPool Name: %s", pool.Name)
+
+	return resourceDigitalOceanDatabaseConnectionPoolRead(d, meta)
+}
+
+func resourceDigitalOceanDatabaseConnectionPoolUpdate(d *schema.ResourceData, meta interface{}) error {
```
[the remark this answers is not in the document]
Photonios: Yup! I was even "smart" enough to comment out the `Update` property, so it's unused. I'll remove it.

Photonios:
Thanks for the quick feedback. All valid points and much appreciated. My apologies for the rough code. I wanted to submit something working and go from there to avoid having to rework massively.

The links regarding testing look useful. I'll add some tests for this.

Thanks the quick turn around. I'll do my best to provide the same :-)

botzill:
Hi.

Any updates on this PR?

andrewsomething:
Thanks again for your work on this @Photonios I've gone ahead and finished it off in https://github.com/terraform-providers/terraform-provider-digitalocean/pull/334

Issue #272, opened 2019-07-22
Terraform crashes while kubernetes cluster is updating


MoritzKn:
### Terraform Version

```
Terraform v0.12.5
+ provider.aws v2.19.0
+ provider.digitalocean v1.5.0
+ provider.kubernetes v1.8.1
```

### Affected Resource(s)

- `digitalocean_kubernetes_cluster` (i think)

### Terraform Configuration Files

```
resource "digitalocean_kubernetes_cluster" "cluster" {
  name    = "some-cluster"
  region  = "fra1"
  version = "1.13.7-do.f.0"

  node_pool {
    name       = "woker-pool"
    size       = "s-2vcpu-4gb"
    node_count = 1
  }
}
```

### Panic Output

Using terraform `0.12.3`:
https://gist.github.com/MoritzKn/1020c309bff80941e6f0fe58be2c503d

Using terraform `0.12.5` (latest):
https://gist.github.com/MoritzKn/ffaf5e646aa8d007f2c5461981701661

### Expected Behavior

No crash

### Actual Behavior

Terraform crashes

### Steps to Reproduce

I'm not sure if it is reproducible but I did the following steps leading up to the crash:

1. I clicked the update Kubernetes button in the DO web interface
2. I changed the version in the terraform config to the one displayed in the web interface
3. I ran `terraform plan` -> crash
4. I updated terraform to the lastest version (from `0.12.3` to `0.12.5`)
5. I ran `terraform plan` again -> still crashing
6. I waited a few minutes and it stated working again

andrewsomething:
Pulling out the actual panic:

```
  panic: runtime error: invalid memory address or nil pointer dereference
  [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x19ac5e0]
  goroutine 50 [running]:
  github.com/terraform-providers/terraform-provider-digitalocean/digitalocean.resourceDigitalOceanKubernetesClusterRead(0xc0001f0cb0, 0x1af3d60, 0xc000541a70, 0xc0001f0cb0, 0x0)
  	/opt/teamcity-agent/work/5d79fe75d4460a2f/src/github.com/terraform-providers/terraform-provider-digitalocean/digitalocean/resource_digitalocean_kubernetes_cluster.go:181 +0xe0
[...]

PR #273, opened 2019-07-23
Consistently protect against nil response in error handling (Fixes: #272).

andrewsomething:
In many cases we handle 404s different than other errors. Though if an error contains no response at all, we panic. (See: #272) Some resources already protect against this, but not all of them.


eddiezane:
Nice catch on all these.

Issue #263, opened 2019-07-01
Extend data source types to include lists


syntaqx:
## Feature request: Data source lists, filtering

With a few exceptions, there seems to be quite a few data sources missing from the provider, specifically list methods. I think it would be a massive improvement to be able to work with existing resources, especially in multi-team environments where sources are controlled by other state files.

This is what I've been using to hack the functionality for the time being, it might be useful for the proposal or anyone else who might need it until progress can be made getting this into the provider -- I simply include this in my `.envrc` (you could also use `.bash_profile`):

```sh
#!/usr/bin/env bash
set -e -o pipefail

JQ_FLAGS=${JQ_FLAGS:-"-cer"}

echoerr() { echo "$@" 1>&2; }

reqs=(jq doctl)
for i in "${reqs[@]}"; do
  if ! command -v "$i" &>/dev/null; then
    echoerr "$i is required but is not installed."
    echoerr "Exiting."
    exit 1
  fi
done

doctlwrap() { doctl "$@" -o json; }

describe_account() {
  doctlwrap account get | jq $JQ_FLAGS \
  '{
    uuid,
    email,
    email_verified,
    limits: {
      droplet: .droplet_limit,
      volume: .volume_limit,
      floating_ip: .floating_ip_limit
    },
    status,
    status_message
  }'
}

list_regions() {
  doctlwrap compute region list | jq $JQ_FLAGS \
  '[ .[] | {
    name,
    slug,
    features,
    available
  } ]'
}

list_sizes() {
  doctlwrap compute size list | jq $JQ_FLAGS \
  '[ .[] | {
    slug,
    resources: {
      cpus: .vcpus,
      disk: "\(.disk)GB",
      memory: "\(.memory)MB"
    },
    price: {
      monthly: .price_monthly,
      hourly: .price_hourly
    },
    bandwidth: {
      transfer: "\(.transfer)TB"
    }
  } ]'
}

list_tags() {
  doctlwrap compute tag list | jq $JQ_FLAGS \
  '[ .[] | {
    name,
    resources
  } ]'
}

export TF_VAR_do_account="$(describe_account)"
export TF_VAR_do_regions="$(list_regions)"
export TF_VAR_do_sizes="$(list_sizes)"
export TF_VAR_do_tags="$(list_tags)"
```

Which is then used to populate an equivalent `dynamic_do_variables.tf`:

```hcl
variable "do_account" {
  type = object({
    uuid           = string
    email          = string
[...]

PR #281, opened 2019-08-12
Add data source for regions

syntaqx:
First draft for being able to list regions for an account. Not sure if I'm happy with the usage but it's the best I could think of for the time being. Would love input.

```hcl
data "digitalocean_regions" "foobar" {
}

output "digitalocean_regions" {
   value = digitalocean_regions.foobar.regions
}

# regions = ["nyc1", "sfo1", "nyc2", ...]
```

In reference to #263

> __Note:__ The update to the `vendor/github.com/mitchellh/copystructure/README.md` is Git fixing CRLF whitespacing. Let me know if I need to figure out how to undo this - it's automatic for my git setup unfortunately.


digitalocean/datasource_digitalocean_regions.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/digitalocean/godo"
+	"github.com/hashicorp/terraform/helper/hashcode"
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+func dataSourceDigitalOceanRegions() *schema.Resource {
+	return &schema.Resource{
+		Read:   dataSourceDigitalOceanRegionsRead,
+		Schema: map[string]*schema.Schema{},
```
andrewsomething: My initial thought would be for the schema to look something like:
```suggestion
		Schema: map[string]*schema.Schema{
			"regions": {
				Type:     schema.TypeSet,
				Computed: true,
				Elem: &schema.Resource{
					Schema: map[string]*schema.Schema{
						"name": {
							Type:     schema.TypeString,
							Computed: true,
						},
						"slug": {
							Type:     schema.TypeString,
							Computed: true,
						},
					},
				},
			},
		},
```

Which makes the UI like:

```
data "digitalocean_regions" "foobar" {
}

resource "digitalocean_droplet" "web" {
  image  = "ubuntu-18-04-x64"
  name   = "web-1"
  region = "nyc2"
  size   = digitalocean_regions.foobar.regions[0].slug
}

output "digitalocean_regions" {
   value = digitalocean_regions.foobar.regions
}

# digitalocean_regions = [
#   {
#     "name" = "New York 1"
#     "slug" = "nyc1"
#   },
#   {
#     "name" = "New York 2"
#     "slug" = "nyc2"
#   },
#  # and so on...
# ]
```

I think this matches the API more closely and makes it more extensible for any additional information about the regions that might be useful.

andrewsomething:
Hi @syntaqx! Sorry for the delayed review. I think this is a great start, and it will definitely be very useful. I've made a suggestion inline about a different approach for the schema. Let me know what you think.

syntaqx:
@andrewsomething If we're going to go that far, might as well just expose the whole API object. Let me know if you like agree!

andrewsomething:
> If we're going to go that far, might as well just expose the whole API object. Let me know if you like/agree!

Sounds like a plan!

digitalocean/datasource_digitalocean_regions.go
```diff
@@ -0,0 +1,86 @@
...
+		page, err := resp.Links.CurrentPage()
+		if err != nil {
+			return fmt.Errorf("Error retrieving ssh keys: %s", err)
+		}
+
+		opts.Page = page + 1
+	}
+
+	d.SetId(fmt.Sprintf("%d", hashcode.String(fmt.Sprintf("%+v\n", opts))))
+	d.Set("regions", regionList)
```
andrewsomething: For this kind of nested schema, you'll need to loop through the results with something like:

```
	regionsResult := make([]map[string]interface{}, 0)
	for _, r := range regionList {
		item := make(map[string]interface{})
		item["name"] = r.Name
		item["slug"] = r.Slug
		regionsResult = append(regionsResult, item)
	}

	d.Set("regions", regionsResult)
```

Example elsewhere in the code base:
https://github.com/terraform-providers/terraform-provider-digitalocean/blob/121c319d89b21926cb92fe1557c6c88f5e4294ce/digitalocean/resource_digitalocean_database_cluster.go#L334

andrewsomething:
Closing in favor of https://github.com/terraform-providers/terraform-provider-digitalocean/pull/380

Issue #283, opened 2019-08-20
DigitalOcean Droplet not destroyed despite Terraform reporting destruction a success.


manibatra:
We ran the terraform job to delete the droplet which was reported as a success. But the droplet was not destroyed and was running when we logged into the dashboard. In the past, we have not noticed that such a droplet rejoins our cluster if not manually deleted from the dashboard.

### Terraform Version
`terraform version = 0.11.1`
`terraform-provider-digitalocean version = 1.0.2`

### Affected Resource(s)
- digitalocean_droplet

### Debug Output
Terraform returning successful destruction : 
```
module.delivery-12.module.delivery.digitalocean_droplet.droplet: Destroying... (ID: 132167273)
module.delivery-12.delivery.digitalocean_droplet.droplet: Still destroying... (ID: 132167273, 10s elapsed)
module.delivery-12.module.delivery.digitalocean_droplet.droplet: Destruction complete after 11s
```

### Expected Behavior
The droplet is destroyed.

### Actual Behavior
The droplet was not destroyed and was running when logged into the DigitalOcean dashboard. Had to be destroyed from the dashboard.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. We have seen this issue happen for droplets that are having issues and are undergoing/will undergo migration due to physical host issues. Hence, find a similar droplet.
2. `terraform apply apply.tfplan` where the plan has droplet marked for destruction.
3. The droplet is running when logged into the DigitalOcean dashboard despite Terraform indicating successful destruction.

eddiezane:
@manibatra thanks for reporting this and apologies for any headaches this caused y'all.

Digging into this with the team. My first reaction is this is something we might need to address internally as opposed to in the provider.

PR #284, opened 2019-08-21
Verify Droplet is destroyed before removing from state (Fixes: #283).

andrewsomething:
This polls a Droplet after a delete request and waits for it to 404 before removing it from state in order to prevent issues like #283 where an unexpected error on the DigitalOcean side left the Droplet active but removed from state. We don't reuse `waitForDropletAttribute` here as we want a much quicker timeout.

### DIFF
--- a/digitalocean/resource_digitalocean_droplet.go
+++ b/digitalocean/resource_digitalocean_droplet.go
@@ -626,18 +626,36 @@ func resourceDigitalOceanDropletDelete(d *schema.ResourceData, meta interface{})
 	log.Printf("[INFO] Deleting droplet: %s", d.Id())
 
 	// Destroy the droplet
-	_, err = client.Droplets.Delete(context.Background(), id)
+	resp, err := client.Droplets.Delete(context.Background(), id)
 
-	// Handle remotely destroyed droplets
-	if err != nil && strings.Contains(err.Error(), "404 Not Found") {
+	// Handle already destroyed droplets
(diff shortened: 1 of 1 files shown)


digitalocean/resource_digitalocean_droplet.go
```diff
@@ -626,20 +626,38 @@ func resourceDigitalOceanDropletDelete(d *schema.ResourceData, meta interface{})
...
 
 	return nil
 }
 
+func waitForDropletDestroy(d *schema.ResourceData, meta interface{}) (interface{}, error) {
+	log.Printf("[INFO] Waiting for droplet (%s) to be destroyed", d.Id())
+
+	stateConf := &resource.StateChangeConf{
+		Pending:    []string{"active"},
+		Target:     []string{"archived"},
```
eddiezane: Where do these states come from?

andrewsomething: `active` is the `status` attribute for a running Droplet. Realizing now that `off` should also be included as you can destroy a powered off Droplet. `archived` is an internal status for deleted Droplets. In practice, this will not be seen on the public API as it will 404 when it is destroyed.

This function will always return an error. 404 when the Droplet is successfully destroyed or a timeout if the destroy request fails unexpectedly on the DigitalOcean. I'm really just taking advantage of the Terraform's existing polling mechanism.

manibatra:
Thanks for this!

PR #289, opened 2019-08-31
Mark the kubernetes_cluster resource's kube_config field as sensitive

RoboticCheese:
This data includes sensitive information about the cluster, such as its private key. This prevents that data from ending up in, for example, logs on a CI system that does a `terraform plan` as part of its build.

With this applied, a `terraform plan` or `terraform apply` instead shows:

```
...
        ipv4_address   = "1.2.3.4"
        kube_config    = (sensitive value)
        name           = "roboticcheese"
...
```

Also updates the acceptance tests, as 1.14.4-do.0 is no longer available.

It's been five years since I last looked at any Go and never since I poked around Terraform code, so apologies in advance for any issues with this that I'm not considering. At first, I thought it could just mark the `client_key` and `client_certificate` as sensitive (matching what gets redacted by default from `kubectl config view`), but it looks like Terraform won't redact fields from a schema nested inside the resource's schema like that.

### DIFF
--- a/digitalocean/resource_digitalocean_kubernetes_cluster.go
+++ b/digitalocean/resource_digitalocean_kubernetes_cluster.go
@@ -100,8 +100,9 @@ func resourceDigitalOceanKubernetesCluster() *schema.Resource {
 
 func kubernetesConfigSchema() *schema.Schema {
 	return &schema.Schema{
-		Type:     schema.TypeList,
-		Computed: true,
+		Type:      schema.TypeList,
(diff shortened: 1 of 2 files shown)


eddiezane:
@RoboticCheese thanks for the PR! This LGTM. Will let @andrewsomething weigh in and then we're all set to merge.

MattIPv4:
Hey, @RoboticCheese - Thanks a ton for this sweet PR! 😄

PR #291, opened 2019-09-07
remove usage of config pkg

appilon:
The terraform SDK is being extracted into its own repository. Some packages that are not considered apart of the API surface have been removed. This PR removes the usage of one such package.


andrewsomething:
Sorry for the delay on reviewing this! I swore I had already approved it :man_facepalming:

PR #294, opened 2019-09-19
Added DatabaseReplica resource support

### LINKED ISSUE #224: Read-only Database Replica Resource
With the addition of support for DigitalOcean Managed Databases, we should extend support for related resources such as read-only replicas.

Proposed format:

```hcl
resource "digitalocean_database_replica" "read-01" {
  cluster_id = "${digitalocean_database.foo.id}"
  name       = "read-replica-01"
  region     = "nyc3"
  size       = "db-s-2vcpu-4gb"
}
```

API: https://developers.digitalocean.com/documentation/v2/#create-a-read-only-replica

Related: https://github.com/terraform-providers/terraform-provider-digitalocean/issues/189

Zyqsempai:
Fixes #224 
Added DatabaseReplica resource support + tests

### DIFF
--- a/digitalocean/resource_digitalocean_database_replica.go
+++ b/digitalocean/resource_digitalocean_database_replica.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/digitalocean/godo"
@@ -17,7 +18,7 @@ func resourceDigitalOceanDatabaseReplica() *schema.Resource {
 		Read:   resourceDigitalOceanDatabaseReplicaRead,
 		Delete: resourceDigitalOceanDatabaseReplicaDelete,
 		Importer: &schema.ResourceImporter{
-			State: schema.ImportStatePassthrough,
+			State: resourceDigitalOceanDatabaseReplicaImport,
 		},
 
 		Schema: map[string]*schema.Schema{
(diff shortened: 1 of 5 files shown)


Zyqsempai:
@andrewsomething PTAL

Zyqsempai:
@andrewsomething Done!

digitalocean/resource_digitalocean_database_replica.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	replica, err = waitForDatabaseReplica(client, clusterId, "online", replica.Name)
+	if err != nil {
+		return fmt.Errorf("Error creating DatabaseReplica: %s", err)
+	}
+
+	log.Printf("[INFO] DatabaseReplica Name: %s", replica.Name)
+
+	return resourceDigitalOceanDatabaseClusterRead(d, meta)
```
andrewsomething: Missed this one on my first pass.

```suggestion
	return resourceDigitalOceanDatabaseReplicaRead(d, meta)
```

Zyqsempai:
@andreaso Done!

andrewsomething:
One thing to note: I did make some additional changes around importing. Since you need both the cluster's ID and the replicas name to access an existing replica, we can't just use the resource passthrough for imports. You need to pass both values in, eg:

```
terraform import digitalocean_database_replica.read-replica 245bcfd0-7f31-4ce6-a2bc-475a116cca97,read-replica
```

So it required a custom import function:

https://github.com/terraform-providers/terraform-provider-digitalocean/blob/master/digitalocean/resource_digitalocean_database_replica.go#L153-L162

I also added an acceptance test for that: https://github.com/terraform-providers/terraform-provider-digitalocean/blob/master/digitalocean/import_digitalocean_database_replica_test.go

PR #315, opened 2019-10-08
doks: Produce a valid kubeconfig file.

andrewsomething:
As a side-effect of moving to the `GetCredentials` endpoint in https://github.com/terraform-providers/terraform-provider-digitalocean/pull/311, the `raw_config` is no longer a usable kubeconfig file. This PR attempts to fix that.

@snormore, I'd love your input here. This feels more fragile than I'd like. Is there a better way to get the cluster and user names for use in a kubeconfig file short of making calls to both `GetKubeConfig` and `GetCredentials`?

### DIFF
--- a/digitalocean/resource_digitalocean_kubernetes_cluster_test.go
+++ b/digitalocean/resource_digitalocean_kubernetes_cluster_test.go
@@ -2,9 +2,11 @@ package digitalocean
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
 	"reflect"
 	"testing"
+	"time"
 
 	"github.com/digitalocean/godo"
 	"github.com/hashicorp/terraform/helper/resource"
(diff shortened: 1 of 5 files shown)


snormore:
I think the way you're doing it here is the best approach we have. It's unfortunate that `GetCredentials` doesn't include a `kubeconfig` sub-field containing this already. One reason we wouldn't want to call both `GetCredentials` and `GetKubeconfig` is that we'd generate 2 tokens, one for each. The approach you're taking here seems reasonable. The values of the cluster name and user names can pretty much be anything as long as they're consistent through the kubeconfig.

digitalocean/resource_digitalocean_kubernetes_cluster_test.go
```diff
@@ -372,3 +374,44 @@ func Test_filterTags(t *testing.T) {
...
+contexts:
+- context:
+    cluster: do-lon1-test-cluster
+    user: do-lon1-test-cluster-admin
+  name: do-lon1-test-cluster
+current-context: do-lon1-test-cluster
+users:
+- name: do-lon1-test-cluster-admin
+  user:
+    client-key-data: ""
```
snormore: How about making `client-key-data` and `client-certificate-data` `omitempty`?

go.mod
```diff
@@ -12,3 +12,5 @@ require (
 )
 
 replace git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999
+
+go 1.13
```
snormore: This kept getting re-added for me too, unsure why 🤔

andrewsomething: Reverted.

PR #324, opened 2019-10-26
Add data source for account

Permagate:
This PR creates a new data source `digitalocean_account` to retrieve information of the authenticated user used in the provider. It is currently a basic implementation as proposed in #263, but cannot retrieve information of other users yet.

### Example Usage

```hcl
data "digitalocean_account" "example" {
}
```

### DIFF
--- a/digitalocean/datasource_digitalocean_account.go
+++ b/digitalocean/datasource_digitalocean_account.go
@@ -1,0 +1,71 @@
+package digitalocean
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dataSourceDigitalOceanAccount() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceDigitalOceanAccountRead,
+		Schema: map[string]*schema.Schema{
+			"droplet_limit": {
+				Type:        schema.TypeInt,
+				Computed:    true,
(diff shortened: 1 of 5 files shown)


Permagate:
I'm also looking for input on implementing this data source to retrieve other accounts since  I'm pretty new in implementing terraform custom plugin.

Just fetching the account using token argument seems easy enough, but I'm afraid the token information will be stored in clear text inside the terraform state file which is a security issue. What's the suggested way to approach this?

andrewsomething:
:+1:  Looks good!

> Just fetching the account using token argument seems easy enough, but I'm afraid the token information will be stored in clear text inside the terraform state file which is a security issue. What's the suggested way to approach this?

Unfortunately, it would be stored in clear text in the state file. You can mark an attribute as sensitive, but that just redacts it from output.

Luckily, Terraform does have some built-in functionality for this. You can have multiple instance of a provider inside a single config. For example:

```hcl
provider "digitalocean" {
  token = "tottallysecrectbettertodowithanenvvar"
}

provider "digitalocean" {
  alias = "other-acct"
  token = "supersekrettokenforadifferentaccount"
}

data "digitalocean_account" "main-acct" {
}

data "digitalocean_account" "other-acct" {
  provider = digitalocean.other-acct
}
```

MattIPv4:
Hey, @Permagate - Thanks a ton for this sweet PR! 😄

PR #325, opened 2019-10-27
Add data source for sizes (with filter and sort)

Permagate:
Related with sizes data source proposal in #263. This PR creates a new data source `digitalocean_sizes` with a rudimentary implementation of filter and sort. The information is retrieved from https://developers.digitalocean.com/documentation/v2/#sizes.

### Example Usage

Most common usage will probably be to supply a size to droplet:

```hcl
data "digitalocean_sizes" "main" {
  filter {
    key    = "slug"
    values = ["s-1vcpu-1gb"]
  }
}

resource "digitalocean_droplet" "web" {
  image  = "ubuntu-18-04-x64"
  name   = "web-1"
  region = "sgp1"
  size   = element(data.digitalocean_sizes.main.sizes, 0).slug
}
```

The data source also supports multiple filters and sorts. For example, fetch sizes with virtual CPU 1 or 2 that are available "sgp1" region. Pick the cheapest one:

```hcl
data "digitalocean_sizes" "main" {
  filter {
    key    = "vcpus"
    values = [1, 2]
  }

  filter {
    key    = "regions"
    values = ["sgp1"]
  }

  sort {
    key       = "price_monthly"
    direction = "asc"
  }
}

resource "digitalocean_droplet" "web" {
  image  = "ubuntu-18-04-x64"
  name   = "web-1"
  region = "sgp1"
  size   = element(data.digitalocean_sizes.main.sizes, 0).slug
}
```

### DIFF
--- a/digitalocean/datasource_digitalocean_sizes.go
+++ b/digitalocean/datasource_digitalocean_sizes.go
@@ -1,8 +1,11 @@
 package digitalocean
 
 import (
+	"context"
 	"fmt"
 
+	"github.com/digitalocean/godo"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 )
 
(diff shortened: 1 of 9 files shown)


iilness2:
cool. works at my side

andrewsomething:
Again, great work @Permagate! This will be very useful for a lot of folks.

One small thing: It's not immediately clear how multiple sorts are intended to interact. Could we get test cases exercising that and perhaps a note in the docs?

Permagate:
@andrewsomething Multiple sorts should be handled in the order it is defined. For example, this will sort by memory in ascending order, then sort by disk in descending order between sizes with same memory:

```hcl
data "digitalocean_sizes" "main" {
  sort {
    // Sort by memory ascendingly
    key       = "memory"
    direction = "asc"
  }

  sort {
    // Then sort by disk descendingly for sizes with same memory
    key       = "disk"
    direction = "desc"
  }
}
```

That was what I intended, however I found out that my implementation didn't quite match what I had in mind. So I added these 2 commits today (in addition to rebasing):

https://github.com/terraform-providers/terraform-provider-digitalocean/pull/325/commits/cbaf381e37597ebb27b6b13717c6ea4c934bdf17 Change the schema type for sort to List instead of Set, since the order it is defined should matter.
https://github.com/terraform-providers/terraform-provider-digitalocean/pull/325/commits/f8f050fa13b0bc090028be520e6ef28f6592e896 Rework the sorting logic so that it actually does what I described above. Previously only the last defined sort applies.

andrewsomething:
:+1: Nice! That's exactly what I had in mind.

Thanks again for all your great work in this PR and the others. We'll get a new release cut soon including all of this.

Issue #222, opened 2019-04-25
Add digitalocean_database and digitalocean_user support on digitalocean_database_cluster resource


pjanuario:
**New Resource(s)** 
digitalocean_database 
digitalocean_user

**Details**
DigitalOcean released a new managed database service. It would be nice to manage that resouce, some work was already done by adding the new resource digitalocean_database_cluster, would be good if we could manage database creation and user configuration via Terraform.

This issue is related with https://github.com/terraform-providers/terraform-provider-digitalocean/issues/189

Yamilquery:
Is there a way to create databases and users with terraform provisionally?

I'm really interested in this feature.

ehamberg:
@Yamilquery There is a [PostgreSQL provider](https://www.terraform.io/docs/providers/postgresql/index.html) that will let you create databases and roles. (You'll quickly find yourself in the chicken-and-egg situation described in https://github.com/hashicorp/terraform/issues/2430 where your PostgreSQL provider will depend on a value from the database cluster resource, though.)

pjanuario:
DO already have API's available for this:
https://developers.digitalocean.com/documentation/v2/#add-a-database-user
https://developers.digitalocean.com/documentation/v2/#add-a-new-database

pjanuario:
@ehamberg I am fine using the Postgres but not sure if it is the best way to perform those actions and even if they will work, but at first glance it seems a good and workable approach since, this goes a bit more into the context of the actual application. I will give it a try and see if it works.

Permagate:
Let me try tackling this, seems simple.

Permagate:
Done, please have a look at my PRs:

#327: `digitalocean_database`
#328: `digitalocean_database_user`

andrewsomething:
Closed via #327 and #328

pjanuario:
Just for future reference, in my case i ended in using the postgres provider since we can have control to do other stuff too.

PR #327, opened 2019-10-28
Add resource for database

Permagate:
Implement `digitalocean_database` as per #222.

What do you think about the resource name? In API and godo, `Database` represents the database cluster while `DB` represents the database itself.

### Example Usage

```hcl
resource "digitalocean_database" "database-example" {
  cluster_id = "${digitalocean_database_cluster.postgres-example.id}"
  name       = "foobar"
}

resource "digitalocean_database_cluster" "postgres-example" {
  name       = "example-postgres-cluster"
  engine     = "pg"
  version    = "11"
  size       = "db-s-1vcpu-1gb"
  region     = "nyc1"
  node_count = 1
}
```


andrewsomething:
Again, awesome work! Thank so much for taking these on.

> What do you think about the resource name? In API and godo, Database represents the database cluster while DB represents the database itself.

While a bit redundant, it would be more consistent with other DigitalOcean tooling, e.g. `doctl databases db <get|list|create|delete>`

```
resource "digitalocean_database_db" "database-example" {
  cluster_id = "${digitalocean_database_cluster.postgres-example.id}"
  name       = "foobar"
}
```

Permagate:
@andrewsomething Agreed. I have updated to use `digitalocean_database_db` name instead.

Also added a sneaky typo fix because of copy-paste in import test for `digitalocean_database_user` resource, sorry 😅

andrewsomething:
:shipit:  Thanks again!

Issue #340, opened 2019-12-05
New resource: digitalocean_database_firewall


andrewsomething:
The DigitalOcean API now supports managing a database cluster’s firewall rules (known as “trusted sources” in the control panel):

https://developers.digitalocean.com/documentation/changelog/api-v2/managed-databases-firewall-and-sql-mode-management/

This should be supported in Terraform as well. A potential config may look like:

```hcl
resource "digitalocean_database_firewall" "example" {
  cluster_id = digitalocean_database_cluster.postgres-example.id

  rule {
    type  = "droplet"
    value = digitalocean_droplet.web.id
  }

  rule {
    type  = "ip_addr"
    value = data.aws_instance.public_ip
  }
}
```

eddiezane:
Closed via #348

PR #348, opened 2019-12-13
Add support for Database Firewalls (Fixes :#340).

andrewsomething:
```
$ make testacc TESTARGS='-run=TestAccDigitalOceanDatabaseFirewall'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccDigitalOceanDatabaseFirewall -timeout 120m
=== RUN   TestAccDigitalOceanDatabaseFirewall_importBasic
--- PASS: TestAccDigitalOceanDatabaseFirewall_importBasic (262.07s)
=== RUN   TestAccDigitalOceanDatabaseFirewall_Basic
--- PASS: TestAccDigitalOceanDatabaseFirewall_Basic (237.98s)
=== RUN   TestAccDigitalOceanDatabaseFirewall_MultipleResourceTypes
--- PASS: TestAccDigitalOceanDatabaseFirewall_MultipleResourceTypes (282.36s)
PASS
ok  	github.com/terraform-providers/terraform-provider-digitalocean/digitalocean	782.432s
```

### DIFF
--- a/digitalocean/resource_digitalocean_database_firewall.go
+++ b/digitalocean/resource_digitalocean_database_firewall.go
@@ -93,11 +93,7 @@ func resourceDigitalOceanDatabaseFirewallRead(d *schema.ResourceData, meta inter
 		return fmt.Errorf("Error retrieving DatabaseFirewall: %s", err)
 	}
 
-	if err := d.Set("rule", flattenDatabaseFirewallRules(rules)); err != nil {
-		return fmt.Errorf("[DEBUG] Error setting DatabaseFirewall rule error: %#v", err)
-	}
-
-	return nil
+	return d.Set("rule", flattenDatabaseFirewallRules(rules))
 }
 
 func resourceDigitalOceanDatabaseFirewallUpdate(d *schema.ResourceData, meta interface{}) error {
@@ -157,12 +153,9 @@ func buildDatabaseFirewallRequest(rules []interface{}) godo.DatabaseUpdateFirewa
 		expandedRules = append(expandedRules, &r)
 	}
 
-	req := godo.DatabaseUpdateFirewallRulesRequest{
+	return godo.DatabaseUpdateFirewallRulesRequest{
 		Rules: expandedRules,
 	}
-
-	return req
-
 }
 
 func flattenDatabaseFirewallRules(rules []godo.DatabaseFirewallRule) []interface{} {
--- a/digitalocean/resource_digitalocean_database_firewall.go
+++ b/digitalocean/resource_digitalocean_database_firewall.go
@@ -1,0 +1,186 @@
+package digitalocean
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/digitalocean/godo"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+)
+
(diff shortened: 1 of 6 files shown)


digitalocean/resource_digitalocean_database_firewall.go
```diff
@@ -0,0 +1,186 @@
...
+	client := meta.(*CombinedConfig).godoClient()
+	clusterID := d.Get("cluster_id").(string)
+
+	rules, _, err := client.Databases.GetFirewallRules(context.TODO(), clusterID)
+	if err != nil {
+		return fmt.Errorf("Error retrieving DatabaseFirewall: %s", err)
+	}
+
+	if err := d.Set("rule", flattenDatabaseFirewallRules(rules)); err != nil {
+		return fmt.Errorf("[DEBUG] Error setting DatabaseFirewall rule error: %#v", err)
```
zachgersh: Could we maybe just return directly from this set (Since it is the line with the last error) instead of a custom error message? Wouldn't the set failure be clear enough?

digitalocean/resource_digitalocean_database_firewall.go
```diff
@@ -0,0 +1,186 @@
...
+
+func resourceDigitalOceanDatabaseFirewallUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*CombinedConfig).godoClient()
+	clusterID := d.Get("cluster_id").(string)
+
+	rules := buildDatabaseFirewallRequest(d.Get("rule").(*schema.Set).List())
+
+	_, err := client.Databases.UpdateFirewallRules(context.TODO(), clusterID, &rules)
+	if err != nil {
+		return fmt.Errorf("Error updating DatabaseFirewall: %s", err)
```
zachgersh: nit: errors really shouldn't be uppercase 👯‍♂

digitalocean/resource_digitalocean_database_firewall.go
```diff
@@ -0,0 +1,186 @@
...
+	return nil
+}
+
+func resourceDigitalOceanDatabaseFirewallUpdate(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*CombinedConfig).godoClient()
+	clusterID := d.Get("cluster_id").(string)
+
+	rules := buildDatabaseFirewallRequest(d.Get("rule").(*schema.Set).List())
+
+	_, err := client.Databases.UpdateFirewallRules(context.TODO(), clusterID, &rules)
```
zachgersh: Should this context actually have a timeout on it? I feel like I've seen providers do that.

digitalocean/resource_digitalocean_database_firewall.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+		if rule["uuid"].(string) != "" {
+			r.UUID = rule["uuid"].(string)
+		}
+
+		expandedRules = append(expandedRules, &r)
+	}
+
+	req := godo.DatabaseUpdateFirewallRulesRequest{
```
zachgersh: nit: return this directly instead of assigning.

digitalocean/resource_digitalocean_database_firewall.go
andrewsomething: Something to think about. It could make sense to add a configurable global timeout. Not sure that I'd want to hardcode something specific to this one resource.

zachgersh: Configurable timeout probably a nice idea! Doesn't necessarily have to be done in this PR but I'd personally start to add timeouts to operations that may take a long time and have a context that can cancel.

Issue #354, opened 2020-01-02
Provider requires token on "terraform validate"


embik:
### Terraform Version
```
$ terraform -v
Terraform v0.12.18
+ provider.digitalocean v1.12.0

```
### Affected Resource(s)
Terraform core functionality

### Terraform Configuration Files
Any Terraform snippet that uses the DigitalOcean provider.
```hcl
data "digitalocean_kubernetes_cluster" "example" {
  name = "prod-cluster-01"
}
```

### Debug Output
https://gist.github.com/embik/328d5b1038f93acd11dda8e78d1afc68

### Panic Output
`-`

### Expected Behavior
`terraform validate` is successful in validating the Terraform configuration. The [documentation for validate](https://www.terraform.io/docs/commands/validate.html) states that `terraform validate` is _"not accessing any remote services such as remote state, provider APIs, etc"_. It therefore should not require an API token. For the sake of security and separation of concerns, I don't want my DigitalOcean API token available where not necessary.

In fact, this is validating a Terraform module, which in itself should not feature a `provider` block at all (this is usually set by a Terraform configuration that imports the module).

### Actual Behavior
It failed because the provider requires a token:

```
$ terraform validate .

Error: Missing required argument

The argument "token" is required, but was not set.
```

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform init`
1. `terraform validate`

### Important Factoids
`-`

### References
`-`

PR #356, opened 2020-01-10
provider: Mark API token as optional (Fixes: #354).

andrewsomething:
This resolves the issue reported in #354 in a way that is consistent with other providers. It has the added benefit of now allowing Spaces only usage.


rayrutjes:
Can we get this released? It is also making `terraform import` commands fail.

PR #366, opened 2020-02-03
add droplet ID to node pools and droplet datasource

tdyas:
Add `droplet_id` attribute to the nodes in a node pool of a Kubernetes cluster. And then allow lookups of droplets by ID for `data.digitalocean_droplet`. This seemed a more robust way of looking up full information on a droplet in a node pool since it was unclear to me whether `name` is guaranteed to be unique within an account. (In my use case, getting the public IP of a node in a node pool to assign to a DNS record.)

### DIFF
--- a/digitalocean/datasource_digitalocean_droplet.go
+++ b/digitalocean/datasource_digitalocean_droplet.go
@@ -15,7 +15,12 @@ func dataSourceDigitalOceanDroplet() *schema.Resource {
 	return &schema.Resource{
 		Read: dataSourceDigitalOceanDropletRead,
 		Schema: map[string]*schema.Schema{
-
+			"id": {
+				Type:         schema.TypeInt,
+				Optional:     true,
+				Description:  "id of the droplet",
+				ValidateFunc: validation.NoZeroValues,
+			},
 			"tag": {
 				Type:         schema.TypeString,
 				Optional:     true,
@@ -174,7 +179,15 @@ func dataSourceDigitalOceanDropletRead(d *schema.ResourceData, meta interface{})
 		opts.Page = page + 1
 	}
 
-	if v, ok := d.GetOk("tag"); ok {
+	if v, ok := d.GetOk("id"); ok {
+		droplet, err := findDropletById(dropletList, v.(int))
+
+		if err != nil {
+			return err
+		}
+
+		exportDropletProperties(d, droplet)
+	} else if v, ok := d.GetOk("tag"); ok {
 		droplet, err := findDropletByTag(dropletList, v.(string))
 
 		if err != nil {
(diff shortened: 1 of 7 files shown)


digitalocean/datasource_digitalocean_droplet.go
```diff
@@ -174,7 +179,15 @@ func dataSourceDigitalOceanDropletRead(d *schema.ResourceData, meta interface{})
 		opts.Page = page + 1
 	}
 
-	if v, ok := d.GetOk("tag"); ok {
+	if v, ok := d.GetOk("id"); ok {
+		droplet, err := findDropletById(dropletList, v.(int))
```
andrewsomething: While Droplet names are not guaranteed to be unique, their IDs are. So there is no need to loop through them all if we have the ID. We can just get it using the ID:

https://godoc.org/github.com/digitalocean/godo#DropletsServiceOp.Get

tdyas: That makes sense; then the ID lookup would short-circuit the need to even do the earlier `for` loop. I'll make the change.

tdyas:
@andrewsomething: I had a question about the data type (string vs integer) for the droplet ID. The droplet API has the [ID as an integer](https://github.com/digitalocean/godo/blob/fce6cf4fb4973e25377d71d291f8613d3d3fdfdc/droplets.go#L43), but the node pool API [returns a string](https://github.com/digitalocean/godo/blob/fce6cf4fb4973e25377d71d291f8613d3d3fdfdc/kubernetes.go#L320). Should the code convert the `droplet_id` attribute on a node in a node pool into an integer?

digitalocean/datasource_digitalocean_droplet.go
tdyas: Change made.

tdyas:
@andrewsomething: What about the `droplet_id` string issue?

andrewsomething:
@tdyas - Was going to comment on that separately from the review approval.

I think we need to use the API as it is provided. If the API returns a string, we shouldn't coerce it into an integer unnecessarily. The way the values are passed around in Terraform itself, means this functions as expected:

```
resource "digitalocean_kubernetes_cluster" "test" {
  name = "test"
  region = "nyc1"
  version = "1.16.2-do.3"
  node_pool {
    name = "pool"
    size = "s-2vcpu-2gb"
    node_count = 1
  }
}

data "digitalocean_droplet" "example" {
  id = digitalocean_kubernetes_cluster.test.node_pool[0].nodes[0].droplet_id
}

output "droplet_id" {
  value = data.digitalocean_droplet.example.id
}

output "droplet_id_from_node_pool" {
  value = digitalocean_kubernetes_cluster.test.node_pool[0].nodes[0].droplet_id
}
```

Returns:

```
Outputs:

droplet_id = 178932046
droplet_id_from_node_pool = 178932046
```

tdyas:
> I think we need to use the API as it is provided. If the API returns a string, we shouldn't coerce it into an integer unnecessarily. The way the values are passed around in Terraform itself, means this functions as expected:

That makes sense from the standpoint of just mirroring the DigitalOcean API. Although it does seem to be an inconsistency in the API itself. Worth raising with the DO API developers?

andrewsomething:
:+1:  I can flag it with DO folks.

PR #375, opened 2020-02-11
datasource_digitalocean_droplet: use ExactlyOneOf to ensure only one way of specifying droplet

tdyas:
Upgrade terraform-plugin-sdk to gain access to the ExactlyOneOf attribute on `Schema`.


tdyas:
> Can you commit the corresponding changes to the contents of the vendor/ directory?

Sure. Just pushed a commit with output of `go mod vendor`.

PR #380, opened 2020-02-18
add digitalocean_regions and digitalocean_region data sources

tdyas:
Add `digitalocean_regions` and `digitalocean_region` data sources. `digitalocean_regions` allows filtering by whether a region is available and the features present in the region. `digitalocean_region` returns metadata on a single region.

This is intended to supersede https://github.com/terraform-providers/terraform-provider-digitalocean/pull/281.

### DIFF
--- a/digitalocean/datasource_digitalocean_region.go
+++ b/digitalocean/datasource_digitalocean_region.go
@@ -56,33 +56,8 @@ func dataSourceDigitalOceanRegionRead(d *schema.ResourceData, meta interface{}) 
 	}
 
 	d.SetId(resource.UniqueId())
-
-	if err := d.Set("slug", region.Slug); err != nil {
-		return fmt.Errorf("Unable to set `slug` attribute: %s", err)
-	}
-
-	if err := d.Set("name", region.Name); err != nil {
-		return fmt.Errorf("Unable to set `name` attribute: %s", err)
-	}
-
-	if err := d.Set("available", region.Available); err != nil {
-		return fmt.Errorf("Unable to set `available` attribute: %s", err)
-	}
-
-	sizesSet := schema.NewSet(schema.HashString, []interface{}{})
-	for _, size := range region.Sizes {
-		sizesSet.Add(size)
-	}
-	if err := d.Set("sizes", sizesSet); err != nil {
-		return fmt.Errorf("Unable to set `sizes` attribute: %s", err)
-	}
-
-	featuresSet := schema.NewSet(schema.HashString, []interface{}{})
-	for _, feature := range region.Features {
-		featuresSet.Add(feature)
-	}
-	if err := d.Set("features", featuresSet); err != nil {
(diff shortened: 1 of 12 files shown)


tdyas:
@andrewsomething: Should `digitalocean_regions` use the `filter` and `sort` attributes similar to `digitalocean_sizes` data source?

andrewsomething:
@tdyas :+1:  I like the idea of using a similar interface in both data sources and it is a pattern used in other providers.

tdyas:
> I like the idea of using a similar interface in both data sources and it is a pattern used in other providers.

The latest push implements filter/sort on `digitalocean_regions`. Let me know what you think of this pattern. (I have code locally to implement `digitalocean_project` and `digitalocean_projects` along similar lines. There is some opportunity for more code sharing among the filter/sort support for each data source.)

digitalocean/datasource_digitalocean_regions.go
```diff
@@ -0,0 +1,188 @@
...
+
+func dataSourceDigitalOceanRegions() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceDigitalOceanRegionsRead,
+		Schema: map[string]*schema.Schema{
+			"filter": filterSchema(dataSourceDigitalOceanRegionsFilterKeys),
+			"sort":   sortSchema(dataSourceDigitalOceanRegionsSortKeys),
+
+			"regions": {
+				Type:     schema.TypeSet,
```
andrewsomething:
```suggestion
				Type:     schema.TypeList,
```

I think we want to make this a list rather than a set. Since we can sort the list, the order does matter. It also has the benefit of allowing you to reference a specific value more easily as it is indexed predictably. Eg:

```
output "slug" {
  value = data.digitalocean_regions.with-storage.regions.0.slug
}
```

https://www.terraform.io/docs/extend/schemas/schema-types.html#typelist

tdyas: Makes sense. Will implement.

tdyas: Pushed the change plus a test helper.

digitalocean/datasource_digitalocean_region.go
```diff
@@ -0,0 +1,72 @@
...
+func dataSourceDigitalOceanRegionRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*CombinedConfig).godoClient()
+
+	regions, err := getDigitalOceanRegions(client)
+	if err != nil {
+		return fmt.Errorf("Unable to load regions: %s", err)
+	}
+
+	slug, ok := d.GetOk("slug")
+	if !ok || slug == "" {
```
eddiezane: nit: is this necessary since `slug` is required? Could we use a validator here?

tdyas: We can. Pushing fix now.

PR #382, opened 2020-02-19
Make DBaaS version required due to DigitalOcean API changes.

andrewsomething:
Previously, the DigitalOcean API only returned a version number for Postgres clusters. More recently, it has begun returning them for both Redis and MySQL ones as well. Creating a cluster without explicitly setting a version will use DigitalOcean's default. This change makes it so that does not return a diff when the resource is refreshed.

Successful test with this change applied:

```
$ make testacc TESTARGS='-run=TestAccDigitalOceanDatabaseCluster_RedisNoVersion'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccDigitalOceanDatabaseCluster_RedisNoVersion -timeout 120m
=== RUN   TestAccDigitalOceanDatabaseCluster_RedisNoVersion
--- PASS: TestAccDigitalOceanDatabaseCluster_RedisNoVersion (365.08s)
PASS
ok  	github.com/terraform-providers/terraform-provider-digitalocean/digitalocean	365.106s
```

Failure before this change:

```
$ make testacc TESTARGS='-run=TestAccDigitalOceanDatabaseCluster_RedisNoVersion'
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccDigitalOceanDatabaseCluster_RedisNoVersion -timeout 120m
=== RUN   TestAccDigitalOceanDatabaseCluster_RedisNoVersion
--- FAIL: TestAccDigitalOceanDatabaseCluster_RedisNoVersion (363.94s)
    testing.go:640: Step 0 error: After applying this step, the plan was not empty:
        
        DIFF:
        
        DESTROY/CREATE: digitalocean_database_cluster.foobar
          database:             "" => "<computed>"
          engine:               "redis" => "redis"
          host:                 "tf-acc-test-u3trulcdlf-do-user-3100395-0.db.ondigitalocean.com" => "<computed>"
          id:                   "fda631a1-a871-4d19-be4e-27ba8dd40c79" => "<computed>"
          maintenance_window.#: "0" => "0"
          name:                 "tf-acc-test-u3trulcdlf" => "tf-acc-test-u3trulcdlf"
          node_count:           "1" => "1"
          password:             "foo" => "<computed>"
          port:                 "25061" => "<computed>"
          private_host:         "private-tf-acc-test-u3trulcdlf-do-user-3100395-0.db.ondigitalocean.com" => "<computed>"
          private_uri:          "rediss://default:foo@private-tf-acc-test-u3trulcdlf-do-user-3100395-0.db.ondigitalocean.com:25061" => "<computed>"
          region:               "nyc1" => "nyc1"
          size:                 "db-s-1vcpu-1gb" => "db-s-1vcpu-1gb"
          tags.#:               "1" => "1"
          tags.0:               "production" => "production"
          uri:                  "rediss://default:foo@tf-acc-test-u3trulcdlf-do-user-3100395-0.db.ondigitalocean.com:25061" => "<computed>"
          urn:                  "do:dbaas:fda631a1-a871-4d19-be4e-27ba8dd40c79" => "<computed>"
          user:                 "default" => "<computed>"
          version:              "5" => "" (forces new resource)
        
        
        
        STATE:
        
        digitalocean_database_cluster.foobar:
          ID = fda631a1-a871-4d19-be4e-27ba8dd40c79
          provider = provider.digitalocean
          database = 
          engine = redis
          host = tf-acc-test-u3trulcdlf-do-user-3100395-0.db.ondigitalocean.com
          name = tf-acc-test-u3trulcdlf
          node_count = 1
          password = foo
          port = 25061
          private_host = private-tf-acc-test-u3trulcdlf-do-user-3100395-0.db.ondigitalocean.com
          private_uri = rediss://default:foo@private-tf-acc-test-u3trulcdlf-do-user-3100395-0.db.ondigitalocean.com:25061
          region = nyc1
          size = db-s-1vcpu-1gb
          tags.# = 1
          tags.0 = production
          uri = rediss://default:foo@tf-acc-test-u3trulcdlf-do-user-3100395-0.db.ondigitalocean.com:25061
          urn = do:dbaas:fda631a1-a871-4d19-be4e-27ba8dd40c79
          user = default
          version = 5
FAIL
FAIL	github.com/terraform-providers/terraform-provider-digitalocean/digitalocean	363.960s
FAIL
make: *** [GNUmakefile:18: testacc] Error 1
```

### DIFF
--- a/digitalocean/resource_digitalocean_database_cluster.go
+++ b/digitalocean/resource_digitalocean_database_cluster.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/digitalocean/godo"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 )
@@ -41,11 +42,6 @@ func resourceDigitalOceanDatabaseCluster() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
-				// Maintain support for clusters created using default version
-				// that is not set explicitly in a config.
-				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
-					return new == ""
-				},
 			},
 
 			"size": {
@@ -150,21 +146,38 @@ func resourceDigitalOceanDatabaseCluster() *schema.Resource {
 			"tags": tagsSchema(),
 		},
 
-		CustomizeDiff: func(diff *schema.ResourceDiff, v interface{}) error {
-			engine := diff.Get("engine")
-			_, hasEvictionPolicy := diff.GetOk("eviction_policy")
-			_, hasSqlMode := diff.GetOk("sql_mode")
+		CustomizeDiff: customdiff.All(
+			schema.CustomizeDiffFunc(func(diff *schema.ResourceDiff, v interface{}) error {
+				engine := diff.Get("engine")
+				_, hasVersion := diff.GetOk("version")
+				old, _ := diff.GetChange("version")
 
-			if hasSqlMode && engine != "mysql" {
-				return fmt.Errorf("sql_mode is only supported for MySQL Database Clusters")
-			}
+				if !hasVersion {
+					if old != "" {
+						return fmt.Errorf(`The argument "version" is now required. Set the %v version to the value saved to state: %v`, engine, old)
+					}
 
-			if hasEvictionPolicy && engine != "redis" {
-				return fmt.Errorf("eviction_policy is only supported for Redis Database Clusters")
-			}
+					return fmt.Errorf(`The argument "version" is required, but no definition was found.`)
+				}
 
-			return nil
-		},
+				return nil
+			}),
+			schema.CustomizeDiffFunc(func(diff *schema.ResourceDiff, v interface{}) error {
+				engine := diff.Get("engine")
+				_, hasEvictionPolicy := diff.GetOk("eviction_policy")
+				_, hasSqlMode := diff.GetOk("sql_mode")
+
+				if hasSqlMode && engine != "mysql" {
+					return fmt.Errorf("sql_mode is only supported for MySQL Database Clusters")
+				}
+
+				if hasEvictionPolicy && engine != "redis" {
+					return fmt.Errorf("eviction_policy is only supported for Redis Database Clusters")
+				}
+
+				return nil
+			}),
+		),
 	}
 }
 
--- a/digitalocean/resource_digitalocean_database_cluster.go
+++ b/digitalocean/resource_digitalocean_database_cluster.go
@@ -41,7 +41,7 @@ func resourceDigitalOceanDatabaseCluster() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				ForceNew: true,
-				// Maintain support for clusters created using defualt version
+				// Maintain support for clusters created using default version
 				// that is not set explicitly in a config.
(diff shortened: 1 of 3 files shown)


eddiezane:
@vkruoso what are your thoughts on this?

vkruoso:
> @vkruoso what are your thoughts on this?

Im really not an expert on how the terraform provider interface works, but from what I can understand you are preventing the creation of a new resource just because the previous state have empty version, and the refreshed resource have the correct version.

Im not sure there is a clean solution here because when a new version is released in the API, all current deployments will be inconsistent as the default version is not the version that is deployed anymore. If there was (Im not sure there is) a way to simply update the state with the API version if the current version in the state is empty, that would make more sense to me.

Other than that, there is a small typo in the code: ```defualt```.

andrewsomething:
You've got it all just about right. One slight clarification.

> If there was (Im not sure there is) a way to simply update the state with the API version if the current version in the state is empty, that would make more sense to me.

The problem is that _it is_ updated in state. This leads to the configuration file and the state differing which is why Terraform wants to recreate the resource. Without this change, it's the user's configuration file that needs to be updated.

@eddiezane, I think the other option is to change the version attribute to be required. It's probably the more correct approach but would require users who just used the default to update their configuration file manually.

eddiezane:
@andrewsomething do you think `StateFunc` would work here to patch folks?

> 	// StateFunc is a function called to change the value of this before
	// storing it in the state (and likewise before comparing for diffs).
	// The use for this is for example with large strings, you may want
	// to simply store the hash of it.

andrewsomething:
I don't think a `StateFunc` will help. The state doesn't need to be updated, the user's configuration file does.

andrewsomething:
I've done a major reworking of this. Making the `version` required is the way to go here. In order to make that transition less painful for anyone with it not set, I use a `CustomizeDiff` to customize the error messaging. For example, if someone had created a Redis cluster without setting a version, they'll see an error giving them a hint on what should be set:

```
digitalocean_database_cluster.redis-example: Refreshing state... [id=43574524-90d9-4a76-a047-fe397f1aa39c]

Error: 1 error occurred:
	* The argument "version" is now required. Set the redis version to the value saved to state: 5

  on terraform-testing.tf line 11, in resource "digitalocean_database_cluster" "redis-example":
  11: resource "digitalocean_database_cluster" "redis-example" {
```

digitalocean/resource_digitalocean_database_cluster.go
```diff
@@ -38,7 +39,9 @@ func resourceDigitalOceanDatabaseCluster() *schema.Resource {
 			},
 
 			"version": {
-				Type:     schema.TypeString,
+				Type: schema.TypeString,
+				// In practice, this is required
+				// See transitionVersionToRequired CustomizeDiffFunc
 				Optional: true,
```
eddiezane: Did you mean to make this required?

andrewsomething: Setting `Required: true,` here makes it error out before the `CustomizeDiff` functions are ever run. So in `transitionVersionToRequired()` I handle both cases:

```
if !hasVersion {
	if old != "" {
		return fmt.Errorf(`The argument "version" is now required. Set the %v version to the value saved to state: %v`, engine, old)
	}

	return fmt.Errorf(`The argument "version" is required, but no definition was found.`)
}
```

For a new resource, if `version` is not set this error is returned:

> The argument "version" is required, but no definition was found.

If no version is set but there is a diff (e.g. new: "8" -> old: ""), then you get the error with the hint:

> The argument "version" is now required. Set the $engine version to the value saved to state: $version

andrewsomething:
@eddiezane  Sounds good. Updated the comment.

Issue #335, opened 2019-11-13
Add Container Registry support


eddiezane:
Initial support landed in [godo](https://github.com/digitalocean/godo/blob/master/registry.go).

This might be a good first issue for @Zelgius. We can work together on this one if you'd like.

zachgersh:
@eddiezane - do we want to wait for this to come out of beta before we move it into terraform? Or do we just do the work beforehand to be prepared for when it ships? I know @andrewsomething will want some messaging about how this functionality is in beta when we ship it.

eddiezane:
Think the best way forward is to silently add support until it's in open beta for folks to use.

PR #383, opened 2020-02-20
adding Container Registry support

Zelgius:
PR for #335


digitalocean/datasource_digitalocean_container_registry.go
```diff
@@ -0,0 +1,40 @@
...
+		},
+	}
+}
+
+func dataSourceDigitalOceanContainerRegistryRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*CombinedConfig).godoClient()
+
+	reg, response, err := client.Registry.Get(context.Background())
+
+	if err != nil {
```
eddiezane: @andrewsomething should we check if the registry name matches?

digitalocean/resource_digitalocean_container_registry.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+	"log"
+
+	"github.com/digitalocean/godo"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+)
+
+func resourceDigitalOceanContainerRegistry() *schema.Resource {
+	return &schema.Resource{
```
eddiezane: Do we have other resources without an update?

Zelgius: Tag doesn't have an update. Also currently in the beta you can't change the name of the registry so there is nothing to update.

digitalocean/datasource_digitalocean_container_registry.go
```diff
@@ -0,0 +1,40 @@
...
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+)
+
+func dataSourceDigitalOceanContainerRegistry() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceDigitalOceanContainerRegistryRead,
+		Schema: map[string]*schema.Schema{
```
andrewsomething: It might be useful to also export a computed value for the registry's endpoint (e.g. `registry.digitalocean.com/foo`). It would match doctl's behaviour. 

https://github.com/digitalocean/doctl/blob/703436d0e4ad535eb31b00bd34c0e599eec9a335/do/registry.go#L31

andrewsomething:
Nice work!

I do think there are still some things to think through. How do we expect users to work with this resource? For example, how would someone integrate it into their cluster?

I think we should probably expose the credentials as a sensitive attribute. This would allow users to interact with the registry via Terraform. For example, I could imagine someone wanting to do something like:

```hcl
data "digitalocean_container_registry" "foo" {
    name = "foo"
}

provider "docker" {
  host = "tcp://localhost:2376"

  registry_auth {
    address = data.digitalocean_container_registry.foo.endpoint
    config_file = data.digitalocean_container_registry.foo.docker-credentials
  }
}

data "docker_registry_image" "app" {
  name = format("%s/app", data.digitalocean_container_registry.foo.endpoint)
}

resource "docker_image" "ubuntu" {
  name          = "${data.docker_registry_image.app.name}"
  pull_triggers = ["${data.docker_registry_image.app.sha256_digest}"]
}
```

https://www.terraform.io/docs/providers/docker/

digitalocean/datasource_digitalocean_container_registry.go
tdyas: Maybe also expose godo's `DockerConfigJSON` for the registry `? That would allow using the Docker config in a provisioner or another resource.

tdyas:
> I think we should probably expose the credentials as a sensitive attribute. This would allow users to interact with the registry via Terraform. For example, I could imagine someone wanting to do something like:

There is also a use case for this in configuring a Kubernetes cluster to be able to pull images from Container Registry (assuming that the DigitalOcean is not already going to automatically configure that on DOKS clusters).

Zelgius:
The request for adding `endpoint` and `docker_credentials` have been added. This surfaced a peculiarity within the API. The API returns credentials that are for the whole registry.digitalocean.com domain, whereas the endpoint returns registry.digitalocean.com/my_registry. After discussing with @andrewsomething we believe the endpoint should have the registry name in the URL. For now we are thinking to require people to specify "registry.digitalocean.com" as their address in the docker or k8s providers. This will be very well documented with an example (I'll do that commit in a bit) so it shouldn't be to jarring. Something similar to this
```
provider "docker" {
  host = "tcp://ubuntu-dev.zelgius.codes:2376"

  registry_auth {
    address = "registry.digitalocean.com"
    config_file_content = data.digitalocean_container_registry.registry.docker_credentials
  }
}
```
The other option is to add another attribute called `registry_address` and have it surface the const that holds `registry.digitalocean.com` up. 

We may need to discuss internally if this is the API response we think is best moving forward. But for now it's what we get and have to program accordingly.

Zelgius:
I added in a `server_url` to the datasource attribute so the docker provider can use the data source instead of hard coding the address
```
provider "docker" {
  host = "tcp://ubuntu-dev.zelgius.codes:2376"

  registry_auth {
    address = data.digitalocean_container_registry.registry.server_url
    config_file_content = data.digitalocean_container_registry.registry.docker_credentials
  }
}
```

digitalocean/datasource_digitalocean_container_registry.go
andrewsomething: The current implementation in godo assumes that there will always be only one registry per account. Checking for the name might help safeguard us from a future change in behaviour, but that is going to be tricky either way.

digitalocean/datasource_digitalocean_container_registry.go
```diff
@@ -0,0 +1,78 @@
...
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				Description:  "name of the container registry",
+				ValidateFunc: validation.NoZeroValues,
+			},
+			//TODO: Need better name?
+			"write": {
```
andrewsomething: Maybe `read_write`? That matches the actual query parameter used and makes the context a bit more clear.

https://developers.digitalocean.com/documentation/v2/#get-docker-credentials-for-container-registry

digitalocean/datasource_digitalocean_container_registry.go
```diff
@@ -0,0 +1,78 @@
...
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+			"endpoint": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"server_url": {
```
andrewsomething: Bit of a nitpick, but since this isn't actually a URL, maybe go with something like `host`?

digitalocean/datasource_digitalocean_container_registry.go
eddiezane: I suggested `write` because I was initially confused by the real parameter name. I think others might be as well. Don't have strong feelings though.

go.sum
```diff
@@ -407,6 +407,7 @@ github.com/hashicorp/terraform-plugin-sdk v1.1.1/go.mod h1:NuwtLpEpPsFaKJPJNGtMc
 github.com/hashicorp/terraform-plugin-sdk v1.4.1 h1:REgN6WbySD6aIYdF6Uob3ic4eQkfh4NXSWU/casmgb4=
 github.com/hashicorp/terraform-plugin-sdk v1.6.0 h1:Um5hsAL7kKsfTHtan8lybY/d03F2bHu4fjRB1H6Ag4U=
 github.com/hashicorp/terraform-plugin-sdk v1.6.0/go.mod h1:H5QLx/uhwfxBZ59Bc5SqT19M4i+fYt7LZjHTpbLZiAg=
+github.com/hashicorp/terraform-plugin-sdk v1.7.0 h1:B//oq0ZORG+EkVrIJy0uPGSonvmXqxSzXe8+GhknoW0=
```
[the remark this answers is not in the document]
eddiezane: I can help you fix this if you get stuck with funkiness. Assuming the tools did this for you. Let me know.

digitalocean/datasource_digitalocean_container_registry.go
Zelgius: We went with the key that shows up in `docker.json`. We could change it to `host` though. 
```
{
	"ServerURL": "https://index.docker.io/v1",
	"Username": "david",
	"Secret": "passw0rd1"
}
```

go.sum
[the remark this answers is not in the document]
Zelgius: Ya. I didn't bring anything in. Probably VSCode doing it.

digitalocean/datasource_digitalocean_container_registry.go
eddiezane: I'd keep the current name if that's the spec.

andrewsomething: Interesting... That's a very different format than what is returned by our API: https://developers.digitalocean.com/documentation/v2/#get-docker-credentials-for-container-registry

Zelgius: So if you're using a credential store for docker is looks like what I posted https://docs.docker.com/engine/reference/commandline/login/#login-to-a-self-hosted-registry#credentials-store but the regular `.docker/config.json` looks like what our API returns. I'm not married to the attribute name, do maybe `docker_host`?

Zelgius: Also Kamal added export for this variable in `godo` https://github.com/digitalocean/godo/pull/306 so I'll be changing it to use this as well.

digitalocean/resource_digitalocean_container_registry.go
```diff
@@ -0,0 +1,84 @@
...
+		Create: resourceDigitalOceanContainerRegistryCreate,
+		Read:   resourceDigitalOceanContainerRegistryRead,
+		Delete: resourceDigitalOceanContainerRegistryDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
```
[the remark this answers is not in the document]
Zelgius: Done :)

digitalocean/datasource_digitalocean_container_registry_test.go
```diff
@@ -0,0 +1,71 @@
...
+	regName := fmt.Sprintf("foo-%s", acctest.RandString(10))
+
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheck(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(testAccCheckDataSourceDigitalOceanContainerRegistryConfig_basic, regName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckDataSourceDigitalOceanContainerRegistryExists("data.digitalocean_tag.foobar", &reg),
```
andrewsomething:
```suggestion
					testAccCheckDataSourceDigitalOceanContainerRegistryExists("data.digitalocean_container_registry.foobar", &reg),
```

digitalocean/import_digitalocean_container_registry_test.go
```diff
@@ -0,0 +1,28 @@
...
+		CheckDestroy: testAccCheckDigitalOceanTagDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckDigitalOceanContainerRegistryConfig_basic,
+			},
+
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
```
andrewsomething: This test is failing for me:

```
=== RUN   TestAccDigitalOceanContainerRegistry_importBasic
--- FAIL: TestAccDigitalOceanContainerRegistry_importBasic (10.02s)
    testing.go:640: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) (len=1) {
         (string) (len=18) "docker_credentials": (string) (len=223) "{\"auths\":{\"registry.digitalocean.com\":{\"auth\":\"asupersecretauthstringthatihaveremoved\"}}}"
        }
        
        
        (map[string]string) (len=2) {
         (string) (len=18) "docker_credentials": (string) (len=223) "{\"auths\":{\"registry.digitalocean.com\":{\"auth\":\"adifferentsupersecretauthstringthatihaveremoved\"}}}",
         (string) (len=5) "write": (string) (len=5) "false"
        }
```

If the API returns a new set of credentials each time it is called, then we might need to use:

```
ImportStateVerifyIgnore: []string{"docker_credentials"},
```

Since we're using `schema.ImportStatePassthrough` the response gets passed through `resourceDigitalOceanContainerRegistryRead`. That function needs to set the value for `write` to state. It's a bit odd case since it's not a value returned by the API, but it needs to be able to refresh the entire resource with `d.Id()` only.

https://www.terraform.io/docs/extend/resources/import.html#importer-state-function

digitalocean/resource_digitalocean_container_registry_test.go
```diff
@@ -0,0 +1,94 @@
...
+		*reg = *foundReg
+
+		return nil
+	}
+}
+
+var testAccCheckDigitalOceanContainerRegistryConfig_basic = fmt.Sprintf(`
+resource "digitalocean_container_registry" "foobar" {
+    name = "foobar"
+}`)
```
andrewsomething:
```suggestion
var testAccCheckDigitalOceanContainerRegistryConfig_basic = `
resource "digitalocean_container_registry" "foobar" {
    name = "foobar"
}`

var testAccCheckDigitalOceanContainerRegistryConfig_write = `
resource "digitalocean_container_registry" "foobar" {
    name  = "foobar"
    write = true
}`
```

No need to use `fmt.Sprintf` here as we are not formatting the string.

Also adding the config for the update step.

digitalocean/resource_digitalocean_container_registry_test.go
```diff
@@ -0,0 +1,94 @@
...
+		CheckDestroy: testAccCheckDigitalOceanContainerRegistryDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckDigitalOceanContainerRegistryConfig_basic,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckDigitalOceanContainerRegistryExists("digitalocean_container_registry.foobar", &reg),
+					testAccCheckDigitalOceanContainerRegistryAttributes(&reg),
+					resource.TestCheckResourceAttr(
+						"digitalocean_container_registry.foobar", "name", "foobar"),
+				),
```
andrewsomething:
```suggestion
					resource.TestCheckResourceAttr(
						"digitalocean_container_registry.foobar", "write", "false"),
					resource.TestCheckResourceAttrSet(
						"digitalocean_container_registry.foobar", "docker_credentials"),
					resource.TestCheckResourceAttr(
						"digitalocean_container_registry.foobar", "endpoint", "registry.digitalocean.com/foobar"),
				),
			},
			{
				Config: testAccCheckDigitalOceanContainerRegistryConfig_write,
				Check: resource.ComposeTestCheckFunc(
					testAccCheckDigitalOceanContainerRegistryExists("digitalocean_container_registry.foobar", &reg),
					testAccCheckDigitalOceanContainerRegistryAttributes(&reg),
					resource.TestCheckResourceAttr(
						"digitalocean_container_registry.foobar", "name", "foobar"),
					resource.TestCheckResourceAttr(
						"digitalocean_container_registry.foobar", "write", "true"),
				),
			},
```

It would be great to test the update path as well as make sure other attributes are being set.

website/docs/d/container_registry.html.md
```diff
@@ -0,0 +1,56 @@
...
+  }
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the container registry.
+* `write` - (Optional) Boolean  to retrieve read/write credentials, suitable for use with the Docker client or in a CI system.
```
andrewsomething: Make sure to mention the default here.

website/docs/r/container_registry.html.markdown
```diff
@@ -0,0 +1,47 @@
...
+  name = "foobar"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the container_registry
+* `write` - (Optional) Boolean  to retrieve read/write credentials, suitable for use with the Docker client or in a CI system.
```
andrewsomething: Mention the default here.

website/docs/d/container_registry.html.md
```diff
@@ -0,0 +1,56 @@
...
+Get information on a container registry. This data source provides the name as 
+configured on your DigitalOcean account. This is useful if the container 
+registry name in question is not managed by Terraform or you need validate if 
+the container registry exists in the account.
+
+An error is triggered if the provided container registry name does not exist.
+
+## Example Usage
+
+Get the container registry:
```
andrewsomething:
```suggestion
### Basic Example

Get the container registry:
```

website/docs/d/container_registry.html.md
```diff
@@ -0,0 +1,56 @@
...
+
+```hcl
+data "digitalocean_container_registry" "example" {
+  name = "example"
+}
+```
+
+Use the endpoint and docker_credentials in the docker provider
```
andrewsomething:
```suggestion
### Docker Provider Example

Use the `endpoint` and `docker_credentials` with the Docker provider:
```

website/docs/d/container_registry.html.md
```diff
@@ -0,0 +1,56 @@
...
+
+```
+provider "docker" {
```
andrewsomething:
```suggestion
data "digitalocean_container_registry" "example" {
  name = "example"
}

provider "docker" {
```

website/docs/d/container_registry.html.md
```diff
@@ -0,0 +1,56 @@
...
+  host = "tcp:localhost:2376"
+
+  registry_auth {
+    address = data.digitalocean_container_registry.example.server_url
+    config_file_content = data.digitalocean_container_registry.example.docker_credentials
+  }
+}
+```
+
```
andrewsomething: I think it would be useful to also provide a full example for the K8s use case. Something like:
(The suggestion feature doesn't seem to like markdown inside of markdown...)

### Kubernetes Example

Combined with the Kubernetes Provider's [`kubernetes_secret` resource](https://www.terraform.io/docs/providers/kubernetes/r/secret.html), you can access the registry from inside your cluster:

```
data "digitalocean_container_registry" "example" {
  name = "example"
}

resource "digitalocean_kubernetes_cluster" "example" {
  name    = "example"
  region  = "nyc1"
  # Grab the latest version slug from `doctl kubernetes options versions`
  version = "1.16.6-do.0"

  node_pool {
    name       = "worker-pool"
    size       = "s-2vcpu-2gb"
    node_count = 3
  }
}

provider "kubernetes" {
  load_config_file = false
  host  = digitalocean_kubernetes_cluster.example.endpoint
  token = digitalocean_kubernetes_cluster.example.kube_config[0].token
  cluster_ca_certificate = base64decode(
    digitalocean_kubernetes_cluster.example.kube_config[0].cluster_ca_certificate
  )
}

resource "kubernetes_secret" "example" {
  metadata {
    name = "docker-cfg"
  }

  data = {
    ".dockerconfigjson" = data.digitalocean_container_registry.example.docker_credentials
  }

  type = "kubernetes.io/dockerconfigjson"
}
```

Zelgius: Done and committed.

website/docs/d/container_registry.html.md
```diff
@@ -0,0 +1,64 @@
+---
+layout: "digitalocean"
+page_title: "DigitalOcean: digitalocean_container_registry"
+sidebar_current: "docs-do-datasource-container-registry"
+description: |-
+  Get information on a container registry.
+---
+
+# digitalocean_container_registry
+
```
Zelgius:
```suggestion
**Note**: DigitalOcean Container Registry is currently in private beta. This feature will become available to the general public soon.
```

andrewsomething: I've notice some providers have callout boxes, eg https://www.terraform.io/docs/providers/google/index.html

Looking at the source, it seems to be some custom markup:

```
-> **Note**: DigitalOcean Container Registry is currently in private beta. This feature will become available to the general public soon.
```

You can preview it with `make website`

digitalocean/datasource_digitalocean_container_registry.go
eddiezane: @andrewsomething any further preference here?

andrewsomething: No strong preference. Mostly made a suggestion because of the `TODO` comment.

eddiezane: @Zelgius I prefer `write` because `read_write` isn't as clear as what it does since read is true by default. Feel free to make the call and remove the `TODO`.

website/docs/d/container_registry.html.md
eddiezane: We can add this to the resource as a warning about being in beta?

```
	// If non-empty, this string is emitted as a warning during Validate.
	DeprecationMessage string
```

Zelgius: I added the note @andrewsomething recommended.

Zelgius:
@andrewsomething @eddiezane How do you suggest going about this? 
> * We need some approach to caching the Docker credentials. Without that, the K8s use case is very painful. As it returns a different set of creds each time, Terraform always sees a diff for a `kubernetes_secret` resource that references the registry resource. This is a similar problem to kubeconfig files, but that helpfully includes an `expires_at` field. See: https://github.com/terraform-providers/terraform-provider-digitalocean/blob/f977288470bc3e9d71f811a4af4ef6020e714296/digitalocean/resource_digitalocean_kubernetes_cluster.go#L274

website/docs/r/container_registry_docker_credentials.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+data "digitalocean_container_registry" "example" {
+  name = "example"
+}
+
+resource "digitalocean_container_registry_docker_credentials" "example" {
+  name = "example"
+}
+
+provider "docker" {
+  host = "tcp:localhost:2376"
```
andrewsomething, line 41: Testing this, I get:

> Error initializing Docker client: unable to parse docker host `tcp:127.0.0.1:2376/`

The Docker provider docs use: `tcp://localhost:2376` which gets past that error. On Ubuntu, I actually needed: `unix:///var/run/docker.sock`

website/docs/r/container_registry_docker_credentials.html.markdown
```diff
@@ -0,0 +1,111 @@
...
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) The name of the container registry.
+* `write` - (Optional) Allow for write access to the container registry. Defaults to false.
+* `expiry_secons` - (Optional) The amount of time to pass before the docker credentials expire. Defaults to 2147483647, or roughly 68 years. Must be greater than 0 and less than 2147483647.
```
andrewsomething, line 101:
```suggestion
* `expiry_seconds` - (Optional) The amount of time to pass before the Docker credentials expire in seconds. Defaults to 2147483647, or roughly 68 years. Must be greater than 0 and less than 2147483647.
```

eddiezane:
@Zelgius great job on this, dude!

I know it was a long road but you pulled it of stunningly!

PR #385, opened 2020-02-24
generic filter/sort framework for datasources

tdyas:
This PR adds a generic filter/sort framework for datasources and ports the regions and sizes datasources to it. (I thought it a good idea to make the filter/sort support generic before I submit  PRs to add some additional datasources which would duplicate a lot of the logic from the regions and sizes).


digitalocean/datalist/schema.go
```diff
@@ -0,0 +1,75 @@
+package datalist
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+type DataListResourceConfig struct {
```
andrewsomething: Exported types and functions should have comments. That's generally good practice, but I think it's especially important in this case. We want anyone familiar with Terraform providers to be able to pick up this code base quickly. As this is something custom to this provider, documenting it is important.

Also, since this will be called as `datalist.DataListResourceConfig`, we might want to consider renaming it `ResourceConfig` so it is less repetitive.

digitalocean/datalist/schema.go
```diff
@@ -0,0 +1,75 @@
...
+	RecordSchema        map[string]*schema.Schema
+	FilterKeys          []string
+	SortKeys            []string
+	ResultAttributeName string
+	FlattenRecord       func(record interface{}) map[string]interface{}
+	GetRecords          func(meta interface{}) ([]interface{}, error)
+}
+
+func NewDataListResource(config *DataListResourceConfig) *schema.Resource {
```
andrewsomething: Same as above, re: comments.

digitalocean/datalist/schema.go
tdyas: Added comments and renamed the config struct. Let me know if you would like me to fill in more details.

tdyas:
> One more general point: I I think it would probably make sense to turn this into an internal package in order to not export the API. Thoughts? See: https://golang.org/doc/go1.4#internalpackages

Makes sense. I could see other Terraform providers using this code eventually, but it should be extracted out into its own library at that point.

digitalocean/datalist/schema.go
tdyas: Updated as requested.

tdyas:
@andrewsomething: This is ready for review again. I even used it as a base for https://github.com/terraform-providers/terraform-provider-digitalocean/pull/391 (`digitalocean_projects` datasource).

Note: I chose to panic if the `datalist` is not configured properly since there does not seem to be a way to thread an `error` back up through the plugin initialization calls.

PR #386, opened 2020-02-25
Require that volume names be lowercase and alphanumeric

danrabinowitz:
I tried to create a volume. Terraform failed with:
```
Error: Error creating Volume: POST https://api.digitalocean.com/v2/volumes: 422 (request "a1952cf3-2088-4e8d-802d-xxxxx") failed to create volume: invalid volume name. names must be lowercase and alphanumeric
```

### DIFF
--- a/digitalocean/resource_digitalocean_volume.go
+++ b/digitalocean/resource_digitalocean_volume.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"regexp"
 	"strings"
 
 	"github.com/digitalocean/godo"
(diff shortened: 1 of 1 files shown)


digitalocean/resource_digitalocean_volume.go
```diff
@@ -36,7 +37,7 @@ func resourceDigitalOceanVolume() *schema.Resource {
 				Type:         schema.TypeString,
 				Required:     true,
 				ForceNew:     true,
-				ValidateFunc: validation.NoZeroValues,
+				ValidateFunc: validation.StringMatch(regexp.MustCompile(`^[a-z0-9]+$`), "Names must be lowercase and alphanumeric"),
```
eddiezane: There might be a validator that already does this. If not works for me.

danrabinowitz: Good point. I just checked and it looks like there isn't one.

PR #394, opened 2020-03-07
add digitalocean_images datasource

tdyas:
Use the datalist library to add a `digitalocean_images` datasource. Also update the `digitalocean_image` datasource to use the new common code.

This should supersede https://github.com/terraform-providers/terraform-provider-digitalocean/pull/79, although an open question is whether there should be a `type` attribute like in the DO API for applications and distributions.

### DIFF
--- a/digitalocean/datasource_digitalocean_image.go
+++ b/digitalocean/datasource_digitalocean_image.go
@@ -11,56 +11,23 @@ import (
 )
 
 func dataSourceDigitalOceanImage() *schema.Resource {
-	return &schema.Resource{
-		Read: dataSourceDigitalOceanImageRead,
-		Schema: map[string]*schema.Schema{
+	recordSchema := imageSchema()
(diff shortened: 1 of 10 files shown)


tdyas:
Open question: should distributions and application be broken out somehow in either or both the `digitalocean_image` and `digitalocean_images` datasources? (Trying to see if that idea in https://github.com/terraform-providers/terraform-provider-digitalocean/pull/394 should be pulled forward.)

tdyas:
> So it would probably still be helpful to have an argument that chooses godo.Images.ListUser, godo.Images.ListApplication, godo.Images.ListDistribution. That also brings some performance. Doesn't necessarily need to block

How about `source` as the name of the parameter? Distinct from `type` but still conveys what it does.

andrewsomething:
@tdyas :+1:  Sounds good!

PR #400, opened 2020-03-19
Fix volume name validation

Nevon:
#386 introduced validation that disallows dashes, which is allowed according to DigitalOcean's API documentation: https://developers.digitalocean.com/documentation/v2/#block-storage

We have several volumes that contain dashes, which means that with the latest version of the provider we cannot update our stack anymore, as it considers the volume names invalid. I assume this would have been caught by the acceptance tests, as they have names like `volume-abcdefghij`, but I suppose they must not have been run (or I am reading them wrong. Go is not my main language).


andrewsomething:
Will merge on green, but Travis looks like it's having a hard time today.

MattIPv4:
Hey, @Nevon - Thanks a ton for this sweet PR! 😄

PR #404, opened 2020-03-26
support volumes created before name rules changed

BrianHicks:
In 2019, DigitalOcean accepted volumes created with underscores in the name. In 2020, this is no longer the case but it leaves us without a way to work with volumes created in the old rules.

This PR allows underscores so that I can work with a couple volumes that I had created in October or so, detailed in ~~https://github.com/hashicorp/terraform/issues/24468~~ https://github.com/terraform-providers/terraform-provider-digitalocean/issues/405


andrewsomething:
Thank you for the PR!

MattIPv4:
Hey, @BrianHicks - Thanks a ton for this sweet PR! 😄

PR #406, opened 2020-04-03
support old volumes created with a dot in the name

Amygos:
Previously DigitalOcean allowed also volumes with a dot in the name, this PR make possible to manage also this case.
Related PR #404

### DIFF
--- a/digitalocean/resource_digitalocean_volume.go
+++ b/digitalocean/resource_digitalocean_volume.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"regexp"
 	"strings"
 
 	"github.com/digitalocean/godo"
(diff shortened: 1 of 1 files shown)


andrewsomething:
Hi @Amygos, thanks for the PR! Sorry that this change caused issues for you. Between this, https://github.com/terraform-providers/terraform-provider-digitalocean/pull/400, and https://github.com/terraform-providers/terraform-provider-digitalocean/pull/404, it seems clear that adding this validation has caused more issues than it has solved. At this point, I think we should probably just revert https://github.com/terraform-providers/terraform-provider-digitalocean/pull/386 and allow the DigitalOcean API to do its own validation and return an error if needed.

Amygos:
Hi, @andrewsomething no problems for me to close this PR

MattIPv4:
Hey, @Amygos - Thanks a ton for this sweet PR! 😄

PR #408, opened 2020-04-05
add digitalocean_spaces_bucket_object resource

tdyas:
Adapt the [`aws_s3_bucket_object` resource](https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_s3_bucket_object.go) for the DigitalOcean provider. (This also adapts the relevant acceptance tests for `aws_s3_bucket_object`.)

I removed the following features that are only supported on AWS:

- ACL: Only `private` and `public-read` are supported.
- Encryption / KMS
- Object holds / legal holds
- Storage class: Only STANDARD is supported.
- Tags: Is this supported by DigitalOcean? It was unclear to me so I removed the AWS-style tags support.

`isAWSErr`: Moved this out of a test file and used the version from the AWS provider which uses `errors.As`.

The main addition to `digitalocean_spaces_bucket_object` is the addition of a region attribute to allow access to the bucket.

This still needs documentation and running some of the more specific acceptance tests.


tdyas:
Several tests adapted from S3 have versioning enabled. Thus, https://github.com/terraform-providers/terraform-provider-digitalocean/pull/409 will need to land first.

digitalocean/resource_digitalocean_spaces_bucket_object_test.go
```diff
@@ -0,0 +1,865 @@
...
+
+func testAccDigitalOceanSpacesBucketObjectConfig_NonVersioned(randInt int, source string) string {
+	return fmt.Sprintf(`
+resource "digitalocean_spaces_bucket" "object_bucket_3" {
+  region = "%s"
+  name   = "tf-object-test-bucket-%d"
+}
+
+resource "digitalocean_spaces_bucket_object" "object" {
+  region = digitalocean_spaces_bucket.object_bucket.region
```
andrewsomething, line 804:
```suggestion
  region = digitalocean_spaces_bucket.object_bucket_3.region
```

```
--- FAIL: TestAccDigitalOceanSpacesBucketObject_NonVersioned (0.03s)
    testing.go:640: Step 0 error: config is invalid: Reference to undeclared resource: A managed resource "digitalocean_spaces_bucket" "object_bucket" has not been declared in the root module.
```

andrewsomething:
> Tags: Is this supported by DigitalOcean? It was unclear to me so I removed the AWS-style tags support.

Object tagging is technically supported but lifecycle policies based on object tagging are not. So it is not clear to me if there is any real use case. I'd rather start focused and add more if there is demand.

https://developers.digitalocean.com/documentation/spaces/#aws-s3-compatibility

tdyas:
> Object tagging is technically supported but lifecycle policies based on object tagging are not. So it is not clear to me if there is any real use case. I'd rather start focused and add more if there is demand.

Understood. Makes sense to leave it out for now. An alternate reason to leave it out is that the AWS provider's code relies on an internal library and it would just further complicate this PR than it needs to be to port all that code to this provider.

digitalocean/resource_digitalocean_spaces_bucket_object_test.go
```diff
@@ -0,0 +1,865 @@
...
+  region = "%s"
+  name   = "tf-object-test-bucket-%d"
+
+  versioning {
+    enabled = %t
+  }
+}
+
+resource "digitalocean_spaces_bucket_object" "object" {
+  region = digitalocean_spaces_bucket.object_bucket.region
```
andrewsomething, line 729:
```suggestion
  region = digitalocean_spaces_bucket.object_bucket_3.region
```

Missed this one as the test is failing for other reasons.

digitalocean/resource_digitalocean_spaces_bucket_object_test.go
```diff
@@ -0,0 +1,865 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDigitalOceanSpacesBucketObjectConfig_updateable(rInt, false, sourceInitial),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckDigitalOceanSpacesBucketObjectExists(resourceName, &originalObj),
+					testAccCheckDigitalOceanSpacesBucketObjectBody(&originalObj, "initial object state"),
+					resource.TestCheckResourceAttr(resourceName, "etag", "647d1d58e1011c743ec67d5e8af87b53"),
+					resource.TestCheckResourceAttr(resourceName, "object_lock_legal_hold_status", ""),
+					resource.TestCheckResourceAttr(resourceName, "object_lock_mode", ""),
+					resource.TestCheckResourceAttr(resourceName, "object_lock_retain_until_date", ""),
```
andrewsomething, line 209:
```suggestion
```

digitalocean/resource_digitalocean_spaces_bucket_object_test.go
```diff
@@ -0,0 +1,865 @@
...
+			},
+			{
+				Config: testAccDigitalOceanSpacesBucketObjectConfig_updateable(rInt, false, sourceModified),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckDigitalOceanSpacesBucketObjectExists(resourceName, &modifiedObj),
+					testAccCheckDigitalOceanSpacesBucketObjectBody(&modifiedObj, "modified object"),
+					resource.TestCheckResourceAttr(resourceName, "etag", "1c7fd13df1515c2a13ad9eb068931f09"),
+					resource.TestCheckResourceAttr(resourceName, "object_lock_legal_hold_status", ""),
+					resource.TestCheckResourceAttr(resourceName, "object_lock_mode", ""),
+					resource.TestCheckResourceAttr(resourceName, "object_lock_retain_until_date", ""),
```
andrewsomething, line 220:
```suggestion
```

digitalocean/resource_digitalocean_spaces_bucket_object_test.go
tdyas: Included this as part of rebase.

digitalocean/resource_digitalocean_spaces_bucket_object_test.go
tdyas: Same

tdyas:
@andrewsomething: I pushed a rebase plus incorporated the fixes you had suggested. One remaining issue is the need to set `force_destroy = true` on the test `digialocean_spaces_bucket` since I am seeing this error on some of the tests:

```
--- FAIL: TestAccDigitalOceanSpacesBucketObject_updates (11.78s)
    testing.go:701: Error destroying resource! WARNING: Dangling resources
        may exist. The full state and error is shown below.
        
        Error: errors during apply: Error deleting Spaces Bucket: BucketNotEmpty: 
                status code: 409, request id: , host id:  "tf-object-test-bucket-2850460655607768354"
```

tdyas:
`TF_LOG=DEBUG TF_LOG_PATH=./log.out make testacc TESTARGS='-run=^TestAccDigitalOceanSpacesBucketObject'` now passes with the change to set `force_destroy = true` on all of the `digitalocean_spaces_bucket` resources in the acceptance tests.

PR #409, opened 2020-04-05
digitalocean_spaces_bucket: support bucket versioning

tdyas:
Support enabling bucket versioning via `versioning` configuration similar to the [`versioning` attribute](https://www.terraform.io/docs/providers/aws/r/s3_bucket.html#versioning) of [`aws_s3_bucket`](https://www.terraform.io/docs/providers/aws/r/s3_bucket.html).

### DIFF
--- a/digitalocean/resource_digitalocean_spaces_bucket.go
+++ b/digitalocean/resource_digitalocean_spaces_bucket.go
@@ -189,7 +189,6 @@ func resourceDigitalOceanBucketUpdate(d *schema.ResourceData, meta interface{}) 
 
 func resourceDigitalOceanBucketRead(d *schema.ResourceData, meta interface{}) error {
 	region := d.Get("region").(string)
-	log.Printf("[DEBUG] region = %v", region)
 	client, err := meta.(*CombinedConfig).spacesClient(region)
 
 	if err != nil {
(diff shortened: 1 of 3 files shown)


digitalocean/resource_digitalocean_spaces_bucket.go
```diff
@@ -81,6 +81,24 @@ func resourceDigitalOceanBucket() *schema.Resource {
...
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enabled": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+						},
+					},
+				},
+			},
```
andrewsomething, line 100:
```suggestion
			"versioning": {
				Type:     schema.TypeList,
				Optional: true,
				MaxItems: 1,
				Elem: &schema.Resource{
					Schema: map[string]*schema.Schema{
						"enabled": {
							Type:     schema.TypeBool,
							Optional: true,
							Default:  false,
						},
					},
				},
				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
					return old == "1" && new == "0"
				},
			},
```

In some testing, I think this should do what we want. By removing `Computed`, it will detect the need for an update when removing a block with `versioning.enabled = true` but the `DiffSuppressFunc` will prevent a diff like:

```diff
      - versioning {
          - enabled = false -> null
        }
```

The AWS provider has something similar (though not used with the `s3_bucket` resource for some reason): 

https://github.com/terraform-providers/terraform-provider-aws/blob/98b8b848ca94031b20c3e626c9d40484e3af80de/aws/diff_suppress_funcs.go#L36-L42

tdyas: That should work. I'll update the acceptance test to also test the case of having versioning enabled but going back to not specifying it.

PR #410, opened 2020-04-07
Add support for VPC.

andrewsomething:
https://developers.digitalocean.com/documentation/changelog/api-v2/api-support-for-digitalocean-vpcs/

This adds a new `digitalocean_vpc` resource and data source along with related changes to the Droplet, Kuberenetes cluster, database cluster, and load balancer resources and data sources. 

CC: @viola


website/docs/r/vpc.html.markdown
```diff
@@ -0,0 +1,71 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `name` - (Required) A name for the VPC. Must be unique and contain alphanumeric characters, dashes, and periods only.
+* `region` - (Required) The DigitalOcean region slug for the VPC's location.
+* `description` - (Optional) A free-form text field up to a limit of 255 characters to describe the VPC.
+* `ip_range` - (Optional) The range of IP addresses for the VPC in CIDR notation. Network ranges cannot overlap with other networks in the same account and must be in range of private addresses as defined in RFC1918. It may not be larger than `/16`.
```
viola, line 54: and smaller than `24`

viola:
LGTM left one minor non-blocking comment.

PR #411, opened 2020-04-08
digitalocean_spaces_bucket: add lifecycle_rule support

tdyas:
Adapt [`lifecycle_rule`](https://www.terraform.io/docs/providers/aws/r/s3_bucket.html#lifecycle_rule) from `aws_s3_bucket` for `digitalocean_spaces_bucket`.

This is not ready for inclusion yet, more work needs to be done on adapting the acceptance tests. Putting it up for early review to get feedback on what should be tested and what should not given that DigitalOcean Spaces does not support all of the features that S3 does.


andrewsomething:
> what should be tested and what should not given that DigitalOcean Spaces does not support all of the features that S3 does

Looking this over and speaking with folks closer to Spaces: 

* `Expiration`,  `AbortIncompleteMultipartUpload`, and `NoncurrentVersionExpiration` are all supported.
* `Transition` and `NoncurrentVersionTransition` are not supported and should be removed.

digitalocean/resource_digitalocean_spaces_bucket.go
```diff
@@ -106,6 +110,116 @@ func resourceDigitalOceanBucket() *schema.Resource {
...
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							Computed:     true,
+							ValidateFunc: validation.StringLenBetween(0, 255),
+						},
+						"prefix": {
+							Type:     schema.TypeString,
+							Optional: true,
```
andrewsomething, line 127: Specifying a prefix starting with `/` is almost never wanted and leads to a good bit of user error. Though there are some edge cases where it might be wanted. So we should log a a non fatal error here. 

```suggestion
							Optional: true,
							ValidateFunc: func(v interface{}, k string) (ws []string, es []error) {
								if strings.HasPrefix(v.(string), "/") {
									ws = append(ws, "prefix begins with `/`. In most cases, this should be excluded.")
								}
								return
							},
```

tdyas:
Will implement https://github.com/terraform-providers/terraform-provider-digitalocean/issues/362 once done.

tdyas:
All of the tests pass now except `TestAccDigitalOceanSpacesBucket_LifecycleRule_Expiration_EmptyConfigurationBlock` which fails with:

```
--- FAIL: TestAccDigitalOceanSpacesBucket_LifecycleRule_Expiration_EmptyConfigurationBlock (10.36s)
    testing.go:640: Step 0 error: errors during apply:
        
        Error: Error putting S3 lifecycle: MalformedXML: 
                status code: 400, request id: tx00000000000021519c693-005e9227ae-3359f-nyc3b, host id: 
        
          on /tmp/tf-test359342063/main.tf line 2:
          (source code not available)
```

I has debug logging turned on but the call into the S3 library didn't end up printing any more to the log. Someone more familiar with whether that test is even a valid test of the Spaces API should weigh in first.

tdyas:
Also I had to clean up region handling a little bit in the acceptance test. I originally had not filled in region in the test configs adapted from the AWS provider. However, other parts of the Spaces acceptance tests do set it to ams3. This caused some errors where API calls would not find the bucket because it had nyc3 (the default region) instead of ams3.

This happened in part because the `region` attribute is not marked ForceNew and probably should be marked as such. (The provider's code assumes it can create a S3 client just from the current value of the `region` attribute, which would be correct if the update logic never observes a change to `region` as if ForceNew were set.)

tdyas:
> I think we should remove TestAccDigitalOceanSpacesBucket_LifecycleRule_Expiration_EmptyConfigurationBlock

Sounds good. I've removed it in the latest commits. I also tried to add an `AtLeastOneOf` config to the `date`, `days`, and `expired_object_delete_marker` arguments but `AtLeastOneOf` does not seem to work with sub-arguments and caused a `Internal validation of the provider failed!` error message.

PR #412, opened 2020-04-08
Remove storage_class attribute from spaces_bucket_object.

andrewsomething:
Thinking this over a bit more, `storage_class` as implemented is essentially a no-op right now.

https://github.com/terraform-providers/terraform-provider-digitalocean/blob/master/digitalocean/resource_digitalocean_spaces_bucket_object.go#L300-L305

We should remove it here to not cause confusion around Spaces support for storage classes and protect against any unforeseen ways Spaces might interact with storage classes in the future.

@tdyas, any thoughts here?

### DIFF
--- a/digitalocean/resource_digitalocean_spaces_bucket_object.go
+++ b/digitalocean/resource_digitalocean_spaces_bucket_object.go
@@ -109,15 +109,6 @@ func resourceDigitalOceanSpacesBucketObject() *schema.Resource {
 				ConflictsWith: []string{"source", "content"},
 			},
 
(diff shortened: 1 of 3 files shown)


tdyas:
> We should remove it here to not cause confusion around Spaces support for storage classes and protect against any unforeseen ways Spaces might interact with storage classes in the future.

Makes sense to remove it.

For context, it was left in because I did the minimum necessary to adapt the AWS provider's code by deleting the other choices and did not give thought to whether leaving in a single choice was a good idea or not.

PR #416, opened 2020-04-19
add preliminary data sources for Spaces buckets

tdyas:
This PR adds `digitalocean_spaces_bucket` and `digitalocean_spaces_buckets` data sources. The attributes exported by the data sources do not include any attributes that require additional Spaces API calls to retrieve. (Subsequent PRs will add those since they will involve refactoring the resources and data sources to use common read methods. I didn't want to complicate this PR with those refactorings.)

Open question:
- It was unclear to me how to differentiate regions with Spaces and regions without. The "features" list in godo.Region does not appear to have a "spaces" feature. The closest is "storage" but both nyc1 and nyc3 have that feature, but only nyc3 has Spaces.

### DIFF
--- a/digitalocean/datasource_digitalocean_spaces_bucket.go
+++ b/digitalocean/datasource_digitalocean_spaces_bucket.go
@@ -1,0 +1,101 @@
+package digitalocean
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/s3"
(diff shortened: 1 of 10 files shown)


andrewsomething:
"storage" in the regions payload should refer to block storage volumes. Unfortunately, the Spaces API is mostly distinct. The list here should be canonical:

https://www.digitalocean.com/docs/platform/availability-matrix/#other-product-availability

Hard-coding them is a bit unfortunate, but it does have the benefit of allowing it to be used with only Spaces credentials if someone has a config not accessing other resource types.

tdyas:
> Hard-coding them is a bit unfortunate, but it does have the benefit of allowing it to be used with only Spaces credentials if someone has a config not accessing other resource types.

Okay will hard-code them then with a comment.

PR #417, opened 2020-04-19
fix transitive dependency in go.mod

tdyas:
Go's module support (as invoked by IntelliJ's Go plugin) was complaining about the incorrect use of an "incompatible" version tag on the terraform-provider-random transitive dep (by way of the terraform-provider-google dep). Modify go.mod to replace the version with a pseudo-version tag for that same commit.

With this, IntelliJ is finally able to fully resolve transitive dependencies because its call to `go list -m` succeeds.


andrewsomething:
Thanks @tdyas!

PR #418, opened 2020-04-20
add digitalocean_droplets data source

tdyas:
Adds a `digitalocean_droplets` data source.

Note: I don't have access to the VPC preview so the existing tests for the `digitalocean_droplet` data source fail when I ran those acceptance tests to ensure that the refactoring didn't cause a regression.

### DIFF
--- a/digitalocean/droplets.go
+++ b/digitalocean/droplets.go
@@ -168,7 +168,7 @@ func flattenDigitalOceanDroplet(rawDroplet, meta interface{}) (map[string]interf
 	}
 
 	if droplet.Image.Slug == "" {
-		flattenedDroplet["image"] = droplet.Image.ID
+		flattenedDroplet["image"] = strconv.Itoa(droplet.Image.ID)
 	} else {
 		flattenedDroplet["image"] = droplet.Image.Slug
 	}
--- a/digitalocean/droplets.go
+++ b/digitalocean/droplets.go
@@ -3,6 +3,7 @@ package digitalocean
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/digitalocean/godo"
--- a/digitalocean/droplets.go
+++ b/digitalocean/droplets.go
@@ -1,0 +1,204 @@
+package digitalocean
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/digitalocean/godo"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func dropletSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"id": {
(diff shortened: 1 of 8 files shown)


tdyas:
> Note: I don't have access to the VPC preview so the existing tests for the digitalocean_droplet data source fail when I ran those acceptance tests to ensure that the refactoring didn't cause a regression.

To clarify, some of the acceptance tests use `digitalocean_vpc` in setting up droplets.

digitalocean/droplets.go
```diff
@@ -0,0 +1,204 @@
...
+		"vcpus":         droplet.Vcpus,
+		"memory":        droplet.Memory,
+		"status":        droplet.Status,
+		"locked":        droplet.Locked,
+		"created_at":    droplet.Created,
+		"vpc_uuid":      droplet.VPCUUID,
+	}
+
+	if droplet.Image.Slug == "" {
+		flattenedDroplet["image"] = droplet.Image.ID
```
andrewsomething, line 171:
```suggestion
		flattenedDroplet["image"] = strconv.Itoa(droplet.Image.ID)
```


Working with both the droplet and droplets data sources, I'm running into:

```
Error: Unable to set `image` attribute: image: '' expected type 'string', got unconvertible type 'int'

  on main.tf line 8, in data "digitalocean_droplet" "drop":
   8: data "digitalocean_droplet" "drop" {
```

website/docs/d/droplets.html.md
```diff
@@ -0,0 +1,107 @@
...
+  filter {
+    key = "size"
+    values = ["s-1vcpu-1gb"]
+  }
+  filter {
+    key = "backups"
+    values = ["true"]
+  }
+  sort {
+    key = "created"
```
andrewsomething, line 48:
```suggestion
    key = "created_at"
```

website/docs/d/droplets.html.md
```diff
@@ -0,0 +1,107 @@
...
+# digitalocean_droplets
+
+Get information on Droplets for use in other resources, with the ability to filter and sort the results.
+If no filters are specified, all Droplets will be returned.
+
+This data source is useful if the Droplets in question are not managed by Terraform or you need to
+utilize any of the Droplets' data.
+
+Note: You can use the [`digitalocean_droplet`](/docs/providers/do/d/droplet.html) data source to obtain metadata
+about a single image if you already know the `id`, unique `name`, or unique `tag` to retrieve.
```
andrewsomething, line 18:
```suggestion
about a single Droplet if you already know the `id`, unique `name`, or unique `tag` to retrieve.
```

tdyas: Thanks, good catch! Obviously, you can tell what I and Copy McPastington used as a template. :)

tdyas:
Thanks for testing this!

PR #451, opened 2020-06-12
Add datasource_digitalocean_tags and expose resource counts.

andrewsomething:
This adds a new digitalocean_tags data source for listing multiple tags. It also expands the exported attributes for all of the tag-related resources to include the resource counts that are contained in the response.


sexybutty995:
Update me

PR #457, opened 2020-07-05
Enable auto_upgrade on Kubernetes clusters

### LINKED ISSUE #237: Add digitalocean_kubernetes_cluster.auto_upgrade
### Affected Resource(s)

### Terraform Configuration Files
```hcl
resource "digitalocean_kubernetes_cluster" "foo" {
  name    = "foo"
  region  = "nyc1"
  version = "1.12.1-do.2"

  node_pool {
    name       = "worker-pool"
    size       = "s-2vcpu-2gb"
    node_count = 3
  }
}
```

### Expected Behavior
Have the option to enable auto upgrade, and even as the updates are patches of the same version of Kubernetes have it default to true.

### Actual Behavior
The cluster gets created without auto upgrade enabled.

### References
- [API endpoint allows it](https://developers.digitalocean.com/documentation/v2/#kubernetes)
- [Godo has the option](https://github.com/digitalocean/godo/blob/master/kubernetes.go#L61)

lfarnell:
This will allow users to enable `auto_upgrade` on a Kubernetes cluster.

This enhancement only handles the case of upgrading to the latest patch version. Upgrading major versions is already handled elsewhere and is not handled by this PR. Users should really use the datasource `digitalocean_kubernetes_versions` when using `auto_upgrade` to eliminate the configuration diff every time the cluster is upgraded automatically.

closes #237


digitalocean/resource_digitalocean_kubernetes_cluster_test.go
```diff
@@ -555,6 +558,26 @@ resource "digitalocean_kubernetes_cluster" "foobar" {
...
+	region  = "lon1"
+	version = data.digitalocean_kubernetes_versions.test.latest_version
+	tags    = ["one","two"]
+	version = true
+
+	node_pool {
+	  name = "default"
+		size  = "s-2vcpu-4gb"
+		node_count = 1
+		tags  = ["foo","bar"]
```
cguertin14, line 575: `name` isn't formatted like the other attributes here

lfarnell:
@andrewsomething Sorry to bug you but would anyone be able to give this PR a review? :smiley:

andrewsomething:
Hey @lfarnell! Sorry for the delay.

I'm still a bit wary about the user experience when there is a version miss-match. Using the version datasource is a good work around. Maybe that just needs to be documented a more explicitly?

lfarnell:
I have no problem re-working the PR if you have any suggestions. I can add the additional documentation around using the data source if you think this would work. Not sure what else could be done to do honest.

andrewsomething:
@lfarnell I think just adding auto-upgrade example to the docs like the one you shared in the issue with a note about preventing configuration diff should work.

andrewsomething:
Thanks so much for your work on this @lfarnell!

I ended up having to do the merge with git directly to work around some of the oddness with the history in this PR. So the actually commit is at: https://github.com/digitalocean/terraform-provider-digitalocean/commit/a975f658324e8422d5ded62f9d214e99f28b9da8

digitalocean/resource_digitalocean_kubernetes_cluster.go
```diff
@@ -315,12 +325,13 @@ func resourceDigitalOceanKubernetesClusterUpdate(d *schema.ResourceData, meta in
...
 
 	// Figure out the changes and then call the appropriate API methods
-	if d.HasChange("name") || d.HasChange("tags") || d.HasChange("surge_upgrade") {
+	if d.HasChange("name") || d.HasChange("tags") || d.HasChange("surge_upgrade") || d.HasChange("auto_upgrade") {
 
 		opts := &godo.KubernetesClusterUpdateRequest{
 			Name:         d.Get("name").(string),
 			Tags:         expandTags(d.Get("tags").(*schema.Set).List()),
 			SurgeUpgrade: d.Get("surge_upgrade").(bool),
+			AutoUpgrade:  d.Get("auto_upgrade").(*bool),
```
andrewsomething, line 334: The one functional change I made was here. This was panicking when running the acceptance tests as it should be a pointer to a bool.

```
AutoUpgrade:  godo.Bool(d.Get("auto_upgrade").(bool)),
```

PR #458, opened 2020-07-06
Fix and extend acceptance test sweeper.

andrewsomething:
I noticed that the acceptance test sweeper has not been running. This was this was because it does not have access to the provider's `DefaultFunc` at runtime. In particular, `DIGITALOCEAN_API_URL` was not being correctly set.

This resolves that issue as well as adding sweepers for database clusters and Kubernetes clusters.


MasonEgger (approved):
LGTM

PR #465, opened 2020-07-20
kubernetes - add support for surge upgrades

varshavaradarajan:
Introduce the cluster config option `surge_upgrade` for users. This will allow the users to choose if they want new worker nodes to be created before destroying the outdated nodes during an upgrade.


varshavaradarajan:
cc @adamwg, @waynr, @andrewsomething

varshavaradarajan:
@andrewsomething - thank you for reviewing this. I have addressed your comments. Do you mind taking another look when free?

PR #474, opened 2020-08-04
Update record.html.markdown

michaelwooley:
Make clear that the `name` field is modifying the hostname, which is more consistent with UI. For whatever reason I lost 15 minutes on this today trying to understand how to add a record on a subdomain. Maybe this would have cleared it up. Oftentimes "name" is more decorative/useful for other resources whereas here it is quite important to the resource itself.


MattIPv4:
Hey, @michaelwooley - Thanks a ton for this PR! 😄

PR #484, opened 2020-08-24
add digitalocean_domains data source

tdyas:
Add a `digitalocean_domains` data source as the list version of `digitalocean_domain`. This data source is not as useful it could be since `datalist` does not support regex or substring matching yet.

### DIFF
--- a/digitalocean/datasource_digitalocean_domains.go
+++ b/digitalocean/datasource_digitalocean_domains.go
@@ -1,0 +1,27 @@
+package digitalocean
+
+import (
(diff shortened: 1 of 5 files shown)


tdyas:
If https://github.com/digitalocean/terraform-provider-digitalocean/pull/481 lands first, then the `FilterKeys` and `SortKeys` fields will need to be deleted.

andrewsomething:
I've merged #481, so the FilterKeys and SortKeys fields can now be removed here too.

andrewsomething (approved):
I somehow missed that you did the rebase here. Sorry for the long delay in merging.

PR #491, opened 2020-09-22
datalist: add `all` and `match_by` attributes to filter blocks

tdyas:
Several improvements to the datalist library used by some of the data sources in this provider:

- Adds an `all` attribute to filter blocks. When set to true (i.e., `all = true`), a filter block will only match if all of the values match the given field. This is useful when matching against multi-valued fields (e.g., lists and sets).

- Adds a `match_by` attribute to filter blocks to allow string-matching to use regular expressions and substring matches. Set `match_by = "re"` for regular expression matching and `match_by = "substring"` for substring matching. This makes https://github.com/digitalocean/terraform-provider-digitalocean/pull/484 actually useful because domains can then be matched by regular expressions.

- Also refactors the filter support so that values are parsed once and not when filter blocks are applied.

Caveats:
- <strike>The docs still need updating.</strike>

- Is substring matching redundant with regular expression matching? I kept it in because it could be faster but maybe not worth it given regular expression matching can accomplish the same result.


website/docs/d/sizes.html.md
```diff
@@ -94,6 +94,13 @@ The following arguments are supported:
   `price_hourly`, or `available`.
 * `values` - (Required) Only retrieves images which keys has value that matches
```
andrewsomething, line 95:
```suggestion
* `values` - (Required) Only retrieves sizes which keys has value that matches
```

If we're going to update the docs, might as well fix this up as well.

website/docs/d/spaces_buckets.html.md
```diff
@@ -58,6 +58,14 @@ data "digitalocean_spaces_buckets" "nyc3" {
 * `values` - (Required) A list of values to match against the `key` field. Only retrieves images
```
andrewsomething, line 58:
```suggestion
* `values` - (Required) A list of values to match against the `key` field. Only retrieves Spaces buckets
```

andrewsomething (approved):
:+1:  Great stuff! This is really useful.

> Is substring matching redundant with regular expression matching? I kept it in because it could be faster but maybe not worth it given regular expression matching can accomplish the same result.

As it's already there, I think it's worth keeping it around. Some people find regex intimidating and for simple uses-cases it's less to think about.

website/docs/d/sizes.html.md
andrewsomething: Seems like these got applied and then overwritten?

tdyas: Hmm, I probably didn't pull the changes after the changes were made on GitHub, then I rebased and force pushed.

tdyas:
Okay, I made the suggested edits again.

PR #497, opened 2020-10-05
Add support for App Platform

andrewsomething:
This PR adds initial support for App Platform.

https://www.digitalocean.com/docs/app-platform/resources/app-specification-reference/
https://godoc.org/github.com/digitalocean/godo#AppsService

### DIFF
--- a/digitalocean/resource_digitalocean_app.go
+++ b/digitalocean/resource_digitalocean_app.go
@@ -1,0 +1,199 @@
+package digitalocean
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/digitalocean/godo"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceDigitalOceanApp() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceDigitalOceanAppCreate,
+		Read:   resourceDigitalOceanAppRead,
+		Update: resourceDigitalOceanAppUpdate,
+		Delete: resourceDigitalOceanAppDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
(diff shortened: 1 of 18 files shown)


digitalocean/resource_digitalocean_app.go
```diff
@@ -0,0 +1,199 @@
...
+	return nil
+}
+
+func waitForAppDeployment(client *godo.Client, id string) error {
+	tickerInterval := 10 //10s
+	timeout := 1800      //1800s, 30min
+	n := 0
+
+	// No deployment seems to be available right way, sleep 10 seconds
+	time.Sleep(10 * time.Second)
```
andrewsomething, line 163: @kamaln7 - Any hints on a better way to handle this? How long should it be before an in-progress deployment becomes available after a POST?

kamaln7: When hitting CreateApp or UpdateApp, the new deployment will be created in the same request so you can skip this altogether and hit ListDeployments right away (unless the UpdateApp call doesn't trigger a new deployment).

The deployment moving into the in-progress state happens async so if you specifically need that I don't think there's a better way than what you already have in here. In that case you should move the `Apps.Get()` call inside the loop and check for `app.InProgressDeployment != nil`. Maybe lower the interval to `5s` or so and bump it back up to `10s` once the deployment is in progress

andrewsomething: Thanks! Is there any way to know that an updated _did not_ trigger a deployment beyond the absence of a new deployment?

kamaln7: Sorry, I took a closer look and these cases all seem to be internal. It's safe to assume that hitting the UpdateApp endpoint will always trigger a new deployment

scotchneat (approved):
Some minor (non-blocking) feedback.
Otherwise, lgtm.

PR #502, opened 2020-10-17
add digitalocean_records data source

tdyas:
Adds a `digitalocean_records` data source to access DNS records of a domain.

Note: This PR modifies the `datalist` package to support "extra queries." I've exposed that by having an alternate set of callbacks that take a `*schema.ResourceData`. It might be better to just port the existing datalist uses to take that argument and avoid having two sets of functions. Thoughts?


andrewsomething:
> It might be better to just port the existing datalist uses to take that argument and avoid having two sets of functions.

:+1:  Sounds like a good approach.

tdyas:
The latest commit switches the PR to just putting the extra query map onto the existing `GetRecords` and `FlattenRecord` functions.

andrewsomething (approved):
:+1:  Looks great!

PR #503, opened 2020-10-17
port data sources and resources to use the context-related methods

tdyas:
Port all of the data sources, resources, and `datalist` package to use the context-related resource functions (e.g., `ReadContext` etc.). The previous Read/Create/Update/Delete functions have been deprecated in SDK v2.

### DIFF
--- a/digitalocean/datasource_digitalocean_account.go
+++ b/digitalocean/datasource_digitalocean_account.go
@@ -2,8 +2,8 @@ package digitalocean
 
 import (
 	"context"
-	"fmt"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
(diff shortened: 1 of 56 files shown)


andrewsomething (approved):
:+1:  LGTM!

andrewsomething:
@tdyas  Working with the new diag package, have you noticed any places we can be taking better advantage of it to return richer error messaging?

tdyas:
> Working with the new diag package, have you noticed any places we can be taking better advantage of it to return richer error messaging?

It would be useful in resources where errors/warnings can be narrowed to particular attributes since each `diag.Diagnostic` can take the a "path steps slice" from the go-cty module to apply only to particular attributes.

PR #504, opened 2020-10-21
fix docs for digitalocean_domains and digitalocean_records data sources

tdyas:
Fix copy/paste leftovers in the `digitalocean_domains` and `digitalocean_records` data sources. Also corrects the example in `digitalocean_records` to use a "list splat" correctly, and demonstrates using a regular expression match in the `digitalocean_domains` example.


andrewsomething (approved):
:+1: Thanks for catching this!

Issue #510, opened 2020-10-26
Spaces Bucket not found on apply ACL


vinicius73:
### Terraform Version
```
Terraform v0.13.5
+ provider registry.terraform.io/digitalocean/digitalocean v2.0.1
```

### Affected Resource(s)
- digitalocean_spaces_bucket

### Terraform Configuration Files

```hcl
resource "digitalocean_spaces_bucket" "index-key" {
  name   = "x-space-name-x"
  region = "nyc3"
  acl    = "private"
}
```

### Debug Output

```
2020-10-26T18:13:47.380-0300 [DEBUG] plugin.terraform-provider-digitalocean_v2.0.1: 2020/10/26 18:13:47 Bucket created
2020-10-26T18:13:47.380-0300 [DEBUG] plugin.terraform-provider-digitalocean_v2.0.1: 2020/10/26 18:13:47 [DEBUG] Spaces put bucket ACL: {
2020-10-26T18:13:47.380-0300 [DEBUG] plugin.terraform-provider-digitalocean_v2.0.1:   ACL: "private",
2020-10-26T18:13:47.381-0300 [DEBUG] plugin.terraform-provider-digitalocean_v2.0.1:   Bucket: "x-space-name-x"
2020-10-26T18:13:47.381-0300 [DEBUG] plugin.terraform-provider-digitalocean_v2.0.1: }
2020-10-26T18:13:47.381-0300 [DEBUG] plugin.terraform-provider-digitalocean_v2.0.1: 2020/10/26 18:13:47 [DEBUG] Waiting for state to become: [success]
2020-10-26T18:13:47.535-0300 [DEBUG] plugin.terraform-provider-digitalocean_v2.0.1: 2020/10/26 18:13:47 [DEBUG] Waiting for state to become: [success]
2020-10-26T18:13:47.656-0300 [DEBUG] plugin.terraform-provider-digitalocean_v2.0.1: 2020/10/26 18:13:47 [WARN] Spaces Bucket (x-space-name-x) not found, error code (404)
[...]

PR #511, opened 2020-10-28
upgrade terraform-plugin-sdk to v2.1.0

tdyas:
Upgrades `terraform-plugin-sdk` to v2.1.0 and switches to using the TypeSet test helpers in the SDK instead of the vendored copy of those helpers in `internal/setutil`.


andrewsomething:
Seeing a lot of failures in the acceptance tests. Quite a few data sources are erroring with:

```
=== CONT  TestAccDataSourceDigitalOceanImages_Basic
    datasource_digitalocean_images_test.go:23: Step 1/1 error: After applying this test step, the plan was not empty.
        stdout:
        
        
         <= read (data resources)
        
          # data.digitalocean_images.ubuntu will be read during apply
          # (config refers to values not yet known)
         <= data "digitalocean_images" "ubuntu"  {
              ~ id     = "terraform-2020102814244965780000002d" -> "terraform-2020102814245267210000002f"
              ~ images = [
                    {
```

This looks to be related to this change https://github.com/hashicorp/terraform-plugin-sdk/pull/594

Seems like we'll need to add checks that the ID has not been previously set in places like:
https://github.com/digitalocean/terraform-provider-digitalocean/blob/4ace704602c13b835b94bffe7b386b0282d9e206/digitalocean/datasource_digitalocean_region.go#L73

---

There are also a number of Spaces related failures:

```
    resource_digitalocean_spaces_bucket_object_test.go:169: Step 1/1 error: Error running apply: 
        Error: Provider produced inconsistent result after apply
        
        When applying changes to digitalocean_spaces_bucket.object_bucket_3, provider
        "registry.terraform.io/hashicorp/digitalocean" produced an unexpected new
        value: Root resource was present, but now absent.
        
        This is a bug in the provider, which should be reported in the provider's own
        issue tracker.
        
2020/10/28 10:23:13 [INFO] DigitalOcean Client configured for URL: https://api.digitalocean.com
--- FAIL: TestAccDigitalOceanSpacesBucketObject_NonVersioned (9.37s)
```

These may not be related to the SDK changes. Running them w/ `TF_LOG=DEBUG` they look simular to https://github.com/digitalocean/terraform-provider-digitalocean/issues/510

tdyas:
At least for the `digitaocean_images` datasource, the plan is also showing changes for the actual data:

```
                 ~ {
                        created        = "2020-10-12T18:57:09Z"
                        description    = "PhpMyAdmin 5.0.3 on Ubuntu 20.04"
                        distribution   = "Ubuntu"
                        error_message  = ""
                        id             = 71628256
                        image          = "71628256"
                        min_disk_size  = 25
                        name           = "PhpMyAdmin 5.0.3 on Ubuntu 20.04"
                        private        = false
                      ~ regions        = [
                            "ams2",
                            "ams3",
                            "blr1",
                            "fra1",
                            "lon1",
                            "nyc1",
                            "nyc2",
                            "nyc3",
                            "sfo1",
                            "sfo2",
                          + "sfo3",
                            "sgp1",
                            "tor1",
                        ]
                        size_gigabytes = 2.42
                        slug           = "phpmyadmin-20-04"
                        status         = "available"
                        tags           = []
                        type           = "snapshot"
                    },
```

Unclear to me why `sfo3` is being added as a region in this plan. The data should not have changed during the test run.

holms:
Wonder if this could be related #520 because no previous version of this provider adds node pools to the cluster.

tdyas:
I'm not going to have time to complete this PR. Closing for now.

andrewsomething:
There is this upstream issue: https://github.com/hashicorp/terraform-plugin-sdk/issues/586

The problem with the IDs for data sources only effects Terraform 0.13.x binaries. Using a Terraform 0.14.x binary, the problem is resolved. Though there no promise that it won't cause issues in future releases as well.

PR #513, opened 2020-10-29
Run unit tests using GitHub Actions on PR and push to master.

andrewsomething:
As we already use GitHub Actions for our build and release workflow,
we should consolidate on one CI system. This removes .travis.yml
and runs the unit tests using GitHub Actions on PR and push to master.


andrewsomething:
Note that as we use `pull_request_target`, this will not run in this PR. It will only run on subsequent PRs after merging.

https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target

bentranter (approved):
Looks great!

PR #532, opened 2020-12-01
load balancers: add support for new size field

anitgandhi:
Internal ticket: LBAAS-1582

### DIFF
--- a/digitalocean/datasource_digitalocean_loadbalancer_test.go
+++ b/digitalocean/datasource_digitalocean_loadbalancer_test.go
@@ -16,7 +16,7 @@ import (
 func TestAccDataSourceDigitalOceanLoadBalancer_Basic(t *testing.T) {
 	var loadbalancer godo.LoadBalancer
 	testName := randomTestName()
-	resourceConfig := testAccCheckDataSourceDigitalOceanLoadBalancerConfig_basic(testName)
+	resourceConfig := testAccCheckDataSourceDigitalOceanLoadBalancerConfig(testName, "lb-small")
 	dataSourceConfig := `
 data "digitalocean_loadbalancer" "foobar" {
   name = digitalocean_loadbalancer.foo.name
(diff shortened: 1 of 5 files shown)


digitalocean/resource_digitalocean_loadbalancer.go
```diff
@@ -122,6 +122,16 @@ func resourceDigitalOceanLoadBalancerV0() *schema.Resource {
 					return strings.ToLower(val.(string))
 				},
 			},
+			"size": {
+				Type:     schema.TypeString,
+				Optional: true,
```
andrewsomething, line 127: In my testing, attempting to resize a load balancer fails. If that is the intended behavior, we should set `ForceNew: true`.

```
Changes to Outputs:
  ~ size = "lb-small" -> "lb-large"

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

digitalocean_loadbalancer.public: Modifying... [id=d633a577-2fd9-4cb8-a3bc-e88591ddcc49]

Error: Error updating Load Balancer: PUT https://api.digitalocean.com/v2/load_balancers/d633a577-2fd9-4cb8-a3bc-e88591ddcc49: 422 (request "2ac0de42-5470-45a8-b127-1e4687cbf85c") size cannot be modified
```

anitgandhi: That's correct; at the moment, changing the size is not allowed. I'll add `ForceNew: true` with a note that it should be removed when resizing is allowed

digitalocean/datasource_digitalocean_loadbalancer_test.go
```diff
@@ -39,6 +39,70 @@ data "digitalocean_loadbalancer" "foobar" {
...
+						"data.digitalocean_loadbalancer.foobar", "enable_proxy_protocol", "false"),
+					resource.TestCheckResourceAttr(
+						"data.digitalocean_loadbalancer.foobar", "enable_backend_keepalive", "false"),
+				),
+			},
+		},
+	})
+}
+
+func TestAccDataSourceDigitalOceanLoadBalancer_Large(t *testing.T) {
```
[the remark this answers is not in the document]
anitgandhi: as discussed in DM, this was due to the feature not being on for the test user. After enabling it, the tests passed.

andrewsomething (approved):
:+1:  Looks good!

Issue #533, opened 2020-12-02
provider not available for openbsd_amd64


eradman:
After upgrading from terraform 0.12.24 to 0.13.5 the digitialocean provider no longer appears to be available on OpenBSD

```
$ terraform init

Initializing the backend...

Initializing provider plugins...
- Finding latest version of digitalocean/digitalocean...

Error: Failed to install provider

Error while installing digitalocean/digitalocean v2.2.0: provider
registry.terraform.io/digitalocean/digitalocean 2.2.0 is not available for
openbsd_amd64
```

```
$ terraform -v
Terraform v0.13.5
```

andrewsomething:
We've just released v2.3.0 with OpenBSD binaries: https://github.com/digitalocean/terraform-provider-digitalocean/releases/tag/v2.3.0

These should appear in the Terraform registry shortly.

eradman:
Just tested on OpenBSD. Worked perfectly. Thanks!

PR #535, opened 2020-12-03
goreleaser: Build OpenBSD binaries.

andrewsomething:
When HashiCorp managed the build and distribution of the module directly, they provided OpenBSD binaries. Looking back at the last release before the transition to the registry distribution model, they were provided for i386 and amd64. So I've skipped the ARM builds for OpenBSD. This also matches with the OpenBSD support for Terraform binaries provided by HashiCorp: https://www.terraform.io/downloads.html

```
curl 'https://registry.terraform.io/v1/providers/terraform-providers/digitalocean/versions' | jq '.versions[]  | select(.version=="1.20.0") | .platforms[] | select(.os=="openbsd")'

{
  "os": "openbsd",
  "arch": "386"
}
{
  "os": "openbsd",
  "arch": "amd64"
}
```

Closes: #533

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -19,6 +19,7 @@ builds:
     - windows
     - linux
     - darwin
+    - openbsd
   goarch:
     - amd64
     - '386'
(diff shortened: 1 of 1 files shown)


scotchneat (approved):
lgtm 👍

PR #538, opened 2020-12-19
Docs update and custom image options update

frezbo:
Signed-off-by: Noel Georgi <git@frezbo.com>

* Fixes the doc update that was missed in the initial implementation.
* Pulls in the changes to update `description` and `distribution` from https://github.com/digitalocean/godo/pull/413

### DIFF
--- a/digitalocean/resource_digitalocean_image.go
+++ b/digitalocean/resource_digitalocean_image.go
@@ -179,15 +179,28 @@ func resourceDigitalOceanCustomImageUpdate(ctx context.Context, d *schema.Resour
 		return diag.Errorf("Error converting id %s to string: %s", imageID, err)
 	}
 
-	if d.HasChange("name") {
-		name := d.Get("name").(string)
-		_, _, err := client.Images.Update(ctx, id, &godo.ImageUpdateRequest{
-			Name: name,
-		})
+	if d.HasChanges("name", "description", "distribution") {
+		imageName := d.Get("name").(string)
+		imageUpdateRequest := &godo.ImageUpdateRequest{}
+
+		if d.HasChange("name") {
+			imageUpdateRequest.Name = imageName
+		}
+		if d.HasChange("distribution") {
+			imageDistribution := d.Get("distribution").(string)
+			imageUpdateRequest.Distribution = imageDistribution
+		}
+
+		if d.HasChange("description") {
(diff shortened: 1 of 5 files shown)


digitalocean/resource_digitalocean_image.go
```diff
@@ -179,15 +179,28 @@ func resourceDigitalOceanCustomImageUpdate(ctx context.Context, d *schema.Resour
...
+		}
+		if d.HasChange("distribution") {
+			imageDistribution := d.Get("distribution").(string)
+			imageUpdateRequest.Distribution = imageDistribution
+		}
+
+		if d.HasChange("description") {
+			imageDescription := d.Get("description").(string)
+			imageUpdateRequest.Description = imageDescription
+		}
```
andrewsomething, line 197:
```suggestion
		imageUpdateRequest := &godo.ImageUpdateRequest{
			Name:         imageName,
			Distribution: d.Get("distribution").(string),
			Description:  d.Get("description").(string),
		}

```

We can simplify this a bit. PUT is idempotent, so no reason not to build the full request.

andrewsomething (commented):
Thanks for picking this up! I had meant to add these here after doing so in godo, but it seems to have slipped my mind.

It would be great to add an additional step to this test exercising the update method:

https://github.com/digitalocean/terraform-provider-digitalocean/blob/374e7a8e4f1f80bc841a36457d07dcf618c0de3d/digitalocean/resource_digitalocean_image_test.go#L13

digitalocean/resource_digitalocean_image.go
frezbo: I didn't noticed it was idempotent, was trying to change only fields that changed. I'll make this change

frezbo:
@andrewsomething I'll add a test for `update` in a bit. Finishing off something.

andrewsomething (approved):

andrewsomething:
I've gone ahead and written a test for the update method to unblock this: https://github.com/digitalocean/terraform-provider-digitalocean/pull/538/commits/21c3341059f6604c79c2fae202619e70eb02571c

Thanks again for your contributions @frezbo!

frezbo:
@andrewsomething I'm so sorry. I forgot about this. Thanks for taking it up :+1: :heart:

Issue #543, opened 2020-12-23
Add support for images attribute as per the App Specification


acraven:
### Affected Resource
`digitalocean_app`

### Expected Behavior
The `image` attribute of `service`, `static_site` and `worker` objects of `digitalocean_app` should support an object with `registry_type`, `registry`, `repository` and `tag` attributes as per the [App Specification](https://www.digitalocean.com/docs/app-platform/references/app-specification-reference/).

### Actual Behavior
The `image` attribute of `service`, `static_site` and `worker` objects of `digitalocean_app` is currently missing.

### Important Factoids
For some reason `static_site` doesn't have an `image` attribute in the App Specification. I'm not sure where the lack of support is as `job` does, but doesn't work for me. I can work around it by specifying the image I want to deploy as the base image in an otherwise empty Dockerfile and deploying that. It's not ideal as I have to them have a repo with just that in it.

### Other
I'm currently unable to use this provider as I use the `image` attribute as opposed to the `git` or `github` attributes.

andrewsomething:
Closed via https://github.com/digitalocean/terraform-provider-digitalocean/pull/565

Issue #542, opened 2020-12-23
Add support for full domain object as per the App Specification


acraven:
### Affected Resource
`digitalocean_app`

### Expected Behavior
The `domains` attribute of `digitalocean_app` should support a list of `domain`, `type`, `wildcard` and `zone` attributes as per the [App Specification](https://www.digitalocean.com/docs/app-platform/references/app-specification-reference/).

### Actual Behavior
The `domains` attribute of `digitalocean_app` is currently a list of strings.

### Other
I'm currently unable to use this provider as I need to set the `zone` of my domain.

andrewsomething:
Closed via https://github.com/digitalocean/terraform-provider-digitalocean/issues/572

PR #544, opened 2020-12-23
Replace domains string array with object, add image attribute and support job components

acraven:
This is  a spike for #542 and #543 - the image addition works for my use case, but I haven't been able to confirm the domain change yet.

GoLang's not in my skill set, so I'm not aware of any problems my changes may cause. I'm more than happy for others to get involved, so if someone could help out (coding or mentoring) that would be appreciated.

### DIFF
--- a/digitalocean/app_spec.go
+++ b/digitalocean/app_spec.go
@@ -21,10 +21,10 @@ func appSpecSchema() map[string]*schema.Schema {
 			Optional:    true,
 			Description: "The slug for the DigitalOcean data center region hosting the app",
 		},
-		"domains": {
-			Type:     schema.TypeSet,
+		"domain": {
+			Type:     schema.TypeList,
 			Optional: true,
-			Elem:     &schema.Schema{Type: schema.TypeString},
+			Elem:     appSpecDomainSchema(),
 		},
 		"service": {
 			Type:     schema.TypeList,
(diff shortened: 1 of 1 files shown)


acraven:
I can confirm that the changes appear to have worked and are visible in the App Spec download.

I do however get the following error thrown as the `terraform apply` completes:
```
Error: [DEBUG] Error setting app spec: &errors.errorString{s:"spec.0.domain.0.wildcard: '' expected type 'bool', got unconvertible type 'string'"}
```

acraven:
The last commit fixes the problem I was seeing after `terraform apply`

acraven:
Also supports changes for #545

acraven:
Job support appears to have worked and is visible in the App Spec download.

armanm:
Would love to see these changes released soon.

Thanks for working on this ❤️

andrewsomething:
Thank you for taking this on @acraven! Sorry for the delayed response over the holidays.

This looks like a great start. I have a bit of high-level feedback:

- It would be great if you could break this up into separate PRs for each feature. It will make it much easier to review and will allow us to merge in parts that are ready to go with out being blocked on ones that might need some more work.
- We should not break existing users with the change to `domains`. We can mark the old `domains` attribute as deprecated to transition users to the new `domain` attribute, but it should continue to function for now. Here's an example from a different resource: https://github.com/digitalocean/terraform-provider-digitalocean/blob/410de8b75b5f214b1a7dfef1fecbfe756f6231df/digitalocean/resource_digitalocean_volume.go#L98-L99
- We should add acceptance tests excising these features. See the existing apps tests for some examples: https://github.com/digitalocean/terraform-provider-digitalocean/blob/master/digitalocean/resource_digitalocean_app_test.go
  - The main Terraform docs should also be useful: https://www.terraform.io/docs/extend/testing/acceptance-tests/index.html
  - The whole test suite takes a _long time_ to run. You can run tests individually using: `make testacc TESTARGS='-run=TestAccDigitalOceanApp_Worker'`

Let me know if there's anything I can do to help.

dcoraboeuf:
Looking forward to have this available soon 👍

andrewsomething:
@acraven Let me know if you need any help on this.

andrewsomething:
@acraven Thank you so much for kicking this off. @rienafairefr has finished up the images work in https://github.com/digitalocean/terraform-provider-digitalocean/pull/565 and jobs in https://github.com/digitalocean/terraform-provider-digitalocean/pull/566

https://github.com/digitalocean/terraform-provider-digitalocean/pull/572 has the domain changes handled in a way that should be non-disruptive for current users.

MattIPv4:
Hey, @acraven - Thanks a ton for this sweet PR! 😄

PR #553, opened 2021-01-14
bump kubernetes version to 1.19.3-do.3

yuta-ron:
Thank you for creating such great documentation!

However, when I try to create a kubernetes cluster using the following as a reference, I get the error.

Referred: 
https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/resources/kubernetes_cluster
https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/resources/kubernetes_node_pool

result: 
```
Error: Error creating Kubernetes cluster: POST https://api.digitalocean.com/v2/kubernetes/clusters: 422 validation error: invalid version slug
```

This is probably due to the fact that it is an older version and is not supported.
So I changed to the current latest version and it works fine now.

What do you feel about this PR?

Thanks!

### DIFF
--- a/docs/resources/kubernetes_cluster.md
+++ b/docs/resources/kubernetes_cluster.md
@@ -15,7 +15,7 @@ resource "digitalocean_kubernetes_cluster" "foo" {
   name   = "foo"
   region = "nyc1"
   # Grab the latest version slug from `doctl kubernetes options versions`
-  version = "1.15.5-do.1"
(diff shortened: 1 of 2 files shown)


andrewsomething (approved):
:+1:  Thanks for the update!

Unfortunately, it is a bit of moving target. That's why we have the comment about checking for a current version with `doctl`.

Issue #554, opened 2021-01-16
Security: Obfuscate secret env vars in tf plan output for apps


pjhampton:
### Terraform Version
```
Terraform v0.12.21
+ provider.digitalocean v2.3.0
```

### Affected Resource(s)
- digitalocean_app

Transistive dependencies

- digitalocean_database_cluster
- digitalocean_database_user

### Terraform Configuration Files
```hcl
resource "digitalocean_app" "REDACTED" {
  spec {
    name   = "REDACTED"
    region = "nyc"

    service {
      name               = "www-REDACTED"
      dockerfile_path    = "REDACTED"
      http_port          = 80
      instance_count     = 1
      instance_size_slug = "professional-xs"

      github {
        repo           = "REDACTED"
        branch         = "master"
        deploy_on_push = false
      }

      env {
        key   = "DB_USER"
        value = digitalocean_database_user.REDACTED.password
        type  = "SECRET"
      }

      env {
        key   = "random"
        value = "pete"
        type  = "SECRET"
      }
    }
  }

  depends_on = []
}
```

### Expected Behavior

Secret values should be redacted, replaced or obfuscated. 

### Actual Behavior

The tf plan is showing the value of an env var that is marked secret:

```
              + env {
                  + key   = "DB_USER"
                  + scope = "RUN_AND_BUILD_TIME"
                  + type  = "SECRET"
                  + value = "o1xzzxkczq709hp3"
                }
              + env {
                  + key   = "random"
                  + scope = "RUN_AND_BUILD_TIME"
                  + type  = "SECRET"
                  + value = "pete"
                }
```

I've killed this user and regenerated password for bug reporting purposes.

### Steps to Reproduce

1. Create a `digitalocean_database_cluster` and `digitalocean_database_user` resource
1. Create a `digitalocean_app` resource and reference the db user in a secret env var
1. `terraform plan`

### Important Factoids
Because I use an e2e gitops flow for everything in my project leaking secret env vars into Github or the CI would be too much of a security risk for me. By not being able to do this I wouldn't be able to stand an e2e test environment up automatically as the app couldn't reference the database user / pass. 

### References
https://github.com/hashicorp/terraform/issues/20051#issuecomment-484154409

PR #555, opened 2021-01-19
apps: Mark env var values as sensitive (Closes: #554).

andrewsomething:
An app env var's `type` can be `GENERAL` or `SECRET`. Unfortunately there is not a way to set an attribute on one field based on an attribute from another. So we should error on the side of safely, and treat all env var values as sensitive when it comes to logging.


ChiefMateStarbuck (approved):
LGTM

Issue #562, opened 2021-01-24
digitalocean_kubernetes_cluster has empty attributes when digitalocean_kubernetes_versions does not match actual cluster version


devurandom:
### Terraform Version

```
❯ terraform -v
Terraform v0.14.5
```

### Affected Resource(s)
Please list the resources as a list, for example:
- digitalocean_kubernetes_versions
- digitalocean_kubernetes_cluster
- kubernetes (provider)

### Terraform Configuration Files

Config excerpt:

```hcl
data "digitalocean_kubernetes_versions" "default" {
  version_prefix = "1.18."
}

resource "digitalocean_kubernetes_cluster" "default" {
  [...]
  auto_upgrade = true
  version      = data.digitalocean_kubernetes_versions.default.latest_version
  [...]
}

provider "kubernetes" {
  host             = digitalocean_kubernetes_cluster.default.endpoint
  token            = digitalocean_kubernetes_cluster.default.kube_config[0].token
  cluster_ca_certificate = base64decode(
    digitalocean_kubernetes_cluster.default.kube_config[0].cluster_ca_certificate
  )
}
```

### Expected Behavior

`terraform plan` or `terraform refresh` should not output an error.

### Actual Behavior

```
Error: Get "http://localhost/api/v1/namespaces/[...]": dial tcp [::1]:80: connect: connection refused
Error: Kubernetes cluster unreachable: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
```

### Steps to Reproduce

1. I recently upgraded my DigitalOcean Kubernetes cluster from 1.18 to 1.19 using the web UI
2. Today I tried to `terraform refresh` and received abovementioned error messages
3. I tried `terraform plan` and saw the same messages
4. Providing a literal string (instead of referring to the `digitalocean_kubernetes_cluster` resource) for the `host` of the `kubernetes` provider turns the error message into: `Error: Get "[...]/api/v1/namespaces/[...]": x509: certificate signed by unknown authority` (now using the correct endpoint instead of `http://localhost`)
5. Changing `version_prefix` from `"1.18."` to `"1.19."` resolves the problem (the error messages do not appear any longer)

andrewsomething:
Hi @devurandom, 

What version of the Kubernetes provider are you using? It looks like they've recently released a new version with some behavior changes:

https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/guides/v2-upgrade-guide

[...]

PR #564, opened 2021-01-28
kubernetes: Expand example to show best practices for interacting with k8s provider.

andrewsomething:
As seen in https://github.com/digitalocean/terraform-provider-digitalocean/issues/562, using interpolation to pass credentials to the Kubernetes provider from other resources can cause some issues due to the order in which Terraform evaluates the provider blocks vs. actual resources. In fact, the Kubernetes provider docs have added a warning to discourage using interpolation to pass credentials: 

https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#stacking-with-managed-kubernetes-cluster-resources

This PR expands the Kubernetes example in this repo to be more complete and better adhere to best practices for interacting with the Kubernetes provider. It also adds a notes to the Kubernetes cluster resource docs addressing the issue and linking to that example. Finally, it updates usage of the Kubernetes provider for it 2.0.x release (specifically the removal of the `load_config_file` attribute).

Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/562

### DIFF
--- a/docs/resources/kubernetes_cluster.md
+++ b/docs/resources/kubernetes_cluster.md
@@ -25,35 +25,6 @@ resource "digitalocean_kubernetes_cluster" "foo" {
 }
 ```
 
-### Kubernetes Terraform Provider Example
-
-The cluster's kubeconfig is exported as an attribute allowing you to use it with the [Kubernetes Terraform provider](https://www.terraform.io/docs/providers/kubernetes/index.html). For example:
-
-```hcl
-resource "digitalocean_kubernetes_cluster" "foo" {
-  name   = "foo"
-  region = "nyc1"
-  # Grab the latest version slug from `doctl kubernetes options versions`
-  version = "1.19.3-do.3"
-  tags    = ["staging"]
-
-  node_pool {
-    name       = "worker-pool"
-    size       = "s-2vcpu-2gb"
-    node_count = 3
-  }
-}
-
-provider "kubernetes" {
-  load_config_file = false
-  host             = digitalocean_kubernetes_cluster.foo.endpoint
-  token            = digitalocean_kubernetes_cluster.foo.kube_config[0].token
-  cluster_ca_certificate = base64decode(
-    digitalocean_kubernetes_cluster.foo.kube_config[0].cluster_ca_certificate
-  )
-}
-```
-
 ### Autoscaling Example
 
 Node pools may also be configured to [autoscale](https://www.digitalocean.com/docs/kubernetes/how-to/autoscale/).
@@ -102,6 +73,47 @@ resource "digitalocean_kubernetes_cluster" "foo" {
 ```
 
 Note that a data source is used to supply the version. This is needed to prevent configuration diff whenever a cluster is upgraded.
+
+### Kubernetes Terraform Provider Example
+
+The cluster's kubeconfig is exported as an attribute allowing you to use it with
+the [Kubernetes Terraform provider](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs).
+
+~> When using interpolation to pass credentials from a `digitalocean_kubernetes_cluster`
+resource to the Kubernetes provider, the cluster resource generally should not
+be created in the same Terraform module where Kubernetes provider resources are
+also used. This can lead to unpredictable errors which are hard to debug and
+diagnose. The root issue lies with the order in which Terraform itself evaluates
+the provider blocks vs. actual resources.
+
+When using the Kubernetes provider with a cluster created in a separate Terraform
+module or configuration, use the [`digitalocean_kubernetes_cluster` data-source](https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/data-sources/kubernetes_cluster)
+to access the cluster's credentials. [See here for a full example](https://github.com/digitalocean/terraform-provider-digitalocean/tree/main/examples/kubernetes).
+
+```hcl
+resource "digitalocean_kubernetes_cluster" "foo" {
+  name   = "foo"
+  region = "nyc1"
+  # Grab the latest version slug from `doctl kubernetes options versions`
+  version = "1.19.3-do.3"
+  tags    = ["staging"]
+
+  node_pool {
+    name       = "worker-pool"
+    size       = "s-2vcpu-2gb"
+    node_count = 3
+  }
+}
+
+provider "kubernetes" {
+  load_config_file = false
+  host             = digitalocean_kubernetes_cluster.foo.endpoint
+  token            = digitalocean_kubernetes_cluster.foo.kube_config[0].token
+  cluster_ca_certificate = base64decode(
+    digitalocean_kubernetes_cluster.foo.kube_config[0].cluster_ca_certificate
+  )
+}
+```
 
 ## Argument Reference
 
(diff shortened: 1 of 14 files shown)


examples/kubernetes/README.md
```diff
@@ -0,0 +1,75 @@
+# DigitalOcean Kubernetes Service (DOKS)
+
+This example shows how to use the Terraform Kubernetes Provider and Terraform
+Helm Provider to configure a DOKS cluster. The example config builds the DOKS
+cluster and applies the Kubernetes configurations in a single operation. This
+guide will also show you how to make changes to the underlying DOKS cluster in
+such a way that Kuberntes/Helm resources are recreated after the underlying
```
devurandom, line 7:
```suggestion
such a way that Kubernetes/Helm resources are recreated after the underlying
```

examples/kubernetes/README.md
```diff
@@ -0,0 +1,75 @@
...
+Optionally, this example can generate a kubeconfig file in the current working
+directory. However, the token in this config will expire. The token can be
+refreshed by running `terraform apply` again and setting `write_kubeconfig` to
+`true`.
+
+```
+terraform apply -var="write_kubeconfig=true"
+export KUBECONFIG=$(terraform output -raw kubeconfig_path
+kubectl get pods -n test
+```
```
devurandom, line 49: I use a combination of `terraform refresh` and `terraform output kubeconfig > .../kubeconfig`. That seems less intimidating (i.e. potentially destructive) that `terraform apply`.

examples/kubernetes/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+Alternatively, a longer-lived configuration can be generated using doctl. This
+command will merge the configuration into your kubeconfig at the default location
+(`$HOME/.kube/config`) or create it if the file does not exist.
+
+```
+doctl kubernetes cluster kubeconfig save $(terraform output -raw cluster_name)
+kubectl get pods -n test
+```
```
devurandom, line 58: Wow, that's great! Thanks for this info!

examples/kubernetes/kubernetes-config/main.tf
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+}
+
+data "digitalocean_kubernetes_cluster" "primary" {
+  name = var.cluster_name
+}
+
+resource "local_file" "kubeconfig" {
+  depends_on = [var.cluster_id]
```
devurandom, line 23: This establishes a dependency on the doks-cluster module, because referencing the `data.digitalocean_kubernetes_cluster` (instead of the `resource`) decoupled them?

examples/kubernetes/kubernetes-config/main.tf
```diff
@@ -0,0 +1,147 @@
...
+  content    = data.digitalocean_kubernetes_cluster.primary.kube_config[0].raw_config
+  filename   = "${path.root}/kubeconfig"
+}
+
+provider "kubernetes" {
+  host             = data.digitalocean_kubernetes_cluster.primary.endpoint
+  token            = data.digitalocean_kubernetes_cluster.primary.kube_config[0].token
+  cluster_ca_certificate = base64decode(
+    data.digitalocean_kubernetes_cluster.primary.kube_config[0].cluster_ca_certificate
+  )
```
devurandom, line 34: What ensures that the `data.digitalocean_kubernetes_cluster` resource is created after the actual `digitalocean_kubernetes_cluster` resource, i.e. what ensures that it contains the actual credentials and is not prone to ordering problems?

examples/kubernetes/README.md
```diff
@@ -0,0 +1,75 @@
...
+below. This is done by removing the `module.kubernetes-config` from Terraform
+State prior to replacing cluster credentials, to avoid passing outdated
+credentials into the providers.
+
+This will create the new cluster and the Kubernetes resources in a single apply.
+
+```
+terraform state rm module.kubernetes-config
+terraform apply
+```
```
devurandom, line 75: This only works if `apply` would actually recreate the cluster, is that correct? If it would not, Terraform would just "forget" about all the things it created on the Kubernetes cluster, using Helm and Kubernetes itself?

I have been in problematic situations, where Helm releases still existed on the cluster, but Terraform had "forgotten" about them and wasn't able to either destroy nor create them. I would want to avoid such states, because it took a while to figure out how to get out of it again.

Maybe we could instead recommend @lfarnell's "trick" with `terraform apply -module=...` as a safer / more newbie-friendly alternative?

devurandom (commented):
Thanks a lot for this improvement! I have a few questions the answers to which might also help other newbies with DOKS and Terraform.

examples/kubernetes/README.md
andrewsomething: Since this creates a local_file resource, apply is needed to replace the file if new credentials are needed.

docs/resources/kubernetes_cluster.md
```diff
@@ -103,6 +74,47 @@ resource "digitalocean_kubernetes_cluster" "foo" {
...
+
+provider "kubernetes" {
+  load_config_file = false
+  host             = digitalocean_kubernetes_cluster.foo.endpoint
+  token            = digitalocean_kubernetes_cluster.foo.kube_config[0].token
+  cluster_ca_certificate = base64decode(
+    digitalocean_kubernetes_cluster.foo.kube_config[0].cluster_ca_certificate
+  )
+}
+```
```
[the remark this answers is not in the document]
andrewsomething: I originally thought the warning was enough, but it probably does make more sense to just use the data source here.

examples/kubernetes/kubernetes-config/main.tf
andrewsomething: Right. In this example, the aim is to still do something that can be created in a single apply. So we do want to establish a dependency here, but we do want to decouple the actual credentials by using the data-source.

examples/kubernetes/README.md
```diff
@@ -0,0 +1,75 @@
...
+## Kubeconfig for manual CLI access
+
+Optionally, this example can generate a kubeconfig file in the current working
+directory. However, the token in this config will expire. The token can be
+refreshed by running `terraform apply` again and setting `write_kubeconfig` to
+`true`.
+
+```
+terraform apply -var="write_kubeconfig=true"
+export KUBECONFIG=$(terraform output -raw kubeconfig_path
```
ChiefMateStarbuck, line 47:
``` suggestion
export KUBECONFIG=$(terraform output -raw kubeconfig_path)
```
would it be proper to close out the command?

examples/kubernetes/kubernetes-config/main.tf
andrewsomething: Data sources always call out to the API, while a resource may use cached information. So stale credentials should not be returned.

examples/kubernetes/README.md
andrewsomething: This should only be needed when apply replaces the entire cluster to ensure that all of the Kubernetes configuration will be correctly recreated.

There a few ways to approach this, but I'm aiming to keep this example mostly inline with the one that the Kubernetes provider has for other clouds.

docs/resources/kubernetes_cluster.md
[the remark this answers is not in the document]
andrewsomething: Updated. Thanks for all the feedback!

ChiefMateStarbuck (approved):
LGTM

scotchneat (approved):
Looks awesome @andrewsomething

PR #566, opened 2021-01-29
App jobs attribute

rienafairefr:
I extracted the jobs attribute work from #544 and added some acceptance tests.

### DIFF
--- a/digitalocean/resource_digitalocean_app_test.go
+++ b/digitalocean/resource_digitalocean_app_test.go
@@ -141,14 +141,6 @@ func TestAccDigitalOceanApp_Basic(t *testing.T) {
 						"digitalocean_app.foobar", "spec.0.database.0.engine", "PG"),
 				),
 			},
-			{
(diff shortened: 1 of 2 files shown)


digitalocean/resource_digitalocean_app_test.go
```diff
@@ -633,3 +664,57 @@ resource "digitalocean_app" "foobar" {
     }
   }
 }`
+
+var testAccCheckDigitalOceanAppConfig_addJob = `
```
andrewsomething, line 668: A job is expected to complete as it is run as part of the deploy process. (Think database migrations.) The `sample-sleeper` program runs indefinitely. So I changed this test up to use something else.

andrewsomething (approved):
:+1: I pushed a few additional commits that take into account GitLab support and images support which were added after this was originally written.

Thanks again for you contributions!

rienafairefr:
Thanks @andrewsomething.

PR #568, opened 2021-01-31
changed up the default for routes

killyosaur:
added an override for the worker as I am betting that is where the issue is

This may be an incomplete change, I am just dropping it in place as a possible solution. Specifically it appears that we are worried about workers which makes no sense so I removed the maxItem and created an override of routes for the worker spec section. If I am missing something (or if the real issue is with workers and services, because I know for a fact I can add more than 1 route to a static site) definitely start the conversation, I'll fix and update whatever necessary to move this project forward...

### DIFF
--- a/digitalocean/app_spec.go
+++ b/digitalocean/app_spec.go
@@ -241,7 +241,6 @@ func appSpecComponentBase() map[string]*schema.Schema {
 			Type:     schema.TypeList,
 			Optional: true,
 			Computed: true,
-			MaxItems: 1,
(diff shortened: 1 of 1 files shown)


andrewsomething:
Hi @killyosaur,

Thanks for raising this issue for us. It looks like there were some faulty assumptions when this was first implemented. Services and static sites can indeed have multiple routes specified while workers and jobs are not publicly rotatable.

https://github.com/digitalocean/terraform-provider-digitalocean/pull/571 should resolve this.

PR #570, opened 2021-02-01
App: internal ports array attribute

rienafairefr:
The digitalocean_app spec was also missing the internal_ports attribute on a service, this should fix this. Acceptance test (`TestAccDigitalOceanApp_InternalPort`) is passing
internal_ports are specified like `internal_ports = ["5000"]`, maybe an array syntax
```
internal_port {
   port = 5000
}
internal_port {
   port = 2000
}
```
would be preferred, similarly to the routes attribute. not sure.

### DIFF
--- a/digitalocean/app_spec.go
+++ b/digitalocean/app_spec.go
@@ -1011,9 +1011,9 @@ func expandAppSpecServices(config []interface{}) []*godo.AppServiceSpec {
 			s.HealthCheck = expandAppHealthCheck(checks)
 		}
 
-		internal_ports := service["internal_ports"].([]interface{})
-		if len(internal_ports) > 0 {
-			s.InternalPorts = expandAppInternalPort(internal_ports)
+		internalPorts := service["internal_ports"].([]interface{})
+		if len(internalPorts) > 0 {
+			s.InternalPorts = expandAppInternalPort(internalPorts)
 		}
 
 		appServices = append(appServices, s)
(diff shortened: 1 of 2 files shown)


andrewsomething:
> passing internal_ports are specified like internal_ports = ["5000"], maybe an array syntax

I think you've got it right as it currently is. I matches how the yaml spec spec treats them:

```
  internal_ports:
  - 5000
  - 3000
  routes:
  - path: /
  - path: /foo
```

https://www.digitalocean.com/docs/app-platform/references/app-specification-reference/

`routes` could grow additional attributes while `internal_ports` will always be a list of integers.

digitalocean/app_spec.go
```diff
@@ -190,6 +190,17 @@ func appSpecRouteSchema() map[string]*schema.Schema {
 	}
 }
 
+func appSpecInternalPortSchema() *schema.Resource {
```
andrewsomething, line 193: Looks like this is not used?

andrewsomething (changes requested):
Looks great. I left one comment in line. Think we can get rid of `appSpecInternalPortSchema()`

digitalocean/app_spec.go
andrewsomething: I've pushed a commit removing this.

andrewsomething (approved):
:shipit: Thanks again for the great contributions!

digitalocean/app_spec.go
rienafairefr: Thanks!

Issue #576, opened 2021-02-04
Database without private networking causes crash on state refresh


flyte:
I'm in the process of upgrading from Terraform 0.11 to 0.14, currently upgrading to 0.12.

I created a DO postgres database using 1.2.0 (at the time it was unreleased, and I had to compile it myself) of this provider. However, now that I'm using Terraform 0.12, and have upgraded to 2.5.0 of the DO provider, I get a crash when I try to run `terraform apply`.

I've narrowed it down to the database, and created a brand new terraform 0.12 stack that just has the db in. It now crashes in the same way when I import the existing database using its ID.

### Terraform Version
```
Terraform v0.12.30
+ provider.digitalocean v2.5.0
```
### Affected Resource(s)

- digitalocean_database_cluster

### Terraform Configuration Files
```hcl
provider "digitalocean" {
  token   = local.digitalocean_token
  version = "~> 2.5.0"
}
```
```hcl
locals {
  digitalocean_token   = "<REDACTED>"
}
```
```hcl
resource "digitalocean_database_cluster" "db" {
  name       = "micro-staging"
  engine     = "pg"
  version    = "11"
  size       = "db-s-1vcpu-1gb"
  region     = "lon1"
  node_count = 1

  maintenance_window {
    day  = "saturday"
    hour = "07:00:00"
  }

  lifecycle {
    ignore_changes = [
      maintenance_window,
    ]
  }
}
```

### Debug Output
https://gist.github.com/flyte/b7be0b1a92b0d6c12c5737289b0f0bdf#file-command-output

### Panic Output
https://gist.github.com/flyte/b7be0b1a92b0d6c12c5737289b0f0bdf#file-crash-log

### Expected Behavior
State should have been updated from the data received from the DO API

### Actual Behavior
Terraform crashed

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply` when you've got the DB in your state file already, or:
2. `terraform import digitalocean_database_cluster.db <REDACTED DB ID>`

### Important Factoids
When comparing the response from the DO API to a freshly made database (with the latest DO provider), the database I'm trying to get the status of (or import), doesn't contain the `private_connection` section or the `private_network_uuid` value.

flyte:
[...]

PR #577, opened 2021-02-04
dbaas: Protect against panic if connection details not available.

andrewsomething:
Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/576

### DIFF
--- a/digitalocean/datasource_digitalocean_database_cluster.go
+++ b/digitalocean/datasource_digitalocean_database_cluster.go
@@ -176,14 +176,20 @@ func dataSourceDigitalOceanDatabaseClusterRead(ctx context.Context, d *schema.Re
 				}
 			}
 
-			d.Set("host", db.Connection.Host)
-			d.Set("private_host", db.PrivateConnection.Host)
-			d.Set("port", db.Connection.Port)
-			d.Set("uri", db.Connection.URI)
-			d.Set("private_uri", db.PrivateConnection.URI)
-			d.Set("database", db.Connection.Database)
-			d.Set("user", db.Connection.User)
-			d.Set("password", db.Connection.Password)
+			if db.Connection != nil {
+				d.Set("host", db.Connection.Host)
+				d.Set("port", db.Connection.Port)
+				d.Set("uri", db.Connection.URI)
+				d.Set("database", db.Connection.Database)
+				d.Set("user", db.Connection.User)
+				d.Set("password", db.Connection.Password)
+			}
+
+			if db.PrivateConnection != nil {
+				d.Set("private_host", db.PrivateConnection.Host)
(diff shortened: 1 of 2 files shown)


digitalocean/datasource_digitalocean_database_cluster.go
```diff
@@ -176,14 +176,20 @@ func dataSourceDigitalOceanDatabaseClusterRead(ctx context.Context, d *schema.Re
...
+				d.Set("password", db.Connection.Password)
+			}
+
+			if db.PrivateConnection != nil {
+				d.Set("private_host", db.PrivateConnection.Host)
+				d.Set("private_uri", db.PrivateConnection.URI)
+			}
+
 			d.Set("urn", db.URN())
 			d.Set("private_network_uuid", db.PrivateNetworkUUID)
```
flyte, line 194: You'll have to check for the existence of `db.PrivateNetworkUUID` as well, as `private_network_uuid` doesn't exist in the API response when the DB isn't assigned to a VPC at all (any DB created before VPCs were added).

digitalocean/resource_digitalocean_database_cluster.go
```diff
@@ -417,14 +417,20 @@ func resourceDigitalOceanDatabaseClusterRead(ctx context.Context, d *schema.Reso
...
+		d.Set("password", database.Connection.Password)
+	}
+
+	if database.PrivateConnection != nil {
+		d.Set("private_host", database.PrivateConnection.Host)
+		d.Set("private_uri", database.PrivateConnection.URI)
+	}
+
 	d.Set("urn", database.URN())
 	d.Set("private_network_uuid", database.PrivateNetworkUUID)
```
flyte, line 435: Same here.

flyte (changes requested):

digitalocean/datasource_digitalocean_database_cluster.go
andrewsomething: The type for a `db.PrivateNetworkUUID` is string. So if it does not exist, it be assigned the default value which is an empty string. The problem that causes the panic is that `db.PrivateConnection` is a struct. The default value of a struct is `nil`. So when you attempt to access an attribute (e.g. `db.PrivateConnection.URI`) it causes a panic as `nil` does not have any attributes.

ChiefMateStarbuck (approved):
Clean and Mean 👍🏾

digitalocean/datasource_digitalocean_database_cluster.go
flyte: Ahh, OK, thanks for the explanation! I've not used Go yet :)

flyte (approved):

Issue #581, opened 2021-02-09
Attempting to set too-high expiry_seconds on registry docker credentials gets into a broken state


flyte:
### Terraform Version
```
Terraform v0.14.6
+ provider registry.terraform.io/digitalocean/digitalocean v2.5.1
```

### Affected Resource(s)
- digitalocean_container_registry_docker_credentials

### Terraform Configuration Files
```hcl
provider "digitalocean" {
  token = local.digitalocean_token
}
```
```hcl
resource "digitalocean_container_registry" "myregistry" {
  name                   = "myregistry"
  subscription_tier_slug = "basic"
}

resource "digitalocean_container_registry_docker_credentials" "myregistrycreds" {
  registry_name = digitalocean_container_registry.myregistry.name
  write         = true
  expiry_seconds = 1576800000  # 50 years
}
```

### Debug Output
https://gist.github.com/flyte/4cdcf871a3990ee1d01e77c6ffa13044

### Panic Output
n/a

### Expected Behavior
When trying to set a wrong value for `expiry_seconds`, the provider shouldn't update the state with the bad value when the DO API returns an error.

Also perhaps if the state does end up with an out-of-bounds value, don't error out when you're trying to apply a change back to a legal value.

### Actual Behavior
The state has been updated to include the `1576800001` value for `expiry_seconds` even though the update was rejected by the DO API. Now when you try to run `terraform apply`, it returns the following error:

```
Error: expiry_seconds outside acceptable range
```

_See [the gist](https://gist.github.com/flyte/4cdcf871a3990ee1d01e77c6ffa13044) for trace output_

### Steps to Reproduce
1. Create the resources above with `terraform apply`
2. Change `expiry_seconds` to `1576800001`
3. Run `terraform apply`

The API will respond saying that it couldn't modify the resource:

```
 # digitalocean_container_registry_docker_credentials.myregistrycreds will be updated in-place
  ~ resource "digitalocean_container_registry_docker_credentials" "myregistrycreds" {
      ~ expiry_seconds             = 1576800000 -> 1576800001
        id                         = "myregistry"
        # (4 unchanged attributes hidden)
    }

...

digitalocean_container_registry_docker_credentials.myregistrycreds: Modifying... [id=myregistry]

Error: Error retrieving docker credentials: GET https://api.digitalocean.com/v2/registry/docker-credentials?expiry_seconds=1576800001&read_write=true: 422 cannot set an expiration greater than 50 years
```

[...]

PR #582, opened 2021-02-09
Fix docs and validation for expiry_seconds on registry docker credentials resource.

andrewsomething:
The `updateExpiredDockerCredentials` function gets called when the state is being refreshed, but the change to the configuration is not available at refresh. So you can encounter a situation like #581 where a bad value has been set in state but the validation prevents the value from being corrected. We should prevent the bad value from ever being entered in the first place.

Fixes: #581


bentranter (approved):
LGTM!

PR #584, opened 2021-02-15
kubernetes - enable surge upgrades by default during cluster creation

varshavaradarajan:
Description: Now that surge upgrades is stable, enable surge upgrades by default as it is a better user experience during upgrades.

cc @adamwg

### DIFF
--- a/digitalocean/resource_digitalocean_kubernetes_cluster.go
+++ b/digitalocean/resource_digitalocean_kubernetes_cluster.go
@@ -26,7 +26,7 @@ func resourceDigitalOceanKubernetesCluster() *schema.Resource {
 		Importer: &schema.ResourceImporter{
 			State: resourceDigitalOceanKubernetesClusterImportState,
 		},
-		SchemaVersion: 2,
+		SchemaVersion: 3,
 
 		Schema: map[string]*schema.Schema{
 			"name": {
@@ -49,6 +49,7 @@ func resourceDigitalOceanKubernetesCluster() *schema.Resource {
 			"surge_upgrade": {
 				Type:     schema.TypeBool,
 				Optional: true,
+				Default: true,
 			},
 
 			"version": {


digitalocean/datasource_digitalocean_kubernetes_cluster.go
```diff
@@ -32,6 +32,7 @@ func dataSourceDigitalOceanKubernetesCluster() *schema.Resource {
 			"surge_upgrade": {
 				Type:     schema.TypeBool,
 				Computed: true,
+				Default: true,
```
andrewsomething, line 35: Data sources just report the state returned by the API for existing resources. You probably want to change the resource:

https://github.com/digitalocean/terraform-provider-digitalocean/blob/c1fb04686f1ffd2fc6f49a8de3102f429335dd93/digitalocean/resource_digitalocean_kubernetes_cluster.go#L49

andrewsomething (changes requested):
Unfortunately, changing the default for a non-required attribute can be intrusive. Unless a user has explicitly set `surge_upgrade` in their configuration, they will be prompted to apply changes to existing clusters. E.g. 

```
  ~ update in-place

  # digitalocean_kubernetes_cluster.foo will be updated in-place
  ~ resource "digitalocean_kubernetes_cluster" "foo" {
        id             = "21ea2185-bc6d-4a64-8797-e7898bc57d2f"
        name           = "foo"
      ~ An execution plan has been generated and is shown below.
  ~ update in-place

  # digitalocean_kubernetes_cluster.foo will be updated in-place
  ~ resource "digitalocean_kubernetes_cluster" "foo" {
        id             = "21ea2185-bc6d-4a64-8797-e7898bc57d2f"
        name           = "foo"
      ~ surge_upgrade  = false -> true
        tags           = []
        # (12 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.-> true
        tags           = []
        # (12 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

The user can add `surge_upgrade = false` explicitly to prevent the diff.

I'm not necessarily opposed to the change if we think surge upgrades will provide a better user experience in the long run, but want to make the consequences of the change clear.

digitalocean/datasource_digitalocean_kubernetes_cluster.go
varshavaradarajan: @andrewsomething  - I'm not sure if I should change the schema version if I just provide a default value. I have done so. But, let me know if that's necessary.

andrewsomething: Generally it's not needed unless you are doing a state migration, but there is no problem with bumping it.

https://www.terraform.io/docs/extend/resources/state-migration.html

andrewsomething (approved):

Issue #585, opened 2021-02-17
Weird behaviour (remove all+add all) when changing resources in digitalocean_project


morsik:
Hi there,

Not sure if this is related to provider itself or DigitalOcean web page, but here what's the problem.

While adding or removing `resources` to `digitalocean_project`, webpage GUI tells that everything was removed and everything was added again even I removed only 1 item from resources. This feels very wrong and I always feel I just destroyed too much.

### Terraform Version
```
Terraform v0.14.6
+ provider registry.terraform.io/digitalocean/digitalocean v2.5.1
```

### Affected Resource(s)
- `digitalocean_project`
- `digitalocean_project_resources` (possibly, not tested)

 

### Expected Behavior
Web page should just show notification about removing exactly 1 item. Not removing everything and adding back again.

### Actual Behavior
Shows information about removing everything!

### Terraform Configuration Files
Just create simpliest `digitalocean_project` like:
```hcl
# of course here you need those actual resources too, I'm not entering them for simplicity

resource "digitalocean_project" "labs" {
  name        = "labs"
  resources = [
    digitalocean_domain.lab.urn,
    digitalocean_droplet.swarmtest_01.urn,
    digitalocean_droplet.swarmtest_02.urn,
    digitalocean_droplet.swarmtest_03.urn,
  ]
}
```

Apply Terraform. Check webpage (you'll see 3 Droplets in this Project).

After checking, remove one Droplet resource (for ex `digitalocean_droplet.swarmtest_03.urn`). Apply and check again. If you slow enough, everything will look fine, but if you were watching DigitalOcean webpage in real-time you'll see that all resources were actually **deleted** from this project (and moved to main project) and then restored back again (without the one you removed).

Also there is a lot of notifications about this:

![Screenshot from 2021-02-17 20-03-47](https://user-images.githubusercontent.com/783794/108255512-a559fc80-715c-11eb-8b31-b1b61d2e0e86.png)

andrewsomething:
Thanks for raising this issue. It looks it currently does remove all resource and then adds them back:

https://github.com/digitalocean/terraform-provider-digitalocean/blob/2c4bf56fe1da163346f8bfa98ec9dbb996355b3b/digitalocean/resource_digitalocean_project.go#L208-L209

We can definitely improve the experience here.

PR #586, opened 2021-02-17
Assign and remove project resources without unnecessary churn (Fixes: #585).

andrewsomething:
The current implementation moves all resources in the old config to the default project and then adds the resources in the new config to the Terraform-managed project. This creates a lot of unnecessary churn.

https://github.com/digitalocean/terraform-provider-digitalocean/blob/2c4bf56fe1da163346f8bfa98ec9dbb996355b3b/digitalocean/resource_digitalocean_project.go#L208-L209

This PR adds a helper function to compares two `*schema.Set`s, producing lists of which should be added and which should be removed. Resources that were in both the old and new configs will no longer be touched.

Fixes: #585


bentranter (approved):
Looks great!

PR #589, opened 2021-02-19
Run tests on pull_request not pull_request_target.

andrewsomething:
Run tests on `pull_request` not `pull_request_target` to  prevent failures like this from only being found after merge into main.

https://github.com/digitalocean/terraform-provider-digitalocean/runs/1936909608?check_suite_focus=true

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,7 +3,7 @@
 name: Unit Test
 
 on:
-  pull_request_target:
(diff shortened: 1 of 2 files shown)


andrewsomething:
Note that `pull_request_target` is expected to fail as it does not have the changes from this branch.

ChiefMateStarbuck (approved):
LGTM

PR #591, opened 2021-02-26
Adds new firewall datasource

### LINKED ISSUE #188: Add digitalocean_firewall datasource
### Terraform Version
```
Terraform v0.11.11
+ provider.digitalocean v1.1.0
```

### Affected Resource(s)
- digitalocean_firewall

### Use case

I'd like to add some rules to the firewall that is automatically created with a `digitalocean_kubernetes_cluster` resource (to fix this bug: https://www.digitalocean.com/community/questions/using-do-managed-kubernetes-cluster-with-helm-chart-stable-prometheus-results-in-some-node_exporters-being-unreachable).

I found a way using `terraform import` mentioned in the docs ( https://github.com/terraform-providers/terraform-provider-digitalocean/blob/master/website/docs/r/firewall.html.markdown#import) but it doesn't allow me to create my infrastructure with only one `terraform apply`, and, the `digitalocean_firewall.id` is not exposed in the state, which doesn't allow me to import it without having to rely on `doctl` and tags.

Is there any other way of importing the resource programmatically within my first `apply` step? Would it work if I add another firewall and associate it with my cluster droplets?

I'm unsure if a datasource for `digitalocean_firewall` would help but still, it misses.

Thanks for your insights.

scotchneat:
*Opening this as a draft for feedback first. 
Pending docs

Closes #188

### DIFF
--- a/digitalocean/datasource_digitalocean_firewalls.go
+++ b/digitalocean/datasource_digitalocean_firewalls.go
@@ -1,0 +1,16 @@
+package digitalocean
+
+import (
+	"github.com/digitalocean/terraform-provider-digitalocean/internal/datalist"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceDigitalOceanFirewalls() *schema.Resource {
+	dataListConfig := &datalist.ResourceConfig{
+		RecordSchema:        firewallSchema(),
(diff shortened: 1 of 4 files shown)


scotchneat:
Closing. Will resubmit when this solution is more mature and ready for feedback.

PR #594, opened 2021-03-01
Adds Firewall datasource

### LINKED ISSUE #188: Add digitalocean_firewall datasource

scotchneat:
* Moved the firewall schema to `firewalls.go` to share between the resource and datasource.
* Added a test to create and update the datasource.
* (unrelated) Removed `-i` flag from `test` make target since it's been deprecated (and caused failures locally)

Closes #188


digitalocean/datasource_digitalocean_firewall_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccDataSourceDigitalOceanFirewall_Basic(t *testing.T) {
+	fwDataConfig := `
+data "digitalocean_firewall" "foobar" {
+	firewall_id = digitalocean_firewall.foobar.id
+	name = digitalocean_firewall.foobar.name
```
andrewsomething, line 16: We shouldn't require users to enter both the name and id. All we really need is the id as we can get the name with it.

digitalocean/datasource_digitalocean_firewall.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+func dataSourceDigitalOceanFirewall() *schema.Resource {
+	fwSchema := firewallSchema()
+
+	for _, f := range fwSchema {
+		if !f.Required {
+			f.Computed = true
+		}
+	}
```
andrewsomething, line 16:
```suggestion
	for _, f := range fwSchema {
		f.Computed = true
		f.Required = false
	}

	fwSchema["name"].ValidateFunc = nil

	fwSchema["firewall_id"] = &schema.Schema{
		Type:     schema.TypeString,
		Required: true,
	}
```

In order to make `name` not required, we can make some additional adjustments to the schema in the data source. We can mark everything as `Computed = true` and `Required = false` as their values will be assigned from the API. We also set the `ValidateFunc` for `name` to `nil` as the SDK complains that "ValidateFunc is for validating user input, there's nothing to validate on computed-only field."

Since `firewall_id` is only needed for the data source, we can add it here instead of the main schema and make it required.

digitalocean/firewalls.go
```diff
@@ -0,0 +1,322 @@
...
+					},
+				},
+			},
+		},
+
+		"tags": tagsSchema(),
+	}
+}
+
+func firewallRuleSchema(prefix string) *schema.Resource {
```
andrewsomething, line 76: :ok_hand:  Great to reduce some of the duplication.

digitalocean/datasource_digitalocean_firewall_test.go
```diff
@@ -0,0 +1,86 @@
...
+						"data.digitalocean_firewall.foobar", "created_at"),
+					resource.TestCheckResourceAttrPair("digitalocean_firewall.foobar", "pending_changes",
+						"data.digitalocean_firewall.foobar", "pending_changes"),
+					resource.TestCheckResourceAttrPair("digitalocean_firewall.foobar", "tags",
+						"data.digitalocean_firewall.foobar", "tags"),
+					func(n string) resource.TestCheckFunc {
+						return func(s *terraform.State) error {
+							r, _ := s.RootModule().Resources[n]
+							t.Logf("RESOURCE: %v", r)
+							return nil
```
andrewsomething, line 59: Was this for debugging?

digitalocean/firewalls.go
```diff
@@ -0,0 +1,322 @@
...
+
+import (
+	"github.com/digitalocean/godo"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+func firewallSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"firewall_id": {
```
andrewsomething, line 11: If we make the other changes I suggestion, this can be removed from the share schema.

andrewsomething (changes requested):
This looks great!

I made a few suggestions inline around making it so that both the name and id are not required for the data source.

We'll also need docs, but that can be a fast follow.

digitalocean/datasource_digitalocean_firewall.go
scotchneat: Ah. I see.

digitalocean/datasource_digitalocean_firewall_test.go
scotchneat: whoops yes lol.

digitalocean/datasource_digitalocean_firewall.go
```diff
@@ -0,0 +1,27 @@
+package digitalocean
+
+import (
+	"context"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func dataSourceDigitalOceanFirewall() *schema.Resource {
```
ChiefMateStarbuck, line 9: why return a pointer here? Wouldn't be simpler just to return an actual value?

digitalocean/datasource_digitalocean_firewall_test.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"fmt"
+	"github.com/digitalocean/godo"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccDataSourceDigitalOceanFirewall_Basic(t *testing.T) {
```
ChiefMateStarbuck, line 12: Great tests.

digitalocean/datasource_digitalocean_firewall.go
scotchneat: The place that references it (DataSourcesMap on the provider in `digitalocean/provider.go`) is a `map[string]*schema.Resource` and I followed the pattern all the other functions in the provider have already implemented.

andrewsomething: Right, it needs to implement https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/schema#Provider

docs/data-sources/firewall.md
```diff
@@ -0,0 +1,84 @@
...
+* `name` - The name of the Firewall.
+* `droplet_ids` - The list of the IDs of the Droplets assigned to
+  the Firewall.
+* `tags` - The names of the Tags assigned to the Firewall.
+* `inbound_rules` - The inbound access rule block for the Firewall.
+* `outbound_rules` - The outbound access rule block for the Firewall.
+
+`inbound_rule` supports the following:
+
+* `protocol` - (Required) The type of traffic to be allowed.
```
andrewsomething, line 50:
```suggestion
* `protocol` - The type of traffic to be allowed.
```

Probably makes sense to remove the optional/required bits for these since it is all read only.

andrewsomething (dismissed):
:+1: Looks great!

I did leave one suggestion about the docs inline.

docs/data-sources/firewall.md
scotchneat: Got it

andrewsomething (approved):
:shipit:

PR #596, opened 2021-03-09
Adds support for kubernetes node pool taints

### LINKED ISSUE #374: Add support for node taint
Hi.

I think will be really useful to have support for taints: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/

At the moment I want to start some nodes which accepts only specific pods to be run.

Thx.

scotchneat:
Includes some refactoring to allow schema reuse.
* moved node pool related schemas to `digitalocean/kubernetes.go`

This allows the `taints` attribute to be exposed to both the `kubernetes_cluster` resource and datasource, as well as the `kubernetes_node_pool` resource.

I'd like some feedback about the attribute name: `taint`. Since this represents a list, I can see this being confusing and wanted to know how others feel about making this plural (would match the API also). Although, the singular tense makes the UI in terraform config files much clearer.

closes #374


andrewsomething (changes requested):

docs/resources/kubernetes_cluster.md
```diff
@@ -173,12 +179,17 @@ In addition to the arguments listed above, the following additional attributes a
...
-     + `droplet_id` - The id of the node's droplet
-     + `created_at` - The date and time when the node was created.
-     + `updated_at` - The date and time when the node was last updated.
+    + `id` -  A unique ID that can be used to identify and reference the node.
+    + `name` - The auto-generated name for the node.
+    + `status` -  A string indicating the current status of the individual node.
+    + `droplet_id` - The id of the node's droplet
+    + `created_at` - The date and time when the node was created.
+    + `updated_at` - The date and time when the node was last updated.
+  - `taint` - A list of taints applied to all nodes in the pool. Each taint exports the following attributes:
```
andrewsomething, line 188: > I'd like some feedback about the attribute name: taint. Since this represents a list, I can see this being confusing and wanted to know how others feel about making this plural (would match the API also). Although, the singular tense makes the UI in terraform config files much clearer.

I think singular works fine here. As you mention it feels more natural in a Terraform config, e.g.:

```
    taint {
      key    = "workloadKind"
      value  = "database"
      effect = "NoSchedule"
    }

    taint {
      key    = "priority"
      value  = "high"
      effect = "NoSchedule"
    }
```

Maybe the solution is to reword the docs a bit?

```suggestion
  - `taint` - A block representing a taint applied to all nodes in the pool. Each taint exports the following attributes:
```

aybabtme:
And here I was about to implement this in https://github.com/aybabtme/terraform-provider-digitalocean/commit/51e3dbdb1f57c1b6f48855120729d2fc64e2ff2a. Looking forward to this getting released!

amilcaru123 (dismissed):

digitalocean/resource_digitalocean_kubernetes_node_pool.go
```diff
@@ -387,6 +244,11 @@ func digitaloceanKubernetesNodePoolUpdate(client *godo.Client, pool map[string]i
 		req.Labels = expandLabels(pool["labels"].(map[string]interface{}))
 	}
 
+	if pool["taint"] != nil {
```
andrewsomething, line 247: This will always be nil here. This is called from `resourceDigitalOceanKubernetesNodePoolUpdate`. The taints must also be set there. 

https://github.com/digitalocean/terraform-provider-digitalocean/blob/9ee1da849a561a4619ddacd37bd717e12058b303/digitalocean/resource_digitalocean_kubernetes_node_pool.go#L93-L110

andrewsomething (changes requested):

amilcaru123 (approved):

digitalocean/resource_digitalocean_kubernetes_node_pool_test.go
```diff
@@ -15,18 +15,75 @@ func TestAccDigitalOceanKubernetesNodePool_Basic(t *testing.T) {
...
+		name = "default"
+		size  = "s-1vcpu-2gb"
+		node_count = 1
+		tags  = ["one","two"]
+	}
+}
+`, testClusterVersion19, rName)
+
+	nodePoolConfig := fmt.Sprintf(`resource digitalocean_kubernetes_node_pool "barfoo" {
+	cluster_id = "${digitalocean_kubernetes_cluster.foobar.id}"
```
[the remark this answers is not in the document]
scotchneat: Thanks! I should have caught these.

amilcaru123 (approved):

andrewsomething (approved):
:+1:  Looks good!

aybabtme:
🥳 thanks for implementing this!

PR #601, opened 2021-03-11
doc: add dependency in k8s example

pyrho:
Hey,

I hope I'm not mistaken I'm by no means an expert.

Cheers.

### DIFF
--- a/examples/kubernetes/kubernetes-config/main.tf
+++ b/examples/kubernetes/kubernetes-config/main.tf
@@ -16,6 +16,7 @@ terraform {
 }
 
 data "digitalocean_kubernetes_cluster" "primary" {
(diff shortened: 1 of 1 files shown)


scotchneat:
@pyrho we really appreciate the contribution!

Please note that the config file you referenced is actually a module of the parent folder. See here:
https://github.com/digitalocean/terraform-provider-digitalocean/blob/main/examples/kubernetes/main.tf#L36

While it's possible to run that config on its own, it would require an existing cluster, and to pass the required variables to terraform as terraform apply -var cluster_id=$CLUSTER_ID -var cluster_name=$CLUSTER_NAME.

The intention with these examples is that the full environment is managed at the examples/kubernetes folder level.

PR #602, opened 2021-03-13
Update the statement below the autoscaling example

3dinfluence:
The existing statement is not accurate as specifying a node pool
min_nodes to 0 will result in the following error.

```
422 validation error: worker_node_pool_spec.min nodes must be greater than or equal to 1
```
For clarity sake I got this error when adding a 2nd node pool with autoscaling to an existing cluster.  So the cluster itself had other nodes.


andrewsomething (approved):
:+1: Good catch. Thanks for the contribution!

Issue #489, opened 2020-09-11
Add digitalocean_database_replica datasource


nathancolgate:
### Terraform Version

```
Terraform v0.12.29
+ provider.digitalocean v1.22.2
```

### Affected Resource(s)

- digitalocean_database_replica

### (Desired) Terraform Configuration Files

```hcl
data "digitalocean_database_replica" "read_only_db" {
  name = "project-read-only"
}
```

### Use Case

Our databases and replicas are created at a shared level, outside our project-specific terraform.

We are able to look up the database and get the attributes to use in our project-specific terraform.

But we can not grab the read-only replica host within the project.

I believe [the DigitalOcean API](https://developers.digitalocean.com/documentation/v2/#retrieve-an-existing-read-only-replica) already supports this.

andrewsomething:
Closed via https://github.com/digitalocean/terraform-provider-digitalocean/pull/608

PR #608, opened 2021-03-23
Adding db replica datasource #489

danaelhe:
running into a strange "id" attribute issue when testing it


digitalocean/datasource_digitalocean_database_replica.go
```diff
@@ -0,0 +1,110 @@
...
+				},
+				Set: HashStringIgnoreCase,
+			},
+		},
+	}
+}
+
+func dataSourceDigitalOceanDatabaseReplicaRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	d.Set("cluster_id", d.Get("cluster_id").(string))
+	d.Set("name", d.Get("name").(string))
```
andrewsomething, line 108:
```suggestion
	clusterId :=  d.Get("cluster_id").(string)
	name := d.Get("name").(string)
	d.Set("cluster_id", clusterId)
	d.Set("name", name)
	d.SetId(makeReplicaId(clusterId, name))
```

You'll need to set a unique ID for the data source. We already have a little helper to make one from the cluster ID and replica name:

https://github.com/digitalocean/terraform-provider-digitalocean/blob/c73e7a5cf4fbe88060c471a808acf862c327b893/digitalocean/resource_digitalocean_database_replica.go#L208-L210

digitalocean/datasource_digitalocean_database_replica_test.go
```diff
@@ -0,0 +1,82 @@
...
+				),
+			},
+			{
+				Config: databaseConfig + replicaConfig + datasourceReplicaConfig,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrPair("digitalocean_database_replica.read-01", "cluster_id",
+						"digitalocean_database_replica.my_db_replica", "cluster_id"),
+					resource.TestCheckResourceAttrPair("digitalocean_database_replica.read-01", "name",
+						"digitalocean_database_replica.my_db_replica", "name"),
+				),
```
andrewsomething, line 72:
```suggestion
					resource.TestCheckResourceAttrPair("digitalocean_database_replica.read-01", "host",
						"data.digitalocean_database_replica.my_db_replica", "host"),
								),
```

Would be great to test that the computed attributes are set as expected. In fact, it's more important to test that here than in the step above which is really just setting things up for this test.

digitalocean/datasource_digitalocean_database_replica_test.go
```diff
@@ -0,0 +1,82 @@
...
+						"digitalocean_database_replica.read-01", "private_network_uuid"),
+				),
+			},
+			{
+				Config: databaseConfig + replicaConfig + datasourceReplicaConfig,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrPair("digitalocean_database_replica.read-01", "cluster_id",
+						"digitalocean_database_replica.my_db_replica", "cluster_id"),
+					resource.TestCheckResourceAttrPair("digitalocean_database_replica.read-01", "name",
+						"digitalocean_database_replica.my_db_replica", "name"),
```
andrewsomething, line 71:
```suggestion
					resource.TestCheckResourceAttrPair("digitalocean_database_replica.read-01", "cluster_id",
						"data.digitalocean_database_replica.my_db_replica", "cluster_id"),
					resource.TestCheckResourceAttrPair("digitalocean_database_replica.read-01", "name",
						"data.digitalocean_database_replica.my_db_replica", "name"),
```

To reference an attribute of a data source, you needed to specify `data.<type>.<name>`

digitalocean/datasource_digitalocean_database_replica.go
scotchneat: Ah nice.

digitalocean/datasource_digitalocean_database_replica_test.go
scotchneat: We were troubleshooting the id issue but this was next

digitalocean/datasource_digitalocean_database_replica.go
danaelhe: thanks andrew!

scotchneat:
Looks good @danaelhe!
Once you merge the latest changes, I'll approve.

danaelhe:
> Looks good @danaelhe!
> Once you merge the latest changes, I'll approve.

hmm. It appears waiting 10 seconds won't cut it anymore. Waiting 10 seconds was consistently passing the other day. Will have to run a few more tests.

digitalocean/datasource_digitalocean_database_replica_test.go
```diff
@@ -0,0 +1,98 @@
...
+			},
+			{
+				Config: databaseConfig + replicaConfig + datasourceReplicaConfig,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttrPair("digitalocean_database_replica.read-01", "cluster_id",
+						"data.digitalocean_database_replica.my_db_replica", "cluster_id"),
+					resource.TestCheckResourceAttrPair("digitalocean_database_replica.read-01", "name",
+						"data.digitalocean_database_replica.my_db_replica", "name"),
+					// resource.TestCheckResourceAttr(
+					// 	"data.digitalocean_database_replica.my_db_replica", "size", "db-s-2vcpu-4gb"),
```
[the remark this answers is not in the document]
danaelhe: Thanks! Will make a note to reach out to DBaaS.

digitalocean/datasource_digitalocean_database_replica.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+func dataSourceDigitalOceanDatabaseReplica() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceDigitalOceanDatabaseReplicaRead,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
```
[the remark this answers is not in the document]
danaelhe: Ah i see. I've removed them from all attributes.

digitalocean/datasource_digitalocean_database_replica_test.go
```diff
@@ -0,0 +1,96 @@
...
+				),
+			},
+			{
+				Config: databaseConfig + replicaConfig,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckDigitalOceanDatabaseReplicaExists("digitalocean_database_replica.read-01", &databaseReplica),
+					testAccCheckDigitalOceanDatabaseReplicaAttributes(&databaseReplica, databaseReplicaName),
+					resource.ComposeTestCheckFunc(
+						func(s *terraform.State) error {
+							time.Sleep(30 * time.Second)
```
andrewsomething, line 49: Is the sleep needed here? Would have thought that once the replica is successfully created, the data source should be able to reach it without the sleep.

andrewsomething (approved):
:+1:  This looks good! 

A couple more comments inline. We'll also need a page for the docs. Here's an example for a data source:

https://github.com/digitalocean/terraform-provider-digitalocean/blob/main/docs/data-sources/database_cluster.md

You can probably reuse most of the descriptions from the resource: https://github.com/digitalocean/terraform-provider-digitalocean/blob/main/docs/resources/database_replica.md

This can be a fast follow if you'd like to get this merged.

digitalocean/datasource_digitalocean_database_replica_test.go
danaelhe: Oh I had that in for testing and was meant to remove it. good catch!

danaelhe:
@andrewsomething I took a crack at writing out the documentations and I had a few questions:
I only put cluster_id and name as required, and kept tags as optional. I figured the rest (as compared to the resource database replica doc) would be computed?

I'm not 100% sure about the import part either.

docs/data-sources/database_replica.md
```diff
@@ -0,0 +1,69 @@
...
+  region     = "nyc3"
+  size       = "db-s-2vcpu-4gb"
+  tags       =  ["staging"]
+}
+
+# Create data source of replica
+data "digitalocean_database_replica" "my_db_replica" {
+  cluster_id = digitalocean_database_cluster.foobar.id
+  name       = "terra-test-ro"
+}
```
andrewsomething, line 35: Generally the use case for data sources is to access information about a resource that is not managed in Terraform. So rather than creating the cluster and replica here, a more real world example would probably just have have the data sources perhaps outputting some connection info. 

```suggestion
data "digitalocean_database_cluster" "example" {
  name = "example-cluster"
}

data "digitalocean_database_replica" "read-only" {
  cluster_id = data.digitalocean_database_cluster.example.id
  name       = "terra-test-ro"
}

output "replica_output" {
  value = data.digitalocean_database_replica.read-only.uri
}
```

andrewsomething (commented):
> I only put cluster_id and name as required, and kept tags as optional. I figured the rest (as compared to the resource database replica doc) would be computed?

`tags` can also be computed. The only user input is `cluster_id` and `name`.

> I'm not 100% sure about the import part either.

You got it. Importing is for resources, not data sources. Importing brings a resource created outside of Terraform (e.g. the control panel) under management by Terraform. Data sources are inherently read-only, so there is no need for importing them.

https://www.terraform.io/docs/cli/import/index.html

docs/data-sources/database_replica.md
danaelhe: ah yes. that makes sense! thanks.

ChiefMateStarbuck (commented):
Looks good! lets give @andrew the final say ;)

andrewsomething (approved):
LGTM :tada:

PR #614, opened 2021-04-05
database cluster: Handle MongoDB password differences.

andrewsomething:
Support for the MongoDB beta.

MongoDB clusters only return their password in response to the initial POST. This means they need to be handled differently than other database clusters:

- We need to set the password to state directly in the create function before doing any polling for status.
- As the host for the cluster is not known until it becomes available, we must save the password and then add it to the URL returned latter to build a usable connection URI.

### DIFF
--- a/digitalocean/resource_digitalocean_database_cluster.go
+++ b/digitalocean/resource_digitalocean_database_cluster.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


digitalocean/resource_digitalocean_database_cluster.go
```diff
@@ -503,3 +509,58 @@ func flattenMaintWindowOpts(opts godo.DatabaseMaintenanceWindow) []map[string]in
...
+		}
+	}
+
+	return nil
+}
+
+// MongoDB clusters only return their password in response to the initial POST.
+// The host for the cluster is not known until it becomes available. In order to
+// build a usable connection URI, we must save the password and then add it to
+// the URL returned latter.
```
ChiefMateStarbuck, line 554: is this a specific mongodb thing?

ChiefMateStarbuck (approved):
awesome!

digitalocean/resource_digitalocean_database_cluster.go
andrewsomething: Yes

PR #616, opened 2021-04-15
Support distributing custom images to multiple regions.

andrewsomething:
When the `digitalocean_custom_image` resource was initially implemented, support for distributing images to more than one region was left as a TODO. This finishes it off. A few things to note:

* An image can only be imported to a single region. It can be distributed to other regions when the initial import is complete.
* DigitalOcean does not currently support removing/deleting an image from a region. So we must `ForceNew` if a region is removed from the list.


ChiefMateStarbuck (approved):

scotchneat (approved):
lgtm!

Issue #617, opened 2021-04-18
Spaces creation fails: "no such host"


matthewhembree:
Hello!

### Terraform Version
```
Terraform v0.15.0
on darwin_amd64
+ provider registry.terraform.io/digitalocean/digitalocean v2.7.0
```

### Affected Resource(s)
Please list the resources as a list, for example:
- resource: digitalocean_spaces_bucket
- data: digitalocean_spaces_bucket

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.
No issues with:
- resource: digitalocean_droplet

### Terraform Configuration Files
```hcl
terraform {
  required_providers {
    digitalocean = {
      source = "digitalocean/digitalocean"
      version = "2.7.0"
    }
  }
  required_version = "0.15.0"
}

provider "digitalocean" {
  token = var.digitalocean_token
  spaces_access_id = var.spaces_access_id
  spaces_secret_key = var.spaces_secret_key
}

resource "digitalocean_spaces_bucket" "i-dont-exist" {
  name = "i-dont-exist"
  region = "ny3"
}

data "digitalocean_spaces_bucket" "i-already-exist" {
  name = "i-already-exist"
  region = "ny3"
}

variable "digitalocean_token" {
  type = string
  description = "Digital Ocean API token"
  sensitive = true
}

variable "spaces_access_id" {
  type = string
  description = "Digital Ocean Spaces access ID"
  sensitive = true
}

variable "spaces_secret_key" {
  type = string
  description = "Digital Ocean Spaces secret key"
  sensitive = true
}

output "i-already-exist_attrs" {
  value = data.digitalocean_spaces_bucket.i-already-exist
}
```

### Debug Output
[trace.log](https://gist.github.com/matthewhembree/3235038c0f11aa761ac82610db7c8634#file-trace-log)
[refresh-trace.log](https://gist.github.com/matthewhembree/3235038c0f11aa761ac82610db7c8634#file-refresh-trace-log)

### Panic Output
n/a

### Expected Behavior
- Space created.
- Existing bucket attributes output.

### Actual Behavior
- Space not created.
- On `tf refresh`: error `no such host`

DNS lookup error on `tf apply`:
```
╷
│ Error: Error creating Spaces bucket: RequestError: send request failed
[...]

PR #618, opened 2021-04-19
spaces: Validate input for region.

andrewsomething:
Do some input validation on the region for Spaces operations as the `no such host` error message is not helpful. The new error would be:

```
Error: expected region to be one of [nyc3 sfo2 sfo3 ams3 sgp1 fra1], got ny2

  on main.tf line 21, in resource "digitalocean_spaces_bucket" "i-dont-exist":
  21:   region = "ny2"
```

Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/617

### DIFF
--- a/digitalocean/spaces_buckets.go
+++ b/digitalocean/spaces_buckets.go
@@ -7,6 +7,11 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+var (
+	// SpacesRegions is a list of DigitalOcean regions that support Spaces.
+	SpacesRegions = []string{"ams3", "fra1", "nyc3", "sfo2", "sfo3", "sgp1"}
+)
+
 type bucketMetadataStruct struct {
 	name   string
 	region string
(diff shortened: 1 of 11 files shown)


digitalocean/spaces_buckets.go
```diff
@@ -7,6 +7,11 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+var (
+	// SpacesRegions is a list of DigitalOcean regions that support Spaces.
+	SpacesRegions = []string{"ams3", "fra1", "nyc3", "sfo2", "sfo3", "sgp1"}
```
ChiefMateStarbuck, line 12: should we make an enum file for these values? Dynamically loading from an enum file can make updating said file much easier.

ChiefMateStarbuck (changes requested):

digitalocean/spaces_buckets.go
andrewsomething: For use in a `ValidateFunc` we'd still need to return a list of strings. Feels a bit heavy to me.

```go
type SpacesRegion int

const (
	SpacesRegionAMS3 SpacesRegion = iota
	SpacesRegionFRA1
	SpacesRegionNYC3
	SpacesRegionSFO2
	SpacesRegionSFO3
	SpacesRegionSGP1
)

func (s SpacesRegion) String() string {
	return [...]string{"ams3", "fra1", "nyc3", "sfo2", "sfo3", "sgp1"}[s]
}

func SpacesRegionValues() []string {
	return []string{
		SpacesRegionAMS3.String(),
		SpacesRegionFRA1.String(),
		SpacesRegionNYC3.String(),
		SpacesRegionSFO2.String(),
		SpacesRegionSFO3.String(),
		SpacesRegionSGP1.String(),
	}
}
```

Or the slightly simplified but not really an enum approach:

```go
const (
	SpacesRegionAMS3 = "ams3"
	SpacesRegionFRA1 = "fra1"
	SpacesRegionNYC3 = "nyc3"
	SpacesRegionSFO2 = "sfo2"
	SpacesRegionSFO3 = "sfo3"
	SpacesRegionSGP1 = "sgp1"
)

func SpacesRegionValues() []string {
	return []string{
		SpacesRegionAMS3,
		SpacesRegionFRA1,
		SpacesRegionNYC3,
		SpacesRegionSFO2,
		SpacesRegionSFO3,
		SpacesRegionSGP1,
	}
}
```

If we were building out a library for working with Spaces (i.e. adding Spaces support to godo) it might be worth it. Not so sure as an internal convenience. Maybe it just shouldn't be exported?

ChiefMateStarbuck: both of these seem to much work for a simple slice that's used only once of twice. LGTM

scotchneat (approved):
👍🏽

Issue #622, opened 2021-04-22
Update the end to end Droplet example


andrewsomething:
This project includes an example of creating a Droplet, installing nginx on it, and assigning a domain to it:

https://github.com/digitalocean/terraform-provider-digitalocean/tree/main/examples/droplet

It was written years ago and it no longer represents Terraform best practices. In fact, it does not even run as written!

Since it was originally written, Terraform's configuration language has undergone a number of changes. When updating the example, Terraform's upgrade guides may be a good resource:

https://www.terraform.io/upgrade-guides/index.html

Beyond syntax changes, the example defaults to using `ubuntu-16-04-x64`. A more recent version of Ubuntu would be preferable.

Feel free to look for additional areas for improvement.

scotchneat:
The docs published to the Terraform registry need to be updated as well - https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs#example-usage

Issue #595, opened 2021-03-04
Can't run `apt` immediately at script startup unless I wait


dwilliamson:
### Terraform Version
Terraform v0.14.7
+ provider registry.terraform.io/digitalocean/digitalocean v1.22.2

### Affected Resource(s)
`digitalocean_droplet`

### Terraform Configuration Files
This is from the start, which is enough to repro the bug
```hcl
terraform {
  required_providers {
    digitalocean = {
      source = "digitalocean/digitalocean"
      version = "1.22.2"
    }
  }
  required_version = ">= 0.13"
}

provider "digitalocean" {
    token = var.digitalocean_api_token
}

resource "digitalocean_ssh_key" "default" {
    name = "root@rendezvous SSH Public Key"
    public_key = file(var.pubkey_root_path)
}

resource "digitalocean_droplet" "rendezvous" {
    name = "rendezvous"
    region = "nyc1"
    size = "s-1vcpu-1gb"
    image = "debian-10-x64"
    monitoring = true
    ssh_keys = [digitalocean_ssh_key.default.id]

    connection {
        host = self.ipv4_address
        user = "root"
        type = "ssh"
        agent = true
        timeout = "2m"
    }

    provisioner "remote-exec" {
        inline = [
            "sudo apt update",
            "apt-get install ufw",
        ]
    }
}
```

### Expected Behavior
Terraform apply completes successfully.

### Actual Behavior
The key problem is with `sudo apt update` and `apt-get install ufw` in the `remote-exec` provisioner. These randomly fail with:

```
E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
```

When I add `ps aux | grep [a]pt` at startup I can see a whole bunch of processes running `apt`. A small example:

```
[...]

PR #625, opened 2021-04-26
Update nginx example to work

atombrella:
I couldn't see CoreOS in the list under the droplets anymore. Because the droplet is created with IPv6, I'd like to encourage people to support it better with a [AAAA DNS](https://tools.ietf.org/html/rfc3596#section-2) record. 

Please see #622


examples/droplet/main.tf
```diff
@@ -1,3 +1,11 @@
+terraform {
+  required_providers {
+    digitalocean = {
+      source = "digitalocean/digitalocean"
```
atombrella, line 4: I'm not sure if the minimum version attribute is strictly required for this. This just downloads the most current version (2.8), it seems.

atombrella:
@andrewsomething @scotchneat Let me know what more is needed to get this merged.

examples/droplet/main.tf
scotchneat: I think it's a good idea to update this when examples are updated.

examples/droplet/main.tf
```diff
@@ -36,9 +46,16 @@ resource "digitalocean_domain" "mywebserver" {
...
   type   = "A"
   name   = "mywebserver"
   value  = "${digitalocean_droplet.mywebserver.ipv4_address}"
 }
+
+resource "digitalocean_record" "mywebserver_ipv6_dns" {
+  domain = "${digitalocean_domain.mywebserver.name}"
+  type   = "AAAA"
+  name   = "mywebserver"
+  value  = "${digitalocean_droplet.mywebserver.ipv6_address}"
```
scotchneat, line 60: like here

```suggestion
  value  = digitalocean_droplet.mywebserver.ipv6_address
```

examples/droplet/outputs.tf
```diff
@@ -1,7 +1,11 @@
-output "Public ip" {
+output "IPv4" {
   value = "${digitalocean_droplet.mywebserver.ipv4_address}"
 }
 
+output "IPv6" {
+  value = "${digitalocean_droplet.mywebserver.ipv6_address}"
```
scotchneat, line 6: here

```suggestion
  value = digitalocean_droplet.mywebserver.ipv6_address
```

examples/droplet/main.tf
```diff
@@ -23,6 +32,7 @@ resource "digitalocean_droplet" "mywebserver" {
     ]
 
     connection {
+      host     = self.ipv4_address
       type     = "ssh"
       private_key = "${file("~/.ssh/id_rsa")}"
```
scotchneat, line 37: While this isn't part of your change, I think this is a good opportunity to clear that deprecation warning

```suggestion
      private_key = file("~/.ssh/id_rsa")
```

scotchneat (changes requested):
The change is great @atombrella!

This isn't a blocker but since the release of [Terraform 0.12.14](https://discuss.hashicorp.com/t/terraform-0-12-14-released/3898), `Interpolation-only expressions are deprecated`. This applies to the `${xxx}` interpolation syntax within quotes but without any other text within the quotes. 

Although it's still a warning, I think we could take this opportunity to address it. It can be resolved by removing the quotes. I included some suggestions inline but didn't cover all the instances. You can check if you still get the warning (as long as you're using terraform 0.12.14 or newer - I suggest >= 0.14) by running `terraform plan` on this example locally.

atombrella:
@scotchneat Thanks :) The PR has been updated. My experience with Terraform is a bit limited, but I have used `terraform plan` and `terraform apply` before on some similar small setups in DigitalOcean.

Please adjust the commit message if needed. I couldn't really see a pattern for changes of this type.

danaelhe:
> @scotchneat No warnings with tflint :) I won't make anymore changes to this PR.

Looks great! I added a minor change to fix a lock issue I ran into. It's the same problem #595 addresses.

examples/droplet/variable.tf
```diff
@@ -65,15 +65,10 @@ variable "do_tor1" {
...
 variable "centos" {
   description = "Default Centos"
-  default     = "centos-72-x64"
-}
-
-variable "coreos" {
-  description = "Defaut Coreos"
-  default     = "coreos-899.17.0"
+  default     = "centos-83-x64"
 }
```
scotchneat, line 74: Adding an input variable with a common default that can be overwritten during an apply: `terraform apply -var 'ssh_key_path=~/.ssh/some_other_key.pub'`

```suggestion
}

variable "ssh_key_path" {
  type        = string
  description = "The file path to an ssh public key"
  default     = "~/.ssh/id_rsa.pub"
}
```

examples/droplet/main.tf
```diff
@@ -6,39 +15,40 @@ provider "digitalocean" {
...
+  connection {
+      host     = self.ipv4_address
       type     = "ssh"
-      private_key = "${file("~/.ssh/id_rsa")}"
+      private_key = file("~/.ssh/id_rsa")
       user     = "root"
       timeout  = "2m"
     }
-  }
 }
```
scotchneat, line 36: Adding an ssh key resource allows the user to upload their real ssh key based on the input variable.
```suggestion

resource "digitalocean_ssh_key" "example" {
  name       = "examplekey"
  public_key = file(var.ssh_key_path)
}

```

examples/droplet/main.tf
```diff
@@ -6,39 +15,40 @@ provider "digitalocean" {
 
 resource "digitalocean_droplet" "mywebserver" {
   # Obtain your ssh_key id number via your account. See Document https://developers.digitalocean.com/documentation/v2/#list-all-keys
-  ssh_keys           = [12345678]         # Key example
-  image              = "${var.ubuntu}"
-  region             = "${var.do_ams3}"
+  ssh_keys           = [29834845]        # Key example
```
scotchneat, line 18: This will allow the example to work with a real ssh key instead of hard coding an example key that wouldn't be on the user's account

```suggestion
  ssh_keys           = [digitalocean_ssh_key.example.fingerprint]  
```

scotchneat (changes requested):
I'd love to merge the change as is since it covers the original purpose of the change.
However, there's one more change that would allow this example to run from a clean clone. 

Currently, the ssh key is hardcoded to a specific key id. This will fail until the user updates the key. I've added some small suggestions inline that will make this dynamic and should allow users to run the example straight away.

danaelhe (approved):
lgtm! Thank you so much!

scotchneat (approved):
LGTM! Thanks again for the contribution

MattIPv4:
Hey, @atombrella - Thanks a ton for this sweet PR! 😄

PR #631, opened 2021-05-02
Add support for Kubernetes maintenance policy

atombrella:
I'm not too familiar with GoLang, but maybe this is a good opportunity to work a bit with it.


atombrella:
@scotchneat Could you please have a look to see if this is moving in the right direction? Thanks.

scotchneat:
Hi @atombrella, thanks for your patience. I'll take a look at it shortly.

scotchneat:
@atombrella I just wanted to drop an update to let you know we haven't forgotten about your contribution. 

While reviewing it, I noticed some strange behavior with the tests which relate to the outdated SDK version of Terraform we're using. We're looking into this now and will likely have to wait for that to be resolved before proceeding with this change.

atombrella:
> @atombrella I just wanted to drop an update to let you know we haven't forgotten about your contribution.
> 
> While reviewing it, I noticed some strange behavior with the tests which relate to the outdated SDK version of Terraform we're using. We're looking into this now and will likely have to wait for that to be resolved before proceeding with this change.

Good :) I think there's still work that needs to be done, so I'm primarily hoping for a few hints on what I'm missing. In particular, regarding the unit tests.

digitalocean/resource_digitalocean_kubernetes_cluster.go
```diff
@@ -345,6 +368,8 @@ func resourceDigitalOceanKubernetesClusterUpdate(ctx context.Context, d *schema.
 			Tags:         expandTags(d.Get("tags").(*schema.Set).List()),
 			AutoUpgrade:  godo.Bool(d.Get("auto_upgrade").(bool)),
 			SurgeUpgrade: d.Get("surge_upgrade").(bool),
+
+			MaintenancePolicy: d.Get("maintenance_policy").(*godo.KubernetesMaintenancePolicy),
```
andrewsomething, line 372: This also needs to call `expandMaintPolicyOpts`.

Also, this line above will need to updated to for the addition:

https://github.com/digitalocean/terraform-provider-digitalocean/blob/24279405c3aca60b167d8531e8332cfa746a0b6b/digitalocean/resource_digitalocean_kubernetes_cluster.go#L364

The SDK has actually grown a new method recently that can make this cleaner:

https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/schema#ResourceData.HasChanges

So it can now look like:

```go
if d.HasChanges("name", "tags", "auto_upgrade", "surge_upgrade", "maintenance_policy") {
```

digitalocean/resource_digitalocean_kubernetes_cluster.go
```diff
@@ -87,6 +87,27 @@ func resourceDigitalOceanKubernetesCluster() *schema.Resource {
...
+						},
+						"duration": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"start_time": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
```
andrewsomething, line 107:
```suggestion
			"maintenance_policy": {
				Type:     schema.TypeList,
				MinItems: 1,
				MaxItems: 1,
				Optional: true,
				Elem: &schema.Resource{
					Schema: map[string]*schema.Schema{
						"day": {
							Type:     schema.TypeString,
							Optional: true,
						},
						"start_time": {
							Type:     schema.TypeString,
							Optional: true,
						},
						"duration": {
							Type:     schema.TypeString,
							Computed: true,
						},
					},
```

Generally, you want to use `Computed: true` for attributes that the user can not configure. So it is correct to set that for `duration` since it is read only, we don't want to set that here for `day` or `start_time`.

https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#computed

It's probably a good idea to set `MinItems` and `MaxItems` as well.

digitalocean/resource_digitalocean_kubernetes_cluster_test.go
```diff
@@ -180,6 +181,41 @@ func TestAccDigitalOceanKubernetesCluster_UpdateCluster(t *testing.T) {
...
+					resource.TestCheckTypeSetElemAttr("digitalocean_kubernetes_cluster.foobar", "tags.*", "one"),
+					resource.TestCheckTypeSetElemAttr("digitalocean_kubernetes_cluster.foobar", "tags.*", "two"),
+					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "node_pool.0.labels.%", "0"),
+					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "surge_upgrade", "true"),
+					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "maintenance_policy.day", "any"),
+					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "maintenance_policy.start_time", "00:00"),
+					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "maintenance_policy.duration", "04:00h"),
+				),
+			},
+		},
```
andrewsomething, line 215: You create a `testAccDigitalOceanKubernetesConfigMaintenancePolicy`, but you are not actually using it in the test. We can probably simplify this to something like:

```suggestion
		Steps: []resource.TestStep{
			{
				Config: testAccDigitalOceanKubernetesConfigMaintenancePolicy(testClusterVersion19, rName),
				Check: resource.ComposeAggregateTestCheckFunc(
					testAccCheckDigitalOceanKubernetesClusterExists("digitalocean_kubernetes_cluster.foobar", &k8s),
					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "name", rName),
					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "maintenance_policy.0.day", "monday"),
					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "maintenance_policy.0.start_time", "00:00"),
					resource.TestCheckResourceAttrSet("digitalocean_kubernetes_cluster.foobar", "maintenance_policy.0.duration"),
				),
			},
		},
```

digitalocean/resource_digitalocean_kubernetes_cluster_test.go
```diff
@@ -294,6 +335,9 @@ func TestAccDigitalOceanKubernetesCluster_CreatePoolWithAutoScale(t *testing.T)
 					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "node_pool.0.min_nodes", "1"),
 					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "node_pool.0.max_nodes", "3"),
 					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "auto_upgrade", "true"),
+					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "maintenance_policy.day", "sunday"),
+					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "maintenance_policy.duration", "04:00"),
+					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "maintenance_policy.start_time", "05:00"),
```
andrewsomething, line 340: Since `maintenance_policy` is technically a list, you need to use:

```suggestion
					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "maintenance_policy.0.day", "sunday"),
					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "maintenance_policy.0.duration", "04:00"),
					resource.TestCheckResourceAttr("digitalocean_kubernetes_cluster.foobar", "maintenance_policy.0.start_time", "05:00"),
```

andrewsomething (changes requested):
Thanks for working on this! I've left a few comments and suggestions in line.

In addition to those, you'll also need to make a change to the read function to set the policy returned by the API to state:

https://github.com/digitalocean/terraform-provider-digitalocean/blob/24279405c3aca60b167d8531e8332cfa746a0b6b/digitalocean/resource_digitalocean_kubernetes_cluster.go#L290

You'll need to create a "flatten" function to do so. (More detail about what that means in one of the inline comments.) Something like this:

```go
func flattenMaintPolicyOpts(opts *godo.KubernetesMaintenancePolicy) []map[string]interface{} {
	result := make([]map[string]interface{}, 0)
	item := make(map[string]interface{})

	item["day"] = opts.Day.String()
	item["duration"] = opts.Duration
	item["start_time"] = opts.StartTime
	result = append(result, item)

	return result
}
```

atombrella (commented):
@andrewsomething Thanks a lot for the input. I'll work a bit more with it, and then ask for a second review, so it can be merged.

digitalocean/kubernetes.go
```diff
@@ -1,10 +1,11 @@
 package digitalocean
 
 import (
+	"strings"
```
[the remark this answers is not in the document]
andrewsomething: It's a good change. Looks like Atom has a Go "sort imports" plugin that runs on save. It's standard practice to put imports from the stdlib together and then external ones after a break.

andrewsomething:
@atombrella  Sounds good! Let us know if you run into any problems.

atombrella:
@andrewsomething I opened an issue with a suggestion how to test the provider locally (and have used this trick :+1: ). I did a `terraform plan` on the Kubernetes example in the repository, with `maintenance_policy` and got back this:

```
      + maintenance_policy {
          + day        = "any"
          + duration   = (known after apply)
          + start_time = "05:00"
        }
```

So seems to work :)

On a related note, I get a warning that the empty provider block in the `doks-cluster` folder is not needed.

digitalocean/resource_digitalocean_kubernetes_cluster_test.go
```diff
@@ -281,6 +305,11 @@ func TestAccDigitalOceanKubernetesCluster_CreatePoolWithAutoScale(t *testing.T)
 							min_nodes = 1
 							max_nodes = 3
 						}
+						maintenance_policy {
+							start_time = "05:00"
+							duration = "04:00"
```
atombrella, line 310: Is a computed attribute (although the API documentation mentions that it can be said manually), so should be removed.

atombrella: @andrewsomething Except for this small thing, I think this PR is ready to be merged :partying_face:

andrewsomething (approved):
:+1:  Looks great!

I did add an additional commit with some more testing and one addressing something I missed in my initial review. Specifically, `maintenance_policy` needs to be both `Optional` and `Computed`. As it does not have a default value, it needs to be "computed" from the API if it is not explicitly set by the user.

Thanks again for working on this!

atombrella:
@andrewsomething Thanks :) I'm learning something, so that's a nice reward.

PR #632, opened 2021-05-03
Use hashicorp/ghaction-import-gpg in release workflow.

andrewsomething:
HashiCorp forked the action we are using to import our GPG key to sign releases. This moves to it so that it is controlled by a trusted organization, not an individual.

Context: https://github.com/hashicorp/terraform-provider-scaffolding/issues/22


scotchneat (approved):
👍🏽

PR #633, opened 2021-05-04
Update issue templates

scotchneat:
Updates the issue templates to the new GitHub format. Includes instructions to contact support for 500 level API responses

PR #635, opened 2021-05-14
move config.yaml to proper directory

scotchneat:
I noticed the new issue templates (from #633) didn't display the custom choice options I added to the `config.yml`.

It turns out I created the file in the wrong directory. This fixes that.


andrewsomething (approved):
:+1:  Good catch

Issue #636, opened 2021-05-19
Error finding provider registry.terraform.io/hashicorp/digitalocean


ElijahGartin:
# Bug Report

Hello,

I ran into an issue while working on migrating a terraform code base from GCP to DO.  Similar issues have seemingly to have popped up a few times from what I found (in the References area at the bottom), albeit my issue is slightly different considering I already had the "solution code" in my provider information and it is still throwing an error.

---

## Describe the bug
When I attempt to build a `digitalocean_vpc`, it errors out being unable to find the provider for this resource.  It was my first module other than connecting via my API token. Hoping I don't run into similar issues with firewalls and droplets.

### Affected Resource(s)
- digitalocean_vpc

### Expected Behavior
Terraform should have initialized.

### Actual Behavior
Terraform Errored out claiming it couldn't find the provider for the vpc resource.

```
> terraform -v       
Terraform v0.15.3
on linux_amd64
```
```
> terraform init

Initializing modules...
- network in modules/network

Initializing the backend...

Initializing provider plugins...
- Finding latest version of hashicorp/digitalocean...
- Finding digitalocean/digitalocean versions matching ">= 2.8.0"...
- Installing digitalocean/digitalocean v2.8.0...
- Installed digitalocean/digitalocean v2.8.0 (signed by a HashiCorp partner, key ID F82037E524B9C0E8)

Partner and community providers are signed by their developers.
If you'd like to know more about provider signing, you can read about it here:
https://www.terraform.io/docs/cli/plugins/signing.html
╷
│ Error: Failed to query available provider packages
│ 
│ Could not retrieve the list of available versions for provider hashicorp/digitalocean:
│ provider registry registry.terraform.io does not have a provider named
│ registry.terraform.io/hashicorp/digitalocean
│ 
│ Did you intend to use digitalocean/digitalocean? If so, you must specify that source
│ address in each module which requires that provider. To see which modules are currently
│ depending on hashicorp/digitalocean, run the following command:
│     terraform providers
╵
```
```
terraform providers

Providers required by configuration:
.
├── provider[registry.terraform.io/digitalocean/digitalocean] >= 2.8.0
└── module.network
    └── provider[registry.terraform.io/hashicorp/digitalocean]
```

### Steps to Reproduce
1. `terraform init`

**Terraform Configuration Files**

### ROOT: Provider.tf
```
terraform {
  required_providers {
    digitalocean = {
      source  = "digitalocean/digitalocean"
[...]

PR #643, opened 2021-05-28
Cover required_providers in the main docs.

andrewsomething:
The example on the index page of our docs does not show the needed `required_providers` block.

This also adds a note for module developers about the  need to include the `required_providers` block in children modules. Rather than go into detail on module development here, I link out to the main Terraform module development docs as the cover it in much more depth than we can or should on this page.

Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/636

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -13,6 +13,15 @@ Use the navigation to the left to read about the available resources.
 ## Example Usage
 
 ```hcl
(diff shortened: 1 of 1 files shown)


scotchneat (approved):
Looks good

PR #656, opened 2021-07-02
Simplified some loops based on go-staticcheck S1011

atombrella:
More details about the pattern are given in https://staticcheck.io/docs/checks#S1011

I found the matches with `staticcheck . | grep S1011` inside the digitalocean directory.

**Edit**: Same with S1031.


atombrella:
@scotchneat @andrewsomething Maybe you can also take a look at this. It's less "controversial" than #657

scotchneat (approved):
Looks great. Thanks for the clean-up!

PR #658, opened 2021-07-06
Allow graceful shutdown of droplet on destroy/replace

gouyelliot:
Currently, the provider directly delete the droplet on a destroy or replace operation, which might result in a loss of data in some our apps.

I've added a simple param allowing to gracefully shutting down the droplet before deleting it.

### DIFF
--- a/digitalocean/resource_digitalocean_droplet.go
+++ b/digitalocean/resource_digitalocean_droplet.go
@@ -63,6 +63,12 @@ func resourceDigitalOceanDroplet() *schema.Resource {
 				ValidateFunc: validation.NoZeroValues,
 			},
 
+			"graceful_shutdown": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
+
 			"created_at": {
 				Type:     schema.TypeString,
 				Computed: true,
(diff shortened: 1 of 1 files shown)


scotchneat (changes requested):
@gouyelliot thanks for the contribution. This is great.
If you wouldn't mind adding a test to cover this scenario (digitalocean/resource_digitalocean_droplet_test.go), it would make this change good to go.

gouyelliot:
So I wasn't sure how to test that the droplet is actually shutdown before being destroyed, as both operations are done in the same function.

Please let me know if the test I added is relevant enough for you

gouyelliot:
FYI, we released a fork of this provider with the change and we are currently using it in production without issues: https://registry.terraform.io/providers/onfocusio/digitalocean/latest

atombrella:
@danaelhe @andrewsomething

scotchneat:
@gouyelliot / @atombrella, apologies for the delay in response. 
The change looks good but I'm seeing some failures in other tests that I'm looking at now. If there's no relevance, I'll go ahead and approve this change and deal with that separately.

scotchneat:
@gouyelliot It looks like the issue is related to the change after all. Since the new `graceful_shutdown` attribute is added to the droplet, `TestAccDigitalOceanDroplet_importBasic` is failing with the following error:

```
    import_digitalocean_droplet_test.go:19: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=1) {
         (string) (len=17) "graceful_shutdown": (string) (len=5) "false"
        }
```

Since we don't save state for `graceful_shutdown`, it can be added to `ImportStateVerifyIgnore` here: 
https://github.com/digitalocean/terraform-provider-digitalocean/blob/main/digitalocean/import_digitalocean_droplet_test.go#L33

A not as to why this issue doesn't appear in the PR checks is that we don't currently have the full set of acceptance tests run for every PR/commit. This is due to how acceptance tests actually modify resources on a live account. We've discussed options for adding them to ensure we cover each PR. Until then, we run the tests adhoc/locally to identify issues like this.

scotchneat:
@gouyelliot I tried to push a commit with the merge conflicts resolved as well as the change I previously suggested to this PR but it looks like I don't have [permission to edit the PR](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests).

Issue #659, opened 2021-07-06
digitalocean_droplet.userdata hash validation leading to conflict with resources created on the fly


voigt:
### Is your feature request related to a problem? Please describe.

I have the following Terraform manifest. Note that the `local-exec` provisioner creates the `controlplane.yaml` based on the loadbalancers IP, which is later consumed by `digitalocean_droplet.userdata`.

This leads to the fact, that the file hash at planning time is a different than at apply time.

```
resource "digitalocean_loadbalancer" "public" {
  name   = "talos-digital-ocean-tutorial-lb"
  region = "fra1"

  forwarding_rule {
    entry_port     = 443
    entry_protocol = "tcp"

    target_port     = 6443
    target_protocol = "tcp"
  }

  healthcheck {
    port                     = 6443
    protocol                 = "tcp"
    check_interval_seconds   = 10
    response_timeout_seconds = 5
    healthy_threshold        = 5
    unhealthy_threshold      = 3
  }

  droplet_tag = "control-plane"

  provisioner "local-exec" {
    command = "talosctl gen config talos-reply-dev https://${self.ip}:443"    <-- creates controlplane.yaml
  }

}

resource "digitalocean_droplet" "master" {
  count = 3

  name   = "talos-control-plane-${count.index+1}"
  size   = "s-2vcpu-4gb"
  image  = var.talos_image_id
  region = "fra1"
  private_networking = true
  ssh_keys = ["30756593"]
  user_data = "${file("controlplane.yaml")}"                <-- consumes controlplane.yaml

  tags = ["control-plane", "master", "reply", "test", "talos"]

  // must only be created after the talos configs have been generated
  // in digitalocean_loadbalancer.public local-exec provisioner
  depends_on = [
    digitalocean_loadbalancer.public,
  ]
}

output "loadbalancer_ip" {
    value = digitalocean_loadbalancer.public.ip
}

```

The resulting error looks like this:

```
[...]
│ Error: Provider produced inconsistent final plan
│ 
│ When expanding the plan for digitalocean_droplet.master[2] to
│ include new values learned so far during apply, provider
│ "registry.terraform.io/digitalocean/digitalocean" produced an
│ invalid new value for .user_data: was
│ cty.StringVal("a02e5d67fcb9b37bccfc1bb032304e35be70ae71"), but
│ now cty.StringVal("cbb80b0212f7146f87743a7bc1d925f4a295a4e4").
│ 
[...]

PR #663, opened 2021-07-27
add option to override user_data

voigt:
This is a proposal to solve #659: an issue that resulted in `Error: Provider produced inconsistent final plan` if a droplets `user_data` is altered between plan and apply phase of terraform.

This PR introduces an additional boolean field `force_override_userdata` which enforces the new user_data value.

More description can be found in #659.

If there is a decision that the proposed solution is good to go, I will add additional tests and docs.

### DIFF
--- a/digitalocean/resource_digitalocean_droplet.go
+++ b/digitalocean/resource_digitalocean_droplet.go
@@ -157,6 +157,13 @@ func resourceDigitalOceanDroplet() *schema.Resource {
 				},
 			},
 
+			"force_override_userdata": {
+				Type:     schema.TypeBool,
+				Optional: true,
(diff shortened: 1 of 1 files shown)


andrewsomething:
I totally understand the use case and why the current user experience is poor here, but I'm not 100% sure on this change.

When set, it would it completely ignore any changes to the provided `user_data` including in additional applies. That does seem to work for what you are doing currently. Though Terraform errors in this case because it finds the change to an attribute in-flight surprising. There is no guarantee that changes to how Terraform works would not break the assumptions here (i.e use the contents from the initial plan rather than post resource creation).

ChiefMateStarbuck:
closing due to staleness

Issue #670, opened 2021-08-11
Assigning database cluster resource's port to an SRV record's port does not seem to work.


aphistic:
# Bug Report

---

## Describe the bug
I'm trying to create a `digitalocean_record` SRV record that references one of my newly provisioned database clusters (done in the same plan/apply) by pointing the `digitalocean_database_cluster`'s `port` output to the `digitalocean_record`'s `port` field but when I try to do a `terraform plan` it tells me "'port' is required for when type is 'SRV'".

### Affected Resource(s)

- digitalocean_database_cluster
- digitalocean_record

### Expected Behavior

I'd expect the SRV record to have the database cluster's port value after the cluster has been provisioned.

### Actual Behavior

Terraform gives me the following error:

```
│ Error: `port` is required for when type is `SRV`
│ 
│   with digitalocean_record.pgsql_default_pub_srv,
│   on pgsql.tf line 21, in resource "digitalocean_record" "pgsql_default_pub_srv":
│   21: resource "digitalocean_record" "pgsql_default_pub_srv" {
```

### Steps to Reproduce

1. Create a `digitalocean_database_cluster` resource.
2. Create a `digitalocean_record` resource with type `SRV` and the port referencing the database cluster's port.
3. Run `terraform plan`

**Terraform Configuration Files**

```terraform
data "terraform_remote_state" "core" {
    backend = "remote"

    config = {
        organization = var.tf_core_org
        workspaces = {
            name = var.tf_core_workspace
        }
    }
}

resource "digitalocean_database_cluster" "pgsql_default" {
    name = "${var.do_project}-pgsql"
    engine = "pg"
    version = "11"
    size = "db-s-1vcpu-1gb"
    region = var.do_region
    node_count = 1

    private_network_uuid = data.terraform_remote_state.core.outputs.vpc.id

    tags = ["default"]
}

resource "digitalocean_record" "pgsql_default_pub" {
    domain = data.terraform_remote_state.core.outputs.domain.name

    type = "A"
    name = "pub.default.pg.${data.terraform_remote_state.core.outputs.domain.env}"
    value = digitalocean_database_cluster.pgsql_default.host
}
resource "digitalocean_record" "pgsql_default_pub_srv" {
    domain = data.terraform_remote_state.core.outputs.domain.name

    type = "SRV"
    name = "_postgresql_.tcp.${digitalocean_record.pgsql_default_pub.fqdn}"
    
    priority = 0
    weight = 0

[...]

PR #671, opened 2021-08-16
record: Move validation by type out of CustomizeDiff.

andrewsomething:
Fixes: #670 

We're doing validation inside of a `CustomizeDiff` function. The interpolated values are not available there on plan. This prevents user from doing things like referencing a database cluster's port in a SRV record. This is similar to hashicorp/terraform-provider-consul#260 This PR moves that validation to later in the process.

This has one side effect. The reason for using a `CustomizeDiff` for something like this is that it allows you to error out early before any of the plan runs. Now the error will not happen until the record is being created. I don't love that, but not sure I see a better way to do it.

### DIFF
--- a/digitalocean/resource_digitalocean_record.go
+++ b/digitalocean/resource_digitalocean_record.go
@@ -125,14 +125,10 @@ func resourceDigitalOceanRecord() *schema.Resource {
 				}
 			}
 
-			_, hasPort := diff.GetOkExists("port")
 			_, hasWeight := diff.GetOkExists("weight")
 			if recordType == "SRV" {
 				if !hasPriority {
 					return fmt.Errorf("`priority` is required for when type is `SRV`")
-				}
-				if !hasPort {
-					return fmt.Errorf("`port` is required for when type is `SRV`")
 				}
 				if !hasWeight {
 					return fmt.Errorf("`weight` is required for when type is `SRV`")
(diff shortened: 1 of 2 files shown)


bentranter (dismissed):
LGTM!

scotchneat:
@andrewsomething does this happen when running a plan for the first time since the interpolated values won't be known until after the apply is run? Or does it happen on every plan?
Is there a case where the referenced value would be known when running plan?

andrewsomething:
> @andrewsomething does this happen when running a plan for the first time since the interpolated values won't be known until after the apply is run? Or does it happen on every plan?
> Is there a case where the referenced value would be known when running plan?

@scotchneat  It only happens if the resource being referenced is being created at the same time. For example, if you already have the database, you can add a domain record pointing to its port.

Thinking about this one some more... `CustomizeDiff` functions are not really meant for this kind of validation, but the fact that they run so early makes them super useful for that. It's much nicer to get the validation error in the plan and before other resources get created. Doing the validation in the create function means you can error out mid-apply. 

Maybe just moving the validation for `port` make sense? It seems to be the one that is most likely to come from interpolation (and the only one we've had a bug reported for). This would allow us to retain the current behavior for other attributes and still unblock the use case in #670.

andrewsomething:
@bentranter, @scotchneat - I decided to take the more limited approach  describe in my last comment. Need a new review due to the changes.

scotchneat (approved):
I haven't tested this manually, but the tests you included cover it well. LGTM

Issue #479, opened 2020-08-15
Add droplet monitoring resource


epicserve:
If possible, please add a resource for adding monitoring alerts.

atombrella:
It doesn't seem like there are any API endpoints for this, which seems like a prerequisite. That was at least the impression I got from https://github.com/digitalocean/godo which is used in different parts of the code.

atombrella:
I have an ongoing effort to implement this :tada: I hope to open a PR in September that's ready for review :100:

atombrella:
@andrewsomething Can be closed :)

@epicserve It's available in 2.12 (https://github.com/digitalocean/terraform-provider-digitalocean/commit/cf4091dc31668ce817ab3a9870ee2d43f1cca627)

PR #679, opened 2021-09-01
Add resource for monitor alerts

atombrella:
It still needs a lot of work, but at least it's a start :) Hoping for a bit of help to get it pushed over the finish line. I'll start also on unit/acceptance tests.

- [x] Have Terraform create the alerts
- [x] Documentation. There is an example, attribute reference. 
- [x] import acceptance test
- [x] More acceptance and uni test
 
#479


digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,291 @@
...
+		return diag.Errorf("Error creating Alert Policy: %s", err)
+	}
+
+	d.SetId(alertPolicy.UUID)
+	log.Printf("[INFO] Alert Policy created, ID: %s", d.Id())
+
+	return resourceDigitalOceanMonitorAlertRead(ctx, d, meta)
+}
+
+func expandAlerts(d *schema.ResourceData) *godo.Alerts {
```
atombrella, line 172: @andrewsomething Can you assist a bit with this function, and the `flattenAlerts`? Thanks. It's one small hurdle that I can't really figure out. I think the other expand/flatten functions are almost fine.

atombrella (commented):
@andrewsomething @danaelhe My first goal is to be able to create a monitor alert with the compiled module, and then work more on the acceptance/unit testing.

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,291 @@
...
+			"description": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Description of the alert policy",
+			},
+
+			"enabled": {
+				Type:     schema.TypeBool,
+				Default:  true,
+				Required: true,
```
danaelhe, line 74: Running the example (good example btw) and it looks like "Default cannot be set with Required", I'm thinking we drop the "default"

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,291 @@
...
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"uuid": {
+				Type:     schema.TypeString,
+				Computed: true,
+				Required: true,
```
danaelhe, line 32: Running the example (good example btw) and it looks like "uuid: Cannot be both Required and Computed". I think we can keep it to just "Computed"

andrewsomething: Right, `uuid` should just be `Computed` and not `Required` as it is not user configurable and it always set from the API response.

digitalocean/resource_digitalocean_monitor_alert.go
andrewsomething: Right, `Default` and `Required: true` can not be set at the same time:

https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#default

`Required: true` enforces that the user always has it specified in their configuration file while providing a  `Default` implied that you do not need to explicitly set it in the configuration file as the default is used if not provided.

:+1:  I like using a default here as most users will want their alerts enabled by default.

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,291 @@
...
+			"alerts": {
+				Type:        schema.TypeList,
+				Computed:    false,
+				Required:    false,
+				Description: "List with details how to notify about the alert. Support for Slack or email.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"slack": {
+							Type:     schema.TypeList,
+							Required: false,
```
andrewsomething, line 94:
```suggestion
							Optional: true,
```

`Optional: true` is needed over `Required: false`

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,291 @@
...
+				Required:     true,
+				ValidateFunc: validation.FloatAtLeast(0),
+			},
+
+			"tags": tagsSchema(),
+
+			"alerts": {
+				Type:        schema.TypeList,
+				Computed:    false,
+				Required:    false,
```
andrewsomething, line 88:
```suggestion
				Optional:    true,
```

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,291 @@
...
+					Schema: map[string]*schema.Schema{
+						"slack": {
+							Type:     schema.TypeList,
+							Required: false,
+							Computed: false,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"channel": {
+										Type:         schema.TypeList,
+										Computed:     false,
```
andrewsomething, line 100:
```suggestion
										Type:         schema.TypeString,
```

This is just a string:

https://github.com/digitalocean/godo/blob/78ca5f7b651de84bf1489c294a5ff59950c7158a/monitoring.go#L69

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,291 @@
...
+										Required:     true,
+										Description:  "The Slack channel to send alerts to",
+										ValidateFunc: validation.StringIsNotEmpty,
+									},
+									"url": {
+										Type:             schema.TypeString,
+										Computed:         false,
```
andrewsomething, line 107:
```suggestion
										Required:         true,
```

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,291 @@
...
+										ValidateFunc:     validation.StringIsNotEmpty,
+									},
+								},
+							},
+						},
+						"email": {
+							Type:         schema.TypeList,
+							Computed:     false,
+							Required:     false,
+							ValidateFunc: validation.ListOfUniqueStrings,
```
andrewsomething, line 119:
```suggestion
							Type:     schema.TypeList,
							Computed: false,
							Optional: true,
							Elem: &schema.Schema{
								Type: schema.TypeString,
							},
```

`TypeList` requires an `Elem` defining what the items in the list should be.

https://www.terraform.io/docs/extend/schemas/schema-types.html#typelist

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,291 @@
...
+						},
+					},
+				},
+			},
+
+			"entities": {
+				Type:        schema.TypeSet,
+				Required:    true,
+				MinItems:    1,
+				Description: "The droplets to apply the alert policy to",
```
andrewsomething, line 129:
```suggestion
				Description: "The droplets to apply the alert policy to",
				Elem: &schema.Schema{
					Type: schema.TypeString,
				},
```

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,291 @@
...
+			},
+
+			"window": {
+				Type:     schema.TypeString,
+				Required: true,
+				ValidateFunc: validation.StringInSlice([]string{
+					"5m", "10m", "30m", "1h",
+				}, false),
+				MinItems: 1,
```
andrewsomething, line 138:
```suggestion
```

`MinItems` is for use with `TypeList` or `TypeSet`

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,291 @@
...
+			},
+		},
+	}
+}
+
+func resourceDigitalOceanMonitorAlertCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*CombinedConfig).godoClient()
+
+	alertCreateRequest := &godo.AlertPolicyCreateRequest{
+		Type:        d.Get("type").(string),
```
andrewsomething, line 148:
```suggestion
		Type:        d.Get("type").(string),
		Enabled:     godo.Bool(d.Get("enabled").(bool)),
```

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,291 @@
...
+		Description: d.Get("description").(string),
+		Tags:        expandTags(d.Get("tags").(*schema.Set).List()),
+		Compare:     d.Get("compare").(godo.AlertPolicyComp),
```
andrewsomething, line 151:
```suggestion
		Compare:     godo.AlertPolicyComp(d.Get("compare").(string)),
```

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,291 @@
...
+		Window:      d.Get("window").(string),
+		Value:       d.Get("value").(float32),
```
andrewsomething, line 153:
```suggestion
		Value:       float32(d.Get("value").(float64)),
```

The native Terraform float is a `float64`, so it needs to get converted to `float32`

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,291 @@
...
+
+	return resourceDigitalOceanMonitorAlertRead(ctx, d, meta)
+}
+
+func expandAlerts(d *schema.ResourceData) *godo.Alerts {
+
+	alerts := &godo.Alerts{
+		Slack: expandSlack(d.Get("slack").([]interface{})),
+		Email: expandEmail(d.Get("email").([]interface{})),
+	}
```
andrewsomething, line 177:
```suggestion
func expandAlerts(config []interface{}) godo.Alerts {
	alertConfig := config[0].(map[string]interface{})
	alerts := godo.Alerts{
		Slack: expandSlack(alertConfig["slack"].([]interface{})),
		Email: expandEmail(alertConfig["email"].([]interface{})),
```

Unfortunately, you can't just `d.Get("slack")` for these nested attributes.

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,291 @@
...
+
+	return alerts
+}
+
+func expandSlack(slackChannels []interface{}) []godo.SlackDetails {
+	expandedSlackChannels := make([]godo.SlackDetails, 0, len(slackChannels))
+
+	for _, slackChannel := range slackChannels {
+		slack := slackChannel.(map[string]interface{})
+		n := &godo.SlackDetails{
```
andrewsomething, line 187:
```suggestion
		n := godo.SlackDetails{
```

andrewsomething (changes requested):
This is looking great! Offered some pointers inline including  for the `expandAlerts`.

digitalocean/resource_digitalocean_monitor_alert.go
atombrella: I've been looking at https://docs.digitalocean.com/reference/api/api-reference/#operation/list_alert_policies to see what's required, and what's not. Thanks for the adjustment.

digitalocean/resource_digitalocean_monitor_alert.go
atombrella: Thanks :)

atombrella:
I enabled the new resource in provider.go, and you can now run the acceptance tests (I'll add more). This one fails. I doubt I'll have much time to work on this during next week.

```
make testacc TESTARGS='-run=TestAccDigitalOceanMonitorAlertNoAlerts'
```

atombrella:
It actually creates an alert for a droplet, with one of the acceptance tests. There is a stack trace that's somewhat tricky for me to debug. Help would be appreciated.

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	d.SetId(alert.UUID)
+	d.Set("description", alert.Description)
+	d.Set("enabled", alert.Enabled)
+	d.Set("compare", alert.Compare)
+	d.Set("alerts", flattenAlerts(alert.Alerts))
+	d.Set("value", alert.Value)
+	d.Set("window", alert.Window)
+	d.Set("entities", flattenEntities(alert.Entities))
```
scotchneat, line 306: This can be simplified since `alert.Entities` is already `[]string`.

```suggestion
	d.Set("entities", alert.Entities)
```

digitalocean/resource_digitalocean_monitor_alert.go
```diff
@@ -0,0 +1,323 @@
...
+	if len(entities) == 0 {
+		return nil
+	}
+
+	return []interface{}{
+		map[string]interface{}{
+			"entities": entities,
+		},
+	}
+}
```
scotchneat, line 269: Since [Line#308](https://github.com/digitalocean/terraform-provider-digitalocean/pull/679/files?file-filters%5B%5D=.go#diff-4f9285880d8045a1340ae2e5142505485dd0ed8a9dfe85d930a8ce90c2b4529bR306) can be changed, this funciton can be removed.
```suggestion
```

docs/resources/monitor_alert.md
```diff
@@ -0,0 +1,80 @@
...
+  entities    = [digitalocean_droplet.web.id]
+  description = "Alert about CPU usage"
+}
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+* `alerts` - (Required) How to send notifications about the alerts.
```
scotchneat, line 43: If we stick with alerts being a list, then we should indicate here that only one alerts element is allowed.

scotchneat (changes requested):
@atombrella When I ran the tests, I received the following panic:
```panic: alerts.0.email: '': source data must be an array or slice, got struct```

It looks like your last commit tries to address what I initially thought was the problem. Unfortunately, I still receive the above error after fixing that. 

The problem is with how the `flatten` functions handle the email and slack data. I've added some suggestions inline to address this and the test cases.

I also added a comment regarding the `alerts` property set as a `schema.TypeList` for some discussion.

File: docs/resources/monitor_alert.md (hunk 18 above)
Comment (atombrella) [reply docs/resources/monitor_alert.md:43]: Yes. I've configured some alert policies inside the web panel. I only used email to get notifications, which is a single list (fine in this PR), but I'm not sure about multiple slack channels. So yes, we can clarify the exact phrasing here.

**Comment by atombrella** [comment]
@scotchneat @danaelhe @andrewsomething Thanks for the assistance. I do have one testing regarding tests: I would have liked to include some tests for validation, e.g., window not in [5m, 10m, 30m, 1h]. However, it doesn't seem like the provider has anything else but acceptance tests (which all require network connections, and a token), so I suppose this is not essential. I think it's approaching a stable state :)

**Comment by atombrella** [comment]
My branch is open for edits/commits from maintainers.

File: digitalocean/import_digitalocean_monitor_alert_test.go [hunk 19]
```diff
@@ -0,0 +1,56 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(testAccAlertPolicy, randName, "", "10m", "v1/insights/droplet/memory_utilization_percent", "Alert about memory usage"),
+			},
+			{
+				ResourceName:      resourceName,
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateIdFunc: testAccMonitorAlertImportID(resourceName),
+			},
```
atombrella, line 31: This part of the test fails.

andrewsomething: Removing `ImportStateIdFunc` was enough to get this working! I pushed a commit with that change. That function is only needed when the resource's ID is not enough to pass to the read function. For example, with a database user you need both the ID of the parent cluster and the username itself. In this case we can just pass in the alert's ID directly.

andrewsomething:
@atombrella  In addition to fixing the import test (f69c6d6), I pushed a few additional commits:

- Support alerts using only tags (48ca9ce).
- Update godo to pick up new consts (9eb4610) and use them for all alert types (67f3812).
- Support updating all attributes (96cc08c).

docs/resources/droplet.md
```diff
@@ -31,7 +31,8 @@ The following arguments are supported:
 * `backups` - (Optional) Boolean controlling if backups are made. Defaults to
    false.
 * `monitoring` - (Optional) Boolean controlling whether monitoring agent is installed.
-   Defaults to false.
+   Defaults to false. If set to `true`, you can configure monitor alert policies
+   [monitor alert resource](/providers/digitalocean/digitalocean/latest/docs/resources/monitor_alert)
```
andrewsomething, line 35: :+1:  Good call out!

andrewsomething (approved):
I think we're probably good to go here, but wouldn't mind a second look :eyes:  @scotchneat 

Thanks for another great contribution @atombrella!

atombrella:
> @atombrella In addition to fixing the import test ([f69c6d6](https://github.com/digitalocean/terraform-provider-digitalocean/commit/f69c6d6ed58344b76b65e1a9807ea433c13d14e6)), I pushed a few additional commits:
> 
> * Support alerts using only tags ([48ca9ce](https://github.com/digitalocean/terraform-provider-digitalocean/commit/48ca9ce6b2f4530846b37852057a0573b0a26a99)).
> * Update godo to pick up new consts ([9eb4610](https://github.com/digitalocean/terraform-provider-digitalocean/commit/9eb4610f94e3b68f37de795f014164a1d0815a54)) and use them for all alert types ([67f3812](https://github.com/digitalocean/terraform-provider-digitalocean/commit/67f38120d83811e4c0a9e0c8b9c2e27a8f45e55a)).

I'm glad you saw the comment in the code about those constants :)

> * Support updating all attributes ([96cc08c](https://github.com/digitalocean/terraform-provider-digitalocean/commit/96cc08c517721c1bc497155e363cbe24fae1f94f)).

Great. I appreciate the help and coaching. A colleague's first reaction to me highlighting the alert monitor feature for droplets was "Can I do it in Terraform?". Now they will be.

digitalocean/resource_digitalocean_monitor_alert_test.go
```diff
@@ -0,0 +1,284 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "type", "v1/insights/droplet/cpu"),
+					resource.TestCheckResourceAttr(resourceName, "compare", "GreaterThan"),
+					resource.TestCheckResourceAttr(resourceName, "alerts.#", "1"),
+					resource.TestCheckResourceAttr(resourceName, "alerts.0.email.0", "benny@digitalocean.com"),
+				),
+			},
+		},
+	})
+}
```
scotchneat, line 170: These tests look identical. Is one supposed to test something different than the other?

scotchneat (approved):
:LGTM: 
Thanks for the contribution @atombrella, and the collaboration!

I just had one question about a couple of tests but it's not a blocker.

digitalocean/resource_digitalocean_monitor_alert_test.go
atombrella: Looks like something that should be deleted. Please do so. Thanks for noticing! @andrewsomething helped add and fix tests so I think it's good.

atombrella: You can create an alert without Slack and Email (which was done with Support alerts using only tags (48ca9ce).). I don't really know what I had in mind with the duplicate test here.

atombrella:
@scotchneat I deleted the duplicated test, and thought it'd be nice to test having two slack channels connected. Please see 65e9d77d1b79f203fb20de452896e3bd4cb36563

andrewsomething:
@atombrella  Good call.

andrewsomething (approved):
:shipit:

PR #709, opened 2021-09-24
Fix parentheses to make clickable links in CHANGELOG

atombrella:


andrewsomething (approved):
:facepalm:  Thanks for catching that!

PR #718, opened 2021-10-06
Add HA support to digitalocean_kubernetes_cluster

bentranter:
Adds support for the `ha` boolean attribute when creating Kubernetes clusters.

Also bumps the version of godo to the latest release.

---

I'm not sure if there's something I need to do to enforce that this behaviour is only for creates, and not for updates, so opening as a draft for now.


digitalocean/resource_digitalocean_kubernetes_cluster.go
```diff
@@ -52,6 +52,12 @@ func resourceDigitalOceanKubernetesCluster() *schema.Resource {
 				Default:  true,
 			},
 
+			"ha": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
```
andrewsomething, line 58: > I'm not sure if there's something I need to do to enforce that this behaviour is only for creates, and not for updates, so opening as a draft for now.

Generally for an option that can only be enabled/disable at create time, we want to set `ForceNew`. This will re-create the resource if the option is changed.

```suggestion
				Default:  false,
				ForceNew: true,
```

PR #734, opened 2021-10-27
Add scheduled acceptance test runs

scotchneat:
This workflow runs the acceptance tests every weekday at 14:00 UTC (10a Eastern)


andrewsomething (approved):
:+1: LGTM!

We should definitely be running `make sweep` at some point too.

Issue #726, opened 2021-10-23
Destroying digitalocean_container_registry_docker_credentials does not delete API token in DO


dmikalova:
# Bug Report

---

## Describe the bug

Creating an API token with the resource `digitalocean_container_registry_docker_credentials`, and then destroying that resource does not delete the API token.

### Affected Resource(s)

- digitalocean_container_registry_docker_credentials
### Expected Behavior

When destroying the resource the token should be deleted from DO

### Actual Behavior

The token remains in place.

### Steps to Reproduce

**Terraform Configuration Files**

```
 resource "digitalocean_container_registry_docker_credentials" "main" {
   registry_name = var.container_registry_info.name
 }
```
run `terraform apply; terraform destroy`, and then check your [API tokens](https://cloud.digitalocean.com/account/api/tokens?i=5437be).

### Version
```
Terraform v1.0.9
on linux_amd64
+ provider registry.terraform.io/digitalocean/digitalocean v2.14.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.6.1
```

PR #735, opened 2021-10-29
docker creds: Revoke token on delete.

andrewsomething:
As the concept being used here is "destroying" the credentials, we should really revoke the generated token. 

Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/726


bentranter (approved):
LGTM!

PR #737, opened 2021-11-01
Acceptance tests: Don't use deprecated size slugs.

andrewsomething:
These size slugs pointing to legacy Droplet plans have been deprecated and are not available on most accounts. This leads to test failures when run under the new acceptance testing account.


scotchneat (approved):
You beat me to it 👍🏽

PR #738, opened 2021-11-01
Limit acceptance test job/workflow to single run at a time

scotchneat:
This sets the [concurrency group](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency) for the acceptance test job and workflow to ensure only one runs at any given time.


andrewsomething (approved):
:+1:  Nice. TIL: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency

Issue #747, opened 2021-11-12
Cannot rename default vpc


dmikalova:
# Bug Report

---

## Describe the bug
I'm trying to use terraform to rename managed default VPCs from `$region` to `$region-default`. This works in the UI, but fails in Terraform.

### Affected Resource(s)
- digitalocean_vpc

### Expected Behavior
The vpc should be renamed.

### Actual Behavior
Terraform errors out:

```
│ Error: Error updating VPC : PUT https://api.digitalocean.com/v2/vpcs/5955a55c-421d-46a3-8147-7f5676752a7a: 403 (request "3d1d98ea-242b-47d7-a6a8-a27d27a6b5ac") Can not unset default VPCs
│ 
│   with module.vpcs.digitalocean_vpc.main["sfo3"],
│   on ../vpcs/main.tf line 1, in resource "digitalocean_vpc" "main":
│    1: resource "digitalocean_vpc" "main" {
```

### Steps to Reproduce

1. import a default vpc into terraform
2. rename the vpc with terraform

andrewsomething:
Hi @dmikalova Thanks for raising this issue. We've identified the problem and should have a fix in https://github.com/digitalocean/terraform-provider-digitalocean/pull/748

PR #748, opened 2021-11-12
vpc: Allow updating name and description of default vpcs.

andrewsomething:
The PUT request for updating a VPC needs to include the `default` attribute even if it is not changed in order to support updating  the name or description of a region's default vpc.

Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/747


scotchneat (approved):
LGTM

PR #753, opened 2021-11-30
Update domains sweeper

scotchneat:
I noticed domain resources aren't getting cleaned up after acceptance tests but it's difficult to identify which tests aren't properly cleaning up after themselves.

This change includes a few things to help diagnose as well clean up:

1. Update the domains resource sweeper to use match resources with the common `testNamePrefix` (a const value set in the [digitalocean/provider_test.go#L17](https://github.com/digitalocean/terraform-provider-digitalocean/blob/main/digitalocean/provider_test.go#L17))
2. Update the names used for domains in tests to call `randomTestName()` (which uses the above common prefix)
3. Extended `randomTestName()` to allow a list of strings that will get concatenated with `-` (and will replace spaces with `_`). 
   * **_I'm curious what thoughts there are on this_**. I made the function variadic so existing calls don't fail. However, I know some resources have a relatively short name length limit and this could cause issues if misused.
4. Lastly, I added a job to the acceptance test workflows to call `make sweep`. This will allow the resources to be cleaned up after every acceptance test run regardless of the tests handling clean up properly, and help reduce future errors due to resource limits being reached.


andrewsomething (approved):
:+1:  Looks great! 

We have a ticket in our internal backlog around standardizing resource names used in tests and adding sweepers for all resource types. It's something we haven't been consistent about.

PR #755, opened 2021-12-07
Sort imports with goimports. Introduce lint in GitHub workflows.

atombrella:
"Stolen" from https://github.com/hashicorp/terraform-provider-azurerm/blob/main/.golangci.yml and https://github.com/hashicorp/terraform-provider-azurerm/tree/main/scripts I introduced linting to check for import sorts in the code. I disabled a great deal of the linters, but IMO it's good to have linting in a code-base.


scripts/goimport-file.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
atombrella, line 1: Maybe this file shouldn't be committed. I just used it once to sort the imports.

atombrella:
@danaelhe @andrewsomething @scotchneat https://golangci-lint.run/usage/configuration/ contains documentation about the linting tool. Maybe you can use my PR as a start to configure what's useful and less useful to include.

atombrella:
@danaelhe @andrewsomething @scotchneat Should I close this, or can you have a look at it?

andrewsomething (approved):
Sorry for over looking this. We've let it sit for an embarrassingly long time. 

I did push a change making the usage more consistent with how we are using it in another project (using the offical GitHub action, enabling some additional check, and only disabling the specific staticchecks that currently fail).

This will be extremely helpful with improving the quality of this project moving forward. Thank you for all the great contributions! :pray:

PR #759, opened 2021-12-17
Makefile: Only run sweep against the digitalocean package.

andrewsomething:
The other packages don't import `github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource`, so that leads to failures when the sweeper is ran. E.g.

https://github.com/digitalocean/terraform-provider-digitalocean/runs/4565075328?check_suite_focus=true#step:4:205


scotchneat (approved):
👍🏽

Issue #769, opened 2022-01-04
Failed to update record when domain gets recreated


JonasTaulienSolutions:
# Bug Report
## Describe the bug
I have three resources configured in my terraform setup. A droplet, that hosts my website, a domain that points to the droplets ipv4-address and a A record "www" that also points to the droplets ipv4-address.
When the droplets gets recreated from Terraform, the domain gets recreated with it, but the A-record does not get deleted and Terraform tries to update its ipv4-address to the ipv4-address of the newly created droplet replacement.
This leads to the following error:
```
...
digitalocean_record.my_record: Modifying... [id=286624857]

│ Error: Failed to update record: PUT https://api.digitalocean.com/v2/domains/example.com/records/286624857: 404 (request "89408e60-6c9a-48db-bbba-b2c34a517470") The resource you were accessing could not be found.
│ 
│   with digitalocean_record.my_record,
│   on main.tf line 46, in resource "digitalocean_record" "my_record":
│   46: resource "digitalocean_record" "my_record" {
│ 
```

### Affected Resource(s)
- digitalocean_droplet
- digitalocean_domain
- digitalocean_record

### Expected Behavior
The record should also get recreated instead of Terraform trying to change it.

### Steps to Reproduce

**Terraform Configuration Files**
```tf
resource "digitalocean_droplet" "my_droplet" {
  image      = "debian-11-x64"
  name       = "my_droplet"
  region     = "fra1"
  size       = "s-1vcpu-1gb"
  backups    = true
  monitoring = true
  user_data  = file("cloud-config.yaml")
}

resource "digitalocean_domain" "example_com" {
  name = "example.com"
  ip_address = digitalocean_droplet.my_droplet.ipv4_address
}

resource "digitalocean_record" "my_record" {
  domain = digitalocean_domain.example_com.name
  type = "A"
  name = "www"
  value = digitalocean_droplet.my_droplet.ipv4_address
}
```

**Expected behavior**
Terraform version:
```
Terraform v1.1.2
on darwin_amd64
+ provider registry.terraform.io/digitalocean/digitalocean v2.16.0
+ provider registry.terraform.io/hashicorp/http v2.1.0

```

## Additional context
Result of `terraform plan` was:
```
Terraform will perform the following actions:

  # digitalocean_domain.example_com must be replaced
-/+ resource "digitalocean_domain" "example_com" {
      ~ id         = "example.com" -> (known after apply)
[...]

PR #770, opened 2022-01-04
docs: Update records examples to use domain id over name.

andrewsomething:
As discussed in https://github.com/digitalocean/terraform-provider-digitalocean/issues/769, using the domain's `id` instead of its `name` can be useful to ensure a record is recreated when a domain is.

Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/769


scotchneat (approved):
👍🏽

Issue #763, opened 2021-12-29
Terraform floating IP assignment import leads to crash


SamWhited:
# Bug Report

While trying to do an import of a digitalocean_floating_ip_assignment (that somehow didn't get updated in the state after the `terraform apply` that created the assignment, which might be another bug but I don't have any information or any indication of why this happened), Terraform crashed. The import was likely incorrect, but the docs don't mention what information is needed so I tried this.

```
$ terraform import module.prod.digitalocean_floating_ip_assignment.gitea "167.99.22.167"
module.prod.digitalocean_floating_ip_assignment.gitea: Importing from ID "167.99.22.167"...
module.prod.digitalocean_floating_ip_assignment.gitea: Import prepared!
  Prepared digitalocean_floating_ip_assignment for import
module.prod.digitalocean_floating_ip_assignment.gitea: Refreshing state... [id=167.99.22.167]
╷
│ Error: Request cancelled
│ 
│ The plugin.(*GRPCProvider).ReadResource request was cancelled.
╵


Stack trace from the terraform-provider-digitalocean_v2.16.0 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xe359ab]

goroutine 38 [running]:
github.com/digitalocean/terraform-provider-digitalocean/digitalocean.resourceDigitalOceanFloatingIpAssignmentRead(0x1247538, 0xc0004a68a0, 0xc0005e4b80, 0xf6ab40, 0xc00056bc80, 0xc000694280, 0xc000693908, 0x40e0f8)
        github.com/digitalocean/terraform-provider-digitalocean/digitalocean/resource_digitalocean_floating_ip_assignment.go:77 +0x2ab
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).read(0xc0002f0700, 0x12474c8, 0xc000053fc0, 0xc0005e4b80, 0xf6ab40, 0xc00056bc80, 0x0, 0x0, 0x0)
        github.com/hashicorp/terraform-plugin-sdk/v2@v2.7.0/helper/schema/resource.go:347 +0x17f
[...]

PR #771, opened 2022-01-04
floating_ip_assignment: Properly support importing existing assignments.

andrewsomething:
`floating_ip_assignment` resources currently use [`schema.ImportStatePassthrough`](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.10.1/helper/schema#ImportStatePassthroughContext) for importing. This is meant for use when an ID-only refresh is possible. It would have never worked for this resource which needs both the floating IP and the Droplet ID. 

Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/763

### DIFF
--- a/digitalocean/import_digitalocean_floating_ip_assignment_test.go
+++ b/digitalocean/import_digitalocean_floating_ip_assignment_test.go
@@ -1,0 +1,84 @@
+package digitalocean
+
+import (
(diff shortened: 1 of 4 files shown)


scotchneat (approved):
lgtm

Issue #772, opened 2022-01-07
create digitalocean_loadbalancer data source from load balancer id


andyli:
Right now the [`digitalocean_loadbalancer` data source](https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/data-sources/loadbalancer) use `name` to query load-balancers. I would like to use id to query instead.

The reason I want to do this is that I'm setting up [Accessing pods over a managed load-balancer from inside the cluster](https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/docs/controllers/services/examples/README.md#accessing-pods-over-a-managed-load-balancer-from-inside-the-cluster), in which I need to create a DNS record for the load balancer ip. The ingress-nginx k8s service records the load balancer id in the `kubernetes.digitalocean.com/load-balancer-id` annotation, but there is no record of the load balancer name.

opeco17:
I've created the PR (https://github.com/digitalocean/terraform-provider-digitalocean/pull/773) to create loadbalancer datasource by its ID.

By this update, we will be able to create loadbalancer datasource by its ID in addition to name like below.

```
data "digitalocean_loadbalancer" "example" {
  id = "loadbalancer_id"
}
```

PR #773, opened 2022-01-10
Fetch loadbalancer resource in datasource by ID

opeco17:
Issue: https://github.com/digitalocean/terraform-provider-digitalocean/issues/772

Currently we can fetch loadbalancer resource in datasource by only its name.
https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/data-sources/loadbalancer

By this update, we will be able to fetch loadbalancer resource by its ID in addition to name like below.

```
data "digitalocean_loadbalancer" "example" {
  id = "loadbalancer_id"
}
```

This update is referring to the following droplet datasource.
https://github.com/digitalocean/terraform-provider-digitalocean/blob/main/digitalocean/datasource_digitalocean_droplet.go#L40

### DIFF
--- a/digitalocean/datasource_digitalocean_loadbalancer.go
+++ b/digitalocean/datasource_digitalocean_loadbalancer.go
@@ -15,11 +15,19 @@ func dataSourceDigitalOceanLoadbalancer() *schema.Resource {
 		ReadContext: dataSourceDigitalOceanLoadbalancerRead,
 		Schema: map[string]*schema.Schema{
 
+			"id": {
+				Type:         schema.TypeString,
+				Optional:     true,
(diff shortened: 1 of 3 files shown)


andrewsomething (approved):
:+1:  Looks great. Thanks for the contribution!

I did add one additional commit to use unique tag names in the acceptance tests. This is useful if the tests are run in parallel so that Droplets from one test are not added to load balancers from another.

https://github.com/digitalocean/terraform-provider-digitalocean/pull/773/commits/05127eb1c007cea222010989fc5131546477d045

andyli:
Just tried it - works perfectly. Thanks!

PR #767, opened 2022-01-01
Setting droplet ID after it might be unset

iwankgb:
How-to-reproduce:

Execute following snippet:
```terraform
resource "digitalocean_droplet" "this" {
  name = "this"
  image = "debian-11-x64"
  region = "fra1"
  size = "s-1vcpu-1gb-amd"

  user_data = "🐼" # note content of user_data - it will make droplet creation fail
}
```
`terraform apply` will fail with unexpected error:
```
│ Error: Error waiting for droplet () to become ready: strconv.Atoi: parsing "": invalid syntax
│ 
│   with digitalocean_droplet.critical-today,
│   on critical-today-droplet.tf line 1, in resource "digitalocean_droplet" "critical-today":
│    1: resource "digitalocean_droplet" "critical-today" {
│ 
```
(note missing Droplet ID in the error message above).

What happens is that on 404 error `resourceDigitalOceanDropletRead()` unsets the ID and makes it impossible to execute further retries. After the patch error changes to:
```
│ Error: Error waiting for droplet (280226673) to become ready: couldn't find resource (61 retries)
│ 
│   with digitalocean_droplet.critical-today,
│   on critical-today-droplet.tf line 1, in resource "digitalocean_droplet" "critical-today":
│    1: resource "digitalocean_droplet" "critical-today" {
│ 
```
Note:

1. Present droplet ID.
2. Retry count.

PR #776, opened 2022-01-10
droplet: Refactor post-create polling code.

andrewsomething:
This PR refactors the code used to poll a Droplet's status after the request to create it. https://github.com/digitalocean/terraform-provider-digitalocean/pull/767 points out one issue with the existing implementation, but it also exposes another issue with it. So I've taken the opportunity to fix those and make the entire process more efficient.

It now first uses `waitForAction` to ensure that the Droplet create action completes successfully. This allows us to fail fast in the case of a real error instead of continuing to poll the non-existent Droplet. Polling the action endpoint allows for an error to be reported while polling the droplet endpoint just 404s (which we retry on) if the create fails.

The call to `waitForDropletAttribute` checking for the Droplet's `status` attribute to move from `new` to `active` is still made after that. This is possibly unnecessary, but I've kept it to make sure we don't introduce any unexpected behavior changes. (I'm not 100% positive that the action completing is entirely equivalent to the Droplet becoming active.)

It also removes the usage of `resourceDigitalOceanDropletRead` in `dropletStateRefreshFunc` in order to reduce the overall number of API calls made. Currently in `main`, each pass through `newDropletStateRefreshFunc` actually calls `client.Droplets.Get` twice. The final call in `resourceDigitalOceanDropletCreate` to `resourceDigitalOceanDropletRead` is also removed to cut another redundant API call.

So while this now makes calls to both the actions and droplets endpoints to make sure the Droplet has been successfully created, it is still more efficient and reduces the overall number of API call made substantially.

Additionally, there are a number of unrelated fixes to the Droplet acceptance tests so that they all now pass. These failure seem to be the result of moving the tests from using legacy Droplet plans to new ones without updating the expected disk sizes.

### DIFF
--- a/digitalocean/resource_digitalocean_droplet.go
+++ b/digitalocean/resource_digitalocean_droplet.go
@@ -286,23 +286,35 @@ func resourceDigitalOceanDropletCreate(ctx context.Context, d *schema.ResourceDa
 
 	log.Printf("[DEBUG] Droplet create configuration: %#v", opts)
 
-	droplet, _, err := client.Droplets.Create(context.Background(), opts)
-
+	droplet, resp, err := client.Droplets.Create(context.Background(), opts)
 	if err != nil {
 		return diag.Errorf("Error creating droplet: %s", err)
 	}
 
 	// Assign the droplets id
 	d.SetId(strconv.Itoa(droplet.ID))
-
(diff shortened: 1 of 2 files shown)


ChiefMateStarbuck (approved):

scotchneat (approved):
👍🏽

Issue #787, opened 2022-02-11
`digitalocean_project` environment defualts to "Development" if not initially set


collinsmuriuki:
I am importing an existing `digitalocean_project` to terraform. On DigitalOcean, the `environment` was not set when initially created. However, once I import and try running `terraform plan` I get this: `+ environment = "Development"`

I omitted the `environment` field after importing the resource, since it was null (not set on DigitalOcean), as well as due to the fact that it's an optional field according the docs.

### Affected Resource(s)
- digitalocean_project

### Expected Behavior
No changes should be detected

### Actual Behavior
A change in `environment` is detected

### Steps to Reproduce
- Create a project on DigitalOcean, set purpose as "Web Application" and do not set the environment
- Create an empty `digitalocean_project` and then import the project to terraform
- Add the required fields to the `digitalocean_project` resource; `name` and `purpose` in this case
- Run `terraform plan`

**Terraform Configuration Files**

```
terraform {
  required_providers {
    digitalocean = {
      source  = "digitalocean/digitalocean"
      version = "~> 2.0"
    }
  }
}

variable "do_token" {
}

provider "digitalocean" {
  token = var.do_token
}

```

**Debug Output**
https://gist.github.com/collinsmuriuki/c459254876a678d5ccbff54b08991faa

andrewsomething:
Thanks for the report @collinsmuriuki  We've got a fix in https://github.com/digitalocean/terraform-provider-digitalocean/pull/788

PR #788, opened 2022-02-11
projects: Environment is optional, don't set default.

andrewsomething:
Setting a default value for an optional filed can create an issue when importing a resource with that field not set. For projects, the environment is optional. 

We were missing an import test, but that wouldn't have actually caught this as the resource being imported would have been created by Terraform and thus would have had the default set. :upside_down_face: 

Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/787


mikeberezowski (approved):
Looks good. Approved

PR #722, opened 2021-10-13
WIP AppSpec Alerts

atombrella:
I don't find the documentation for the alerts under https://docs.digitalocean.com/reference/api/api-reference/#operation/create_app
It's quite unfinished, and needs more work before I'll ask for a review or assistance!

PR #797, opened 2022-02-25
apps: Implement support for configuring alert policies.

andrewsomething:
This PR builds on @atombrella's work in https://github.com/digitalocean/terraform-provider-digitalocean/pull/722 to get it over the line. It implement support for configuring alert policies for apps. 

As I mentioned in a comment on that PR, the destination for the alerts defaults to the email of the account. Alerts can be updated with additional emails or Slack webhooks, but this is not part of the app spec itself. You need to call a separate endpoint: https://docs.digitalocean.com/reference/api/api-reference/#operation/assign_alert_destinations

This initial work only supports the default destination. We'll need to think through how to properly support addition destinations in Terraform. It might require a separate app alert destinations resource.


scotchneat (approved):
lgtm

PR #818, opened 2022-04-21
Fix loop until timeout when updating App without triggering deployment

rgdcardoso:
The current DO App update implementation gets stuck in a loop until timeout each time a resource change doesn't trigger an App redeployment (update the App `name`, for instance) or if the deployment is so quick that it finishes before the first App GET is called to fetch the `deploymentID`.

The current PR fixes this issue, ensuring that if there is no `in_progress_deployment` field in the App GET, it cleanly exits the loop.


andrewsomething:
Hi @rgdcardoso,

Thank you so much for the PR, and sorry for the lack of response. :pray:  Unfortunately we somehow missed it. I've fixed this issue as I came across it myself while implementing another feature (https://github.com/digitalocean/terraform-provider-digitalocean/pull/831/commits/d95782344d77a609acf036501e9b54bf5b10d1cf in  https://github.com/digitalocean/terraform-provider-digitalocean/pull/831). I approached it in a slightly different manner as the in-progress deployment is not always available at the first `Apps.Get()`.

Issue #546, opened 2020-12-30
429 Too many requests or 429 API Rate limit exceeded


chasebolt:
### Terraform Version
```
Terraform v0.13.5
+ provider registry.terraform.io/-/aws v3.22.0
+ provider registry.terraform.io/-/digitalocean v2.3.0
+ provider registry.terraform.io/digitalocean/digitalocean v2.3.0
+ provider registry.terraform.io/hashicorp/aws v3.22.0
```

### Expected Behavior
Terraform to complete successfully

### Actual Behavior
Terraform is getting errors back from DO API stating either Too Many Requests or API Rate limit exceeded.

### Steps to Reproduce
Manage 100 domain records and it will fail while trying to refresh the state.

### Important Factoids
I am managing 10 domains with about 10-20 records each. I have dropped parallelism down to 1 and I still can't get a successful run without hitting one of these two errors. This isn't an absurdly large terraform run, and  I am unable to get past just a handful of state refreshes before it errors out.

zeppelinen:
+1. 
Same problem for me. I need to manage 200+ droplets.

lfarnell:
Based on the Docs
```
Requests through the API are rate limited per OAuth token. Current rate limits:

5,000 requests per hour
250 requests per minute (5% of the hourly total)
```
So one thing you can possibly do is reach out to support and see if they can increase the limit for `requests per minute`. The only other thing you might want to investigate is looking at breaking down the terraform code into a module per domain so that you won't hit the limit. I realize this might not be possible or suitable for your use case by something to think about.

joaomnmoreira:
+1.

Hundreds of domain records (same domain) to manage.

ghomem:
+1

This happens to me white managing a single large domain (~ 150 records)

djmaze:
Same problem here.

> So one thing you can possibly do is reach out to support and see if they can increase the limit for requests per minute.

I just tried that. The answer was simply that they are doing this "to protect the platform". So they won't increase the limits.

I now asked if they could implement at least something like a *burst limit* which would allow to overreach the rate limit for a short amount of time. That would probably help with use cases such as Terraform.

atmosx:
> Same problem here.
> 
> > So one thing you can possibly do is reach out to support and see if they can increase the limit for requests per minute.
> 
> I just tried that. The answer was simply that they are doing this "to protect the platform". So they won't increase the limits.
> 
[...]

PR #820, opened 2022-04-25
Switch to retryable http client addresses #546

tback:
I replaced the default http client with hashicorps retryable httpclient. That client detects and adapts to http status code 429. 
This is my first PR in this project.


tback:
Let me know if there is anything I can do to make this any easier on your end.

scotchneat:
@tback thank you for your contribution. I can see where this will be useful in many cases. 
I'm working on understanding the impact to the behavior of the provider and need to determine if there are any cases where we'd want to avoid retries before we accept this change.

tback:
Hi Cesar,

thanks for looking into this. One more thing to add since it might not be
obvious from the PR:
Hashicorp/go-retryablehttp behaves very well in this situation: it
evaluates the retry-after header if available and uses exponential backoff
as fallback (
https://pkg.go.dev/github.com/hashicorp/go-retryablehttp?utm_source=godoc#DefaultBackoff
).

scotchneat:
@tback, I was reading into that backoff behavior and think that's one of the great values this change would offer.

I do have some thoughts. 
1. It would be practical to allow users to override the retryablehttp client's [default values](https://pkg.go.dev/github.com/hashicorp/go-retryablehttp#Client) (like the `RetryWaitMin`, `RetryWaitMax`, and `RetryMax` values) to suite their needs. This would require adding some properties to the [Config](https://github.com/digitalocean/terraform-provider-digitalocean/blob/main/digitalocean/config.go#L21) struct, as well as schema values to the [provider](https://github.com/digitalocean/terraform-provider-digitalocean/blob/main/digitalocean/provider.go#L15) to load from environment variables. 
2. More notable, I'm thinking that a better place to apply this change would be in the [digitalocean/godo](https://github.com/digitalocean/godo) pacakge as that would enable any go client to leverage the retry capabilities.

tback:
@scotchneat thanks for your suggestions. They absolutely make sense to me. I won't have time to implement them over the next months though.
Anybody can take it from here?

tback:
Nevermind.

PR #822, opened 2022-04-29
alerts: Support load balancer alert types.

andrewsomething:
A number of load balancer alert types are now supported by the API. This enables them in Terraform as well.


scotchneat (approved):
LGTM

Issue #828, opened 2022-05-16
Add timeout configuration for digitalocean_droplet resource


dfradejas:
Hi,

When you create a big droplet from a snapshot in Digitalocean, it usually spends more than 1 hour creating this droplet. When you do it with terraform, the provider throws a timeout error because default timeout is 1 hour in the provider.

Could be possible to add timeouts block configuration to digitalocean_droplet resources?

Regards,
Daniel.

ngharrington:
I have proposed #839 as a possible solution but am very open to feedback on choices made.

dfradejas:
Hi, 

I'm trying to create the resource with timeout configuration, but it is not doing anything, the default timeout (1h) is still present, i can't configure it to, for example, 2 hours.

I am writting this block inside digitalocean_droplet resource and it goes to 1h again:
`  timeouts {
    create = "1m"
  }
`

How can i configure it?

Regards,
Daniel

dfradejas:
The droplet configuration is:

```
resource "digitalocean_droplet" "postgres_master_droplet" {
      ...
      timeouts {
          create = "2h"
          delete = "2h"
          update = "2h"
        }
    }
```
And the error code after an hour:
```
digitalocean_droplet.postgres_master_droplet: Still creating... [57m51s elapsed]
digitalocean_droplet.postgres_master_droplet: Still creating... [58m1s elapsed]
digitalocean_droplet.postgres_master_droplet: Still creating... [58m11s elapsed]
digitalocean_droplet.postgres_master_droplet: Still creating... [58m21s elapsed]
digitalocean_droplet.postgres_master_droplet: Still creating... [58m31s elapsed]
digitalocean_droplet.postgres_master_droplet: Still creating... [58m41s elapsed]
digitalocean_droplet.postgres_master_droplet: Still creating... [58m51s elapsed]
digitalocean_droplet.postgres_master_droplet: Still creating... [59m1s elapsed]
digitalocean_droplet.postgres_master_droplet: Still creating... [59m11s elapsed]
digitalocean_droplet.postgres_master_droplet: Still creating... [59m21s elapsed]
digitalocean_droplet.postgres_master_droplet: Still creating... [59m31s elapsed]
digitalocean_droplet.postgres_master_droplet: Still creating... [59m41s elapsed]
digitalocean_droplet.postgres_master_droplet: Still creating... [59m51s elapsed]
[...]

PR #839, opened 2022-06-15
Allow for Timeouts to be set for Droplet Create Operations

ngharrington:
This is a proposal to solve issue #828  . I do not necessarily think it is an ideal or preferred solution and I am more than open to feedback.

Add Timeout to the resource_digital_ocean_droplet schema for Create, Update, Delete operations. These default to values that keep behavior the same as before if they are not set.  Namely, timeout for all operations is at least 60 minutes. Specifically, by default, timeout for the &resource.StateChangeConf inside waitForDropletAttribute function stays the same by default.

Notably, for deletes and updates, these attribute refreshes happen multiple times per operation, so I decided to just pass through the relevant timeout for every attribute refresh.  Meaning the true timeout for a delete or update would be a multiple of the set timeouts in the schema. 

I additionally updated the droplet tests. The slug used for the image in the current code appears to not exist from my PoV, I changed to an ubuntu distribution but the choice doesn't matter to me one way or the other.


ngharrington:
@andrewsomething do you or other maintainers have thoughts on this PR, or what I may be able to do to bring it closer to what you folks would want.  (alternatively, if you want to forgo this feature for now due to the way the timeouts behave with respect to state refreshes)

danaelhe:
Thank you so much for your efforts and time on this contribution! I just kicked off the Tests workflow for this PR and will review your changes as soon as we can.

ngharrington:
Thank you so much for the response and for kicking of the tests!

andrewsomething (approved):
:+1:  Thank you for this improvement. Sorry for taking so long to get it merged. As you say, this might not be the ideal solution, but I do think it is a good step in the right direction and will help unblock people impacted by https://github.com/digitalocean/terraform-provider-digitalocean/issues/828

I did push one additional commit changing Ubuntu 21.10 to 22.04 as it is a long term support release. So we will not need to update it again as soon.

PR #840, opened 2022-06-16
Remove database cluster and/or database firewall rule from state if missing

liamjcooper:
Uncovered an issue whereby refreshing Terraform state would return an error stating a database cluster didn't exist if a previously existing cluster with a firewall rule is manually deleted rather than deleting the resource from its state. This PR hopes to address this problem by checking the status code, and if it's a 404, then clear the ID for that resource and return nil from its read method.

Since one rule can only be attached to one cluster, simply deleting the rule from state should be safe.

### DIFF
--- a/digitalocean/resource_digitalocean_database_firewall.go
+++ b/digitalocean/resource_digitalocean_database_firewall.go
@@ -89,8 +89,12 @@ func resourceDigitalOceanDatabaseFirewallRead(ctx context.Context, d *schema.Res
 	client := meta.(*CombinedConfig).godoClient()
 	clusterID := d.Get("cluster_id").(string)
 
-	rules, _, err := client.Databases.GetFirewallRules(context.TODO(), clusterID)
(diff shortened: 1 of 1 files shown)


andrewsomething (approved):
:+1:  Thanks for the PR @liamjcooper! 

I did push one additional commit with some small changes:

```diff
-               if resp.StatusCode == 404 {
+               if resp != nil && resp.StatusCode == 404 {
                        d.SetId("")
-                       return nil;
+                       return nil
                }
```

We need to first ensure that `resp` is not `nil` before accessing `resp.StatusCode`. If we got into a state were there was an error but no response (e.g. a networking problem), it would panic.

liamjcooper:
Awesome, thank you! Golang is new to me so the feedback is much appreciated.

MattIPv4:
Hey, @liamjcooper - Thanks a ton for this sweet PR! 😄

Issue #842, opened 2022-06-20
Unable to retry app deployment after it fails.


Igosuki:
# Bug Report

---

## Describe the bug
When first deploy of a job with a service fails, the plugin fails to register or import the resource in tfstate, resulting in a stale state and inability to do a second `terraform apply` to retry.

### Affected Resource(s)
- digitalocean_app

### Expected Behavior
Created service should exist in tfstate because it is created just not deployed.

### Actual Behavior
Getting an error when importing
```
digitalocean_app.kratos_prod: Importing from ID "..."...
digitalocean_app.kratos_prod: Import prepared!
  Prepared digitalocean_app for import
digitalocean_app.kratos_prod: Refreshing state... [id=...?i=...]
╷
│ Error: No active deployment found for app: kratos (...)
│ 
```

PR #843, opened 2022-06-21
apps: Only warn on read if there is no active deployment.

andrewsomething:
`resourceDigitalOceanAppRead` currently errors out if there is no active deployment. This can cause issues in a few circumstances as an app can exist in a state where there are no active deployments. Changing this to be a warning allows for importing existing apps that do not have active deployments. It also allows for `terraform apply` to succeed if something outside of Terraform has caused an existing app to have no active deployments. Since deployments can be triggered by pushes to the app's repo, Terraform doesn't really handle the state of deployments, only the "app" resource.

This does not change the existing behavior for creates of new apps. We still poll until the app deployment succeeds and error if it fails. It's possible we might want to reconsider this in the future, but that has other considerations. This does help recovering from a failed create. Running `terraform untaint digitalocean_app.{name}` will now allow applies to complete successfully.

Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/842

### DIFF
--- a/digitalocean/resource_digitalocean_app.go
+++ b/digitalocean/resource_digitalocean_app.go
@@ -122,7 +122,12 @@ func resourceDigitalOceanAppRead(ctx context.Context, d *schema.ResourceData, me
 	if app.ActiveDeployment != nil {
 		d.Set("active_deployment_id", app.ActiveDeployment.ID)
 	} else {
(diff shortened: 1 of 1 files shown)


scotchneat (approved):
lgtm

Issue #848, opened 2022-07-13
Allow marking as Default Project when provisioning projects


yordis:
### Describe the solution you'd like

Allow to provision `digitalocean_project` resource as "Default Project" 
 
### Describe alternatives you've considered

Manually do it in the Console UI.

### Additional context

![Screen Shot 2022-07-13 at 4 01 17 AM](https://user-images.githubusercontent.com/4237280/178682480-eae11474-2b26-4ce6-a94c-b7236d97a26c.png)

yordis:
closed by #860

PR #849, opened 2022-07-15
Altering is_default to be configurable for projects

danaelhe:
Per #848.

@Yordis had a great suggestion for allowing to provision digitalocean_project resource as "Default Project".  Making "is_default" configurable. This means:

- You can initialize a new project to be the default.
- You can’t edit a project to no longer be default (`400 (request "acb5906e-bf48-49e3") you must always have a default project`) even if you have other projects, it won’t automatically assign another project to be default
- You can’t delete a project that is the default (^^)
- You must first assign another project to be default first before you can delete the old default project

### DIFF
--- a/digitalocean/resource_digitalocean_project_test.go
+++ b/digitalocean/resource_digitalocean_project_test.go
@@ -43,6 +43,42 @@ func TestAccDigitalOceanProject_CreateWithDefaults(t *testing.T) {
 	})
 }
 
+func TestAccDigitalOceanProject_CreateWithIsDefault(t *testing.T) {
+
+	expectedName := generateProjectName()
+	expectedIsDefault := "true"
+	createConfig := fixtureCreateWithIsDefault(expectedName, expectedIsDefault)
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: testAccProviderFactories,
+		CheckDestroy:      testAccCheckDigitalOceanProjectDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: createConfig,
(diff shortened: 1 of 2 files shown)


yordis:
💜

digitalocean/resource_digitalocean_project_test.go
```diff
@@ -43,6 +44,75 @@ func TestAccDigitalOceanProject_CreateWithDefaults(t *testing.T) {
...
+				t.Errorf("Error locating default project %s", defaultProjErr)
+			}
+			originalDefaultProject = defaultProject
+		},
+		ProviderFactories: testAccProviderFactories,
+		CheckDestroy:      testAccCheckDigitalOceanProjectDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config:             createConfig,
+				ExpectNonEmptyPlan: true,
```
danaelhe, line 74: Without this, I was getting the following error:

```
resource_digitalocean_project_test.go:57: Step 1/1 error: After applying this test step and performing a `terraform refresh`, the plan was not empty.
        stdout
        
        
          ~ update in-place
        
          # digitalocean_project.myproj will be updated in-place
          ~ resource "digitalocean_project" "myproj" {
                id         = "1e7c1f73-9924-4cb2-b706-1ef6bf339dbb"
              ~ is_default = false -> true
                name       = "tf-proj-test-4023944325181221421"
                # (6 unchanged attributes hidden)
            }
        
        Plan: 0 to add, 1 to change, 0 to destroy.
```

ExpectNonEmptyPlan can be set to true for specific types of tests that are looking to verify that a diff occurs. Doc [here](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource#TestStep)

PR #859, opened 2022-08-05
chore: fix incorrect heading in bug template

artis3n:
Based on the context in the section and the earlier, larger heading for expected behavior, along with looking at older issues in the repo, I'm pretty sure this title is supposed to be "Terraform version."


andrewsomething (approved):
:+1:  LGTM

PR #868, opened 2022-08-17
make user optional in db connection pool, update acc tests

DMW2151:
The DO API allows DB pools to be created without a user specified. This PR modifies the `digitalocean_database_connection_pool` resource and removes user as a required field. Without user specified, pool uses the connection credentials of the inbound user. 

Adds  new test case for creates w.o specified user, `TestAccDigitalOceanDatabaseConnectionPool_InboundUser`, and modifies `TestAccDigitalOceanDatabaseConnectionPool_Basic` to cover pool updates.


andrewsomething (approved):
:+1:  LGTM

PR #873, opened 2022-08-30
Suppress diffs on forced Redis version upgrades

scotchneat:
To ensure customer Redis instances do not get destroyed and recreated after their Redis clusters are forced to upgrade, this change updates the `DiffSuppressFunc` logic for Database Cluster resources to suppress diffs when the DB engine is `redis` and `version` is newer on the remote resource than what is specified in the configuration.

This should also help us avoid having to make this change for forced upgrades in the future.

for APICLI-1526

### DIFF
--- a/digitalocean/resource_digitalocean_database_cluster.go
+++ b/digitalocean/resource_digitalocean_database_cluster.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"net/url"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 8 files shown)


DMW2151:
May want to generalize the test `TestAccDigitalOceanDatabaseCluster_oldRedisVersion` to accommodate all version changes.

scotchneat:
I switched this to a Draft PR so I can look into integrating a call to the [databases_list_operations](https://docs.digitalocean.com/reference/api/api-reference/#operation/databases_list_options) endpoint to ensure the acceptance tests can update dynamically.

danaelhe (approved):
LGTM

PR #882, opened 2022-09-21
Add SOA as possible record type

Nosmoht:
Add `SOA` as possible record type.

Without it it's not possible to manage `SOA` records as the following error would be raised.

```shell
 Error: expected type to be one of [A AAAA CAA CNAME MX NS TXT SRV], got SOA
```


Nosmoht:
Don't know what to do to make `acceptance` test green as it was canceled. Could someone please give me some advice?

andrewsomething (approved):
:+1:  LGTM, thanks for the contribution!

PR #893, opened 2022-11-02
Update Monitor Alert resource with new DBAAS public alert types

dweinshenker:
Adding additional monitoring alerts that are now supported for dbaas


andrewsomething (approved):
:+1:  LGTM; thanks!

PR #897, opened 2022-11-11
Support HTTP idle timeout & Project ID

StephenVarela:
LBAAS-2430


digitalocean/resource_digitalocean_loadbalancer_test.go
```diff
@@ -728,6 +740,8 @@ resource "digitalocean_loadbalancer" "foobar" {
 
   enable_proxy_protocol    = true
   enable_backend_keepalive = true
+  http_idle_timeout_seconds = 60
```
scotchneat, line 743: I'm wondering if it would be better to set this to a different value from the default to ensure the resource is set to the provided value.

```suggestion
  http_idle_timeout_seconds = 45
```

scotchneat (changes requested):

bentranter (approved):

PR #918, opened 2023-01-24
databases: Support upgrading the database version

### LINKED ISSUE #471: In-place upgrade of to new PostgreSQL major version
Currently, when changing the `version` of a `digitalocean_database_cluster` resource, terraform will try to replace the resource. It should instead do an in-place upgrade of the major version, as one can do through the control panel: https://www.digitalocean.com/docs/databases/postgresql/how-to/upgrade-version/

scotchneat:
Enables upgrading the database version without forcing a new database deployment.

For DBAAS-3899

Closes #471


danaelhe (approved):
lgtm!

andrewsomething (approved):
:+1:  Looks great!

PR #927, opened 2023-02-03
Package reorganization 

andrewsomething:
This pull request reorganizes the structure of this project to use multiple packages. As the provider has grown, it has become a bit unwieldy to work with as the majority of files were all in the same directory. By breaking this up into subpackages, it should be easier to quickly find the files you need. It should also simplify acceptance testing by being able to run the tests for a subpackage with out needing to craft regexps. For example, to only run the acceptance tests for the account package use:

    make testacc PKG_NAME=digitalocean/account

When `PKG_NAME` is excluded, all acceptance tests are run.

One place where this does add a bit of complexity is with the test sweepers. In order to collect sweepers across the packages, we need to use "blank" imports which can be easy to forget. See: `digitalocean/sweep/sweep_test.go` In a future iteration, we could potentially generate that. The AWS provider has some prior art that could provide some inspiration.

While this is a huge PR, most of the changes are file renames and moving code to break import cycles.


danaelhe (approved):
Huge progress from what we have 👍

scotchneat (approved):
Looks great!

PR #933, opened 2023-02-08
Use consistent naming for all Droplets created in tests.

andrewsomething:
All resources created by our provider's acceptance tests should use a standard naming structure to make sweeping simpler and reduce risk of running the sweeper for external contributors. This PR ensures we are using a consistent naming for all Droplets created in tests. 

I'll be following this up with PRs for other resource types. As Droplets are used by other resources, this is likely the most invasive PR. Hopefully the others should be easier to review.

### DIFF
--- a/digitalocean/droplet/sweep.go
+++ b/digitalocean/droplet/sweep.go
@@ -33,15 +33,18 @@ func sweepDroplets(region string) error {
 	}
 	log.Printf("[DEBUG] Found %d droplets to sweep", len(droplets))
 
+	var swept int
 	for _, d := range droplets {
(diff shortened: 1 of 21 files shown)


digitalocean/droplet/sweep.go
```diff
@@ -33,15 +33,18 @@ func sweepDroplets(region string) error {
 	}
 	log.Printf("[DEBUG] Found %d droplets to sweep", len(droplets))
 
+	var swept int
 	for _, d := range droplets {
-		if strings.HasPrefix(d.Name, "foo-") || strings.HasPrefix(d.Name, "bar-") || strings.HasPrefix(d.Name, "baz-") || strings.HasPrefix(d.Name, "tf-acc-test-") || strings.HasPrefix(d.Name, "foobar-") {
+		if strings.HasPrefix(d.Name, sweep.TestNamePrefix) {
```
andrewsomething, line 38: This is the main benefit of these change. Instead of needing to check for multiple different patterns before deleting, we just need to look for one. It simplifies things and should make the sweeper a safer to run.

danaelhe (approved):
🥳

scotchneat (approved):
🚀

PR #942, opened 2023-02-10
Add sweeper for VPCs.

andrewsomething:
Straight forward one. Just a sweeper. Already using the correct naming pattern.


scotchneat (approved):
Looks good 🚀

PR #958, opened 2023-03-01
Remove .go-version and add to .gitignore

ChiefMateStarbuck:
The `.go-version` file is unnecessary. Removing to reduce clutter.


danaelhe (approved):
thanks 👍

PR #963, opened 2023-03-20
Add registry_integration to Schema

mohsenSy:
registry_integration can be true to integrate the kubernetes cluster with container registry, or false to disable this integration If there is no container registry in the team's or user's account then terraform will taint the resource and re-create the clsuter on next apply, it is recommended to make your doks resource depends on the Digital Ocean container registry resource to avoid re-creating the kubernetes cluster


digitalocean/kubernetes/resource_kubernetes_cluster.go
```diff
@@ -294,6 +301,18 @@ func resourceDigitalOceanKubernetesClusterCreate(ctx context.Context, d *schema.
...
+		err = enableRegistryIntegration(client, cluster.ID)
+		if err != nil {
+			return diag.Errorf("Error enabling registry integration: %s", err)
+		}
+	} else {
+		err = disableRegistryIntegration(client, cluster.ID)
+		if err != nil {
+			return diag.Errorf("Error disabling registry integration: %s", err)
+		}
+	}
```
andrewsomething, line 314: In the create method, I think we can remove this else statement. As the registry integration is not enabled, this adds an unneeded API call.

andrewsomething (changes requested):
Thank you for this great contribution! Sorry that we've taken so long to review it.

I had one small suggestion in-line. It would also be great to document the new option here: 

https://github.com/digitalocean/terraform-provider-digitalocean/blob/main/docs/resources/kubernetes_cluster.md#argument-reference

It would also be good to add an acceptance test. Let us know if you are interested in doing that. If not, we can take it from here.

digitalocean/kubernetes/resource_kubernetes_cluster.go
mohsenSy: Thanks for the note, I changed the code now

mohsenSy:
@andrewsomething could you please have a second look? I added some tests, I hope they are correct. I cannot afford to test them locally

andrewsomething (approved):
Thanks for the quick follow up! I've pushed a commit that fixes up a few things with the tests and consolidates them a bit. Creating multiple clusters adds some over head and the full suite already takes quite awhile to run. 

We'll be looking to cut a release soon that includes this!

Issue #878, opened 2022-09-08
Change `id` value of `digitalocean_database_replica` to match replica UUID


timkrins:
### Is your feature request related to a problem? Please describe.

The `id` that is returned by the current `digitalocean_database_replica` resource is generated (in `makeReplicaId`) rather than fetched from the DigitalOcean API.
The new `id` field was added to the API yesterday, under
`{"replicas":[{"id" ...` at `https://api.digitalocean.com/v2/databases/{cluster_id}/replicas`
It has not yet landed in the public API documentation, but it is there.

### Describe the solution you'd like

The `digitalocean_database_replica` `id` field should track the DigitalOcean replica UUID so that replicas can be used with the `digitalocean_database_firewall` provider.
 
### Describe alternatives you've considered

Not sure of any other alternative.

### Additional context

https://github.com/digitalocean/terraform-provider-digitalocean/issues/224
https://github.com/digitalocean/terraform-provider-digitalocean/issues/639
https://github.com/digitalocean/terraform-provider-digitalocean/issues/799

cphalpin:
I'm currently having to hardcode the id of my replica since there doesn't seem to be any way to get it from Terraform at all (I tried fetching it from the `digitalocean_database_replica` data source as a workout around). `doctl` returns the UUID in the `id` field as expected.

danaelhe:
Hi there, 

Thank you for bringing this to our attention. This has been addressed in #921 and will be available in the next release (which we plan to do in the next few days).

Thank you so much for your feedback! Please don't hesitate to reach out again. 

Dana

timkrins:
Hello,
Thanks for completing this!

Can I ask how importing an existing `digitalocean_database_replica` resource works?

I can't seem to get the UUID from imported resources.
I'm still investigating but will update when I have done a bit more research.

timkrins:
I think with the above changes the replica UUID is only ever assigned when the replica is created via terraform, rather than imported.
I managed to perform an import, then export my (remote) terraform state, update the UUID that I got from the DO UI manually in the JSON, then push my state back up.
Maybe there is a way to look up the UUID when performing the `terraform import` command?

andrewsomething:
Good catch @timkrins I have a PR up that should fix the problem for imports https://github.com/digitalocean/terraform-provider-digitalocean/pull/964

PR #964, opened 2023-03-21
database replicas: set uuid on read.

andrewsomething:
As pointed out in https://github.com/digitalocean/terraform-provider-digitalocean/issues/878, we've exposed the `uuid` of a database replica, but it is not set during imports. We only set it in the create method. It should be set in the read method as well.

Test change on its own shows the problem:

```
=== CONT  TestAccDigitalOceanDatabaseReplica_importBasic
    import_database_replica_test.go:23: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=1) {
         (string) (len=4) "uuid": (string) (len=36) "bcb92ee3-5f0b-4b90-8a43-a0350b718396"
        }
--- FAIL: TestAccDigitalOceanDatabaseReplica_importBasic (509.30s)
```

Fixed with the code change:

```
=== RUN   TestAccDigitalOceanDatabaseReplica_importBasic
=== PAUSE TestAccDigitalOceanDatabaseReplica_importBasic
=== CONT  TestAccDigitalOceanDatabaseReplica_importBasic
--- PASS: TestAccDigitalOceanDatabaseReplica_importBasic (601.54s)
PASS
ok      github.com/digitalocean/terraform-provider-digitalocean/digitalocean/database   601.579s
```

### DIFF
--- a/digitalocean/database/import_database_replica_test.go
+++ b/digitalocean/database/import_database_replica_test.go
@@ -41,7 +41,7 @@ func TestAccDigitalOceanDatabaseReplica_importBasic(t *testing.T) {
 				// Requires passing both the cluster ID and replica name
 				ImportStateIdFunc: testAccDatabaseReplicaImportID(resourceName),
 				// The DO API does not return the size on read, but it is required on create
-				ImportStateVerifyIgnore: []string{"size", "uuid"},
+				ImportStateVerifyIgnore: []string{"size"},
 			},
 			// Test importing non-existent resource provides expected error.
 			{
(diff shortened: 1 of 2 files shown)


scotchneat (approved):
👍🏽

JoeyHoutenbos:
Thanks for fixing this. Any idea when it will be released?

andrewsomething:
@JoeyHoutenbos  We'll get a new release cut: https://github.com/digitalocean/terraform-provider-digitalocean/pull/965

Issue #968, opened 2023-03-29
Add possibility to use backup_restore for a new database cluster


bery:
### Is your feature request related to a problem? Please describe.
I would like to be able to restore a managed database backup to a new cluster. It's very frustrating to have to create a backup first and then restore it. 

### Describe the solution you'd like
Add the possibility to take advantage of DO API https://docs.digitalocean.com/reference/api/api-reference/#operation/databases_create_cluster and backup_restore option
 
### Describe alternatives you've considered
Switching to AWS

### Additional context
None is needed if its feasible

danaelhe:
Hi @bery !

Thank you for the write up! It's funny you bring this up because I actually just added this functionality in another DO tool, [doctl](https://github.com/digitalocean/doctl/pull/1346), so it probably won't be too much of an effort to add it terraform too. I've created a ticket for this and have moved it to the top of the backlog, hopefully we can get to it soon.

Thanks,
Dana

bery:
That's what I call perfect timing! Thank you very much @danaelhe, if you need any help with testing, let me know!

andrewsomething:
Closed via https://github.com/digitalocean/terraform-provider-digitalocean/pull/970

AAverin:
The documentation and examples at https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/resources/database_cluster just don't make sense to me.
The field is called `database_name` and example implies that a database is used for the backup, while description states `Create a new database cluster based on a backup of an existing cluster`.
So is it a cluster or a database? 
Example also shows how to restore backup from another, already existing cluster and its database.
What if neither the cluster and database don't exist yet and are just being created?

AAverin:
@andrewsomething @danaelhe could you please clarify?

danaelhe:
Hi there, 

Thank you for the write up- I've answered your questions in sections:

> Are there examples on how does it work exactly, maybe some article with an explanation?
Which db cluster are you trying to create? [Here's](https://docs.digitalocean.com/products/databases/postgresql/how-to/restore-from-backups/) an article for back ups for Postgres. And [here's](https://docs.digitalocean.com/products/databases/mysql/how-to/restore-from-backups/) an article on back ups for MySQL.


> Would DO automatically do backups of my managed database? 
[...]

PR #970, opened 2023-04-07
databases: Add backup-restore functionality to db create

danaelhe:
Addresses #968

Adding API parity of create a new database cluster based on a backup of an existing cluster:

```
resource "digitalocean_database_cluster" "foobar" {
	name       = "%s"
	engine     = "pg"
	version    = "11"
	size       = "db-s-1vcpu-2gb"
	region     = "nyc1"
	node_count = 1
	tags       = ["production"]
  }
resource "digitalocean_database_cluster" "foobar_backup" {
	name       = "%s"
	engine     = "pg"
	version    = "11"
	size       = "db-s-1vcpu-2gb"
	region     = "nyc1"
	node_count = 1
	tags       = ["production"]
	
	backup_restore {
	  database_name  = "%s"
	}
	depends_on = [
		digitalocean_database_cluster.foobar
	  ]
  }
```

APICLI-2018


digitalocean/database/resource_database_cluster_test.go
```diff
@@ -384,6 +384,34 @@ func TestAccDigitalOceanDatabaseCluster_WithVPC(t *testing.T) {
 	})
 }
 
+func TestAccDigitalOceanDatabaseCluster_WithBackupRestore(t *testing.T) {
```
danaelhe, line 387: This test passes but it gave me some unexpected issues that I still can't make sense of. Commented below.

digitalocean/database/resource_database_cluster_test.go
```diff
@@ -384,6 +384,34 @@ func TestAccDigitalOceanDatabaseCluster_WithVPC(t *testing.T) {
...
+	var backupDatabase godo.Database
+
+	originalDatabaseName := acceptance.RandomTestName()
+	backupDatabasename := acceptance.RandomTestName()
+
+	backUpRestoreConfig := fmt.Sprintf(testAccCheckDigitalOceanDatabaseClusterConfigWithBackupRestore, originalDatabaseName, backupDatabasename, originalDatabaseName)
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:          func() { acceptance.TestAccPreCheck(t) },
+		ProviderFactories: acceptance.TestAccProviderFactories,
```
danaelhe, line 398: adding the `CheckDestroy:      testAccCheckDigitalOceanDatabaseClusterDestroy` here for some reason results in a `cluster cannot be found`.

So naturally I figured it was a timing issue between the cluster being online and the call to preform a backup. So I broke up the calls in the test below and added a sleeper but that resulted in a `code = Internal desc = failed to list cluster backups before a fork attempt` [here](https://github.internal.digitalocean.com/digitalocean/cthulhu/blob/master/docode/src/do/teams/paas/dbaas/dbaas/server.go#L1707)

What ended up working consistently for this test was 1) removing the `checkDestory` func and 2) combining the create cluster config and backup restore config into one config.

Not sure why this behavior occurred...

digitalocean/database/resource_database_cluster.go
```diff
@@ -532,6 +562,58 @@ func waitForDatabaseCluster(client *godo.Client, d *schema.ResourceData, status
 	return nil, fmt.Errorf("Timeout waiting to database cluster to become %s", status)
 }
 
+func waitForDatabaseBackups(client *godo.Client, d *schema.ResourceData, originalDatabaseName string) error {
+	var (
+		tickerInterval = 15 * time.Second
+		timeoutSeconds = d.Timeout(schema.TimeoutDelete).Seconds()
```
andrewsomething, line 568: Not sure this needs to be user configurable, but either way `TimeoutDelete` probably isn't what we want to use here. It returns 20m by default. 


https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.24.1/helper/schema#ResourceData.Timeout

digitalocean/database/resource_database_cluster.go
```diff
@@ -261,6 +279,18 @@ func resourceDigitalOceanDatabaseClusterCreate(ctx context.Context, d *schema.Re
...
 
+	if v, ok := d.GetOk("backup_restore"); ok {
+		opts.BackupRestore = expandBackupRestore(v.([]interface{}))
+		originalDBName := opts.BackupRestore.DatabaseName
+		// waits for a a back up to be created before forking
+		err := waitForDatabaseBackups(client, d, originalDBName)
+		if err != nil {
+			d.SetId("")
+			return diag.Errorf("Error retrieving database cluster backups: %s", err)
+		}
```
andrewsomething, line 290: I wonder if this might be better as a test step? I'm not sure a real user would have this problem. So in most cases, it's an unneeded API call.

danaelhe: Oh that's a good point. Highly unlikely someone will fork a newly created database. Just moved the logic to the test 👍

digitalocean/database/resource_database_cluster.go
danaelhe: Oh yes I see- thanks for the docs!

andrewsomething (approved):

PR #972, opened 2023-04-14
Fix inconsistencies in digitalocean_uptime_alert documentation

nicwortel:
Hi!

I noticed some inconsistencies in [the documentation of the `digitalocean_uptime_alert` resource](https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/resources/uptime_alert). The `type` argument mentions the options `ping`, `http` and `https`, which actually belong to the `digitalocean_uptime_check` resource instead. When diving into the source code I also noticed some other issues in the documentation.

This PR should fix those, and also adds the contents of the `notifications` object to the argument reference. Last but not least I have fixed some formatting issues and a small typo in the example code.

### DIFF
--- a/docs/resources/uptime_alert.md
+++ b/docs/resources/uptime_alert.md
@@ -11,26 +11,26 @@ resource. Uptime Alerts provide the ability to add alerts to your [DigitalOcean 
 ### Basic Example
 
 ```hcl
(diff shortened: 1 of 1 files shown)


andrewsomething (approved):
:+1:  Thanks for cleaning this up!

PR #985, opened 2023-04-29
custom image: use custom create timeout.

andrewsomething:
Custom images can take a long time to create. There is a hard coded timeout here, but that is actually ignored. The default create timeout for Terraform is 20m. To support anything longer and make it configurable, we need to set a `ResourceTimeout` block in the schema.


danaelhe (approved):
👍

Issue #986, opened 2023-05-02
Custom create timeout for database cluster is ignored, default 20 min (delete) timeout is used


bery:
# Bug Report
Custom create timeout is ignored and a default value for delete is used

---

## Describe the bug
Custom create timeout is ignored and a default value for delete is used

### Affected Resource(s)
- digitalocean_database_cluster

### Expected Behavior
Custom create timeout should be honored

### Actual Behavior
Delete timeout is used

### Steps to Reproduce
```
resource "digitalocean_database_cluster" "database-cluster" {
  name       = local.instance_name
  engine     = var.engine
  version    = var.engine_version
  size       = var.size
  region     = var.region
  node_count = var.node_count
  backup_restore {
    database_name = var.backup_restore.database_name
  }
  timeouts {
    create = "3h"
  }
}
```

## Additional context
-

andrewsomething:
Thanks for flagging this. The fix should be available in the latest release: https://github.com/digitalocean/terraform-provider-digitalocean/releases/tag/v2.28.1

PR #987, opened 2023-05-02
databases: Fix custom create timeouts.

andrewsomething:
Looks like we accidentally used `TimeoutDelete` where it should be `TimeoutCreate` :man_facepalming: 

Also noticed `=` when it should be `>=` to prevent the corner case of the timeout being set lower than the interval. 

Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/986


danaelhe (approved):
thanks!!

PR #990, opened 2023-05-03
Set initial_filesystem_type and initial_filesystem_label when reading volume resource

mohsenSy:


andrewsomething (changes requested):
Hi @mohsenSy, thanks for looking into this one.  Unfortunately, it is a bit more complicated that this. 

`initial_filesystem_type` and `initial_filesystem_label` are create time options that should not be set on read. They mean different things than `filesystem_type` and `filesystem_label` which are computed attributes that can change (e.g. the user re-formats the volume manually). There is some context in https://github.com/digitalocean/terraform-provider-digitalocean/pull/136. We also need to support creating un-formatted volumes. As the API reports the current file system type, this would cause Terraform to try to recreate the volume once formatted. 

We could potentially create a custom import function that sets them on import. Though that would effectively make them required on import.  It would take some testing, by maybe a `DiffSuppressFunc` could work here? :thinking:

PR #1004, opened 2023-07-06
doc: how to get id for record import

nimboya:
This changes explain how a user can get the `id` of a DNS record which is required when importing an existing DNS record


andrewsomething (approved):
:+1:  Thanks for the contribution!

PR #1007, opened 2023-07-14
k8s: Add destroy_all_associated_resources option.

andrewsomething:
This adds an option to destroy all associated DigitalOcean resources created via the Kubernetes API (load balancers, volumes, and volume snapshots) along with the cluster when it is destroyed.


bentranter (approved):
LGTM!

PR #1014, opened 2023-08-08
Docs: Update Version of Postgress DB

danaelhe:
When running the example as is,  you get a:

`│ Error: Error creating database cluster: POST https://api.digitalocean.com/v2/databases: 422 (request "4f7f-a1a4-") invalid cluster engine version`

changing the version to 13 worked fine.


andrewsomething (approved):
:+1:  Good catch!

Looks like 15 is the most recent. Maybe use that since it will presumably be around longer?

```
$ doctl databases options versions ls
Engine     Versions
mongodb    [4.4,5.0,6.0]
mysql      [8]
pg         [12,13,14,15]
redis      [7]
```

danaelhe:
> 👍 Good catch!
> 
> Looks like 15 is the most recent. Maybe use that since it will presumably be around longer?
> 
> ```
> $ doctl databases options versions ls
> Engine     Versions
> mongodb    [4.4,5.0,6.0]
> mysql      [8]
> pg         [12,13,14,15]
> redis      [7]
> ```

Good point- just updated 👍

andrewsomething (approved):

PR #1016, opened 2023-08-08
Configure Godo Retryable HTTP Client

danaelhe:
Update terraform-provider to use godo's retryable http client. 

When the retryable feature is not set or set to 0 (`DIGITALOCEAN_HTTP_RETRY_MAX=0 `) and you reach the 250 req/min threshold- you are met with a 429 error:
```
│ Error: Error retrieving Database User: GET https://api.digitalocean.com/v2/databases/f7ce-4843-a0d9/users/86-app: 429 (request "adfaa613-e244-4cb1-aa4d-bd82d14bef7b") Too many requests
```

When the retryable feature is set (`DIGITALOCEAN_HTTP_RETRY_MAX=2 `) and you reach the 250 req/min threshold- the client performs automatic retries and exponential backoff and continues:
```
2023-08-08T15:13:01.396-0400 [DEBUG] provider.terraform-provider-digitalocean: 2023/08/08 15:13:01 [DEBUG] GET https://api.digitalocean.com/v2/databases/7d76-48e3-/users/173-app (status: 429): retrying in 44s (3 left)
```

The retryable feature to be on by default, with 4 tries.


digitalocean/config/config.go
```diff
@@ -76,25 +74,28 @@ func (c *Config) Client() (*CombinedConfig, error) {
...
+			RetryWaitMin: godo.PtrTo(c.HTTPRetryWaitMin),
+			RetryWaitMax: godo.PtrTo(c.HTTPRetryWaitMax),
+		}
+
+		godoOpts = []godo.ClientOpt{godo.WithRetryAndBackoffs(retryConfig)}
 
-		client = retryableClient.StandardClient()
 		client.Transport = &oauth2.Transport{
 			Base:   client.Transport,
 			Source: oauth2.ReuseTokenSource(nil, tokenSrc),
```
danaelhe, line 93: I was a little surprised I needed to maintain this transport. I thought having set the tokenSrc in [line 79 ](https://github.com/digitalocean/terraform-provider-digitalocean/pull/1016/files#diff-698d451af8f49ff452e38c9ce233bcdd7af4278bbb353fa01c3a1abedfb59a19R79) would suffice. Without maintaing the transport, a nil error would occur in the godo client [here](https://github.com/digitalocean/terraform-provider-digitalocean/pull/1016/files#diff-698d451af8f49ff452e38c9ce233bcdd7af4278bbb353fa01c3a1abedfb59a19R80).


I believe the `source.Source `gets lost when the transport gets set here [line 89](https://github.com/digitalocean/terraform-provider-digitalocean/pull/1016/files#diff-698d451af8f49ff452e38c9ce233bcdd7af4278bbb353fa01c3a1abedfb59a19R80). When this is removed, you do not need to maintain the transport in [line 91](https://github.com/digitalocean/terraform-provider-digitalocean/pull/1016/files#diff-698d451af8f49ff452e38c9ce233bcdd7af4278bbb353fa01c3a1abedfb59a19L87)

danaelhe:
I'm curious if we should make this the default behavior (maybe set `DIGITALOCEAN_HTTP_RETRY_MAX=2` as default).

danaelhe:
> I'm curious if we should make this the default behavior (maybe set `DIGITALOCEAN_HTTP_RETRY_MAX=2` as default).

After team discussion, I've made the retryable feature to be on by default, with 4 tries.

digitalocean/config/config.go
andrewsomething: With the changes that we made in godo, we should be able to remove the `client.Transport` modification here.

digitalocean/provider.go
```diff
@@ -76,7 +76,7 @@ func Provider() *schema.Provider {
 			"http_retry_max": {
 				Type:        schema.TypeInt,
 				Optional:    true,
-				DefaultFunc: schema.EnvDefaultFunc("DIGITALOCEAN_HTTP_RETRY_MAX", 0),
+				DefaultFunc: schema.EnvDefaultFunc("DIGITALOCEAN_HTTP_RETRY_MAX", 4),
```
andrewsomething, line 79: We'll need to update the docs to reflect that as well: https://github.com/digitalocean/terraform-provider-digitalocean/blame/56ef0b601261e395a597ef7395c98cd44466a94c/docs/index.md#L69

andrewsomething (changes requested):
Raising this to the top in case you missed my comment in the thread.

With the changes we've made to godo, we should be able to remove this bit:

```
client.Transport = &oauth2.Transport{
	Base:   client.Transport,
	Source: oauth2.ReuseTokenSource(nil, tokenSrc),
}
```

digitalocean/config/config.go
```diff
@@ -76,30 +75,37 @@ func (c *Config) Client() (*CombinedConfig, error) {
...
 	if c.HTTPRetryMax > 0 {
-		retryableClient := retryablehttp.NewClient()
-		retryableClient.RetryMax = c.HTTPRetryMax
-		retryableClient.RetryWaitMin = time.Duration(c.HTTPRetryWaitMin * float64(time.Second))
-		retryableClient.RetryWaitMax = time.Duration(c.HTTPRetryWaitMax * float64(time.Second))
+		retryConfig := godo.RetryConfig{
+			RetryMax:     c.HTTPRetryMax,
+			RetryWaitMin: godo.PtrTo(c.HTTPRetryWaitMin),
+			RetryWaitMax: godo.PtrTo(c.HTTPRetryWaitMax),
+			Logger:       log.New(os.Stderr, "", log.LstdFlags),
```
andrewsomething, line 87:
```suggestion
			Logger:      log.Default(),
```

With some additional testing, it looks like `log.Default()` will work a bit better in this case. With my original suggestion, it was still making it to the debug logs, but for some reason it was wrapped in an "unexpected data" warning.

```
2023-08-17T15:51:37.311-0400 [WARN]  unexpected data: registry.terraform.io/digitalocean/digitalocean:stderr="2023/08/17 15:51:37 [DEBUG] GET https://api.digitalocean.com/v2/domains/example-c4a4a3d0-3d1d-11ee-9746-43ce1079c844.com/records/1694880375"
```

danaelhe:
> Raising this to the top in case you missed my comment in the thread.
> 
> With the changes we've made to godo, we should be able to remove this bit:
> 
> ```
> client.Transport = &oauth2.Transport{
> 	Base:   client.Transport,
> 	Source: oauth2.ReuseTokenSource(nil, tokenSrc),
> }
> ```

oh yes thank you! Missed this somehow.

andrewsomething (approved):
:tada:  LGTM!

PR #1024, opened 2023-08-29
fix(docs): make it clear that volume name has to start with a letter

ahasna:
In the DigitalOcean [API docs](https://docs.digitalocean.com/reference/api/api-reference/#operation/volumes_create) for creating a volume, it is stated that the volume must start wit a letter which is missing from the provider docs.

```text
name required | string 

A human-readable name for the block storage volume.
Must be lowercase and be composed only of numbers, letters and "-", up to a limit of 64 characters. 
The name must begin with a letter.
```


andrewsomething (approved):
:+1:  Nice catch. Thanks!

PR #1027, opened 2023-09-08
database_user: Prevent creating multiple users for the same cluster in parallel.

andrewsomething:
This PR resolves an issue where creating multiple users for a MongoDB cluster at once can return a 500-level response by preventing parallel creation or deletion of users for the same cluster using a mutex.

(Annoyingly using `for_each` is not supported in the acceptance test framework. https://github.com/hashicorp/terraform-plugin-sdk/issues/536)


danaelhe (approved):
kinda odd `for_each` isn't supported in test framework 🤷‍♀️  but looks good 👍

PR #1036, opened 2023-09-27
Use terrafmt on docs directory.

andrewsomething:
Went to use `make terrafmt` on some new docs I'm writing and found we aren't pointing it at the docs directory in the makefile target. This updates the makefile to run it against the docs and fixes up existing formatting issues.


danaelhe (approved):
Amazing!

PR #1043, opened 2023-10-05
Parse nil region for global lb

asaha2:
Global lb creates are crashing on read after creates as it tries to parse nil `region` object: global lb doesn't have an associated region.


andrewsomething (approved):
:+1:  lgtm

PR #1045, opened 2023-10-06
Database: Clarify Docs for Referencing DB Replicas

danaelhe:
Clarify that when referencing `digitalocean_database_replica`, the `uuid` must be used instead of `id`


andrewsomething (approved):
:+1:  Good one

PR #1048, opened 2023-10-11
Rework deployment logic

### LINKED ISSUE #1047: FAILED_DEPLOY jobs cause infinite polling on app deployment
# Bug Report

---

## Describe the bug
I have an app on  app platform that consists of some jobs and a service which I deploy with terraform. I recently added a job with kind `"FAILED_DEPLOY"` which is intended to run after my service has failed deployment.

Since adding this, I've noticed that apply runs fr the `digitalocean_app` resource will hit the 30 minute timeout. I had a quick browse of the code for that resource and suspect that the polling does not take into account that in a successful deploy, the `FAILED_DEPLOY` step should never be ran. I think that it is waiting to see that step as complete when it never is. [Here](https://github.com/digitalocean/terraform-provider-digitalocean/blob/main/digitalocean/app/resource_app.go#L224) is the part of the code I'm referring to.

### Affected Resource(s)
- digitalocean_app

### Expected Behavior
My deployment should succeed

### Actual Behavior
The deployment succeeds in the digital ocean user interface, but the terraform apply hits the timeout.

### Steps to Reproduce

**Terraform version**
v1.5.7

**Debug Output**

**Panic Output**

## Additional context

**Important Factoids**

**References**

davidsbond:
This commit modifies the logic used by the `digitalocean_app` resource to check each step for a success status rather than using the counts of steps. The step counting method fails to work with jobs of kind `FAILED_DEPLOY` as these will always be pending, meaning that any app deployed with a job of this type will always time out even when it is successful.

Closes #1047

### DIFF
--- a/digitalocean/app/resource_app.go
+++ b/digitalocean/app/resource_app.go
@@ -221,7 +221,14 @@ func waitForAppDeployment(client *godo.Client, id string, timeout time.Duration)
 				return fmt.Errorf("Error trying to read app deployment state: %s", err)
 			}
 
-			allSuccessful := deployment.Progress.SuccessSteps == deployment.Progress.TotalSteps
+			allSuccessful := true
+			for _, step := range deployment.Progress.Steps {
+				if step.Status != godo.DeploymentProgressStepStatus_Success {
+					allSuccessful = false
+					break
+				}
+			}
+
 			if allSuccessful {
 				ticker.Stop()
 				return nil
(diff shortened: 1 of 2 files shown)


andrewsomething (approved):
Thanks so much for flagging this issue and providing a great solution! :pray:  

I did push one additional commit adding a `FAILED_DEPLOY` job to the existing acceptance test for app platform jobs.

davidsbond:
Thanks very much @andrewsomething! Been really easy using app platform, DO in general and this terraform provider. Really good work from y'all and happy to contribute!

PR #1056, opened 2023-10-25
Add support for Kafka Topic User ACL management

dweinshenker:
**What**

- Add support for kafka topic user acls based on godo updates (https://github.com/digitalocean/godo/pull/633)

**Testing**

Ran`TestAccDigitalOceanDatabaseUser_KafkaACLs,TestAccDigitalOceanDatabaseUser_Basic,TestAccDigitalOceanDatabaseUser_MongoDB,TestAccDigitalOceanDatabaseUser_MongoDBMultiUser,TestAccDigitalOceanDatabaseUser_MySQLAuth` acceptance tests


andrewsomething (approved):
:+1:  Looks great! Sorry for the delayed review.

digitalocean/database/resource_database_user.go
```diff
@@ -67,6 +78,32 @@ func ResourceDigitalOceanDatabaseUser() *schema.Resource {
...
+				Computed: true,
+			},
+			"topic": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.NoZeroValues,
+			},
+			"permission": {
+				Type:     schema.TypeString,
+				Required: true,
```
andrewsomething, line 95: On second look, I'm realizing that updates are not actually supported. Combined with the fact that we're not setting state on read as `GetUser` responses not including `settings`, we mislead users into thinking that an update is successfully applied.  So we should add `ForceNew` to these to make it clear that in-place updates are not supported. 

```suggestion
				Type:         schema.TypeString,
				Required:     true,
				ForceNew:     true,
				ValidateFunc: validation.NoZeroValues,
			},
			"permission": {
				Type:     schema.TypeString,
				Required: true,
				ForceNew: true,
```

andrewsomething (changes requested):

digitalocean/database/resource_database_user.go
dweinshenker: I agree with your suggestion, that makes it more clear that these changes are not allowed in-place. Thanks for the review @andrewsomething !

andrewsomething (approved):
:+1:

PR #1067, opened 2023-11-03
load balancers: ignore 404 on delete.

andrewsomething:
As brought up in this discussion thread, there are cases where Kubernetes will delete a load balancer itself. This can cause an error when Terraform attempts to do so. We can safely ignore a 404 error when attempting to delete as that just indicates that the load balancer has already been destroyed. 

https://github.com/digitalocean/terraform-provider-digitalocean/discussions/1065


danaelhe (approved):
👍 lgtm

Issue #901, opened 2022-11-18
When creating volumes, "region" is not being converted to lowercase letters and causes an error during `terraform apply`


nixc0:
# Bug Report

---

## Describe the bug

When using uppercase letters in region (example NYC3 or AMS1) when creating a new volume causes `terraform apply` to fail

### Affected Resource(s)
- digitalocean_volume

If this issue appears to affect multiple resources, it may be an issue with
Terraform's core, so please mention this. -->

### Expected Behavior

Region should have been converted to lowercase nyc3 and volume created. This is the behavior that happens when I create a Droplet using the region NYC3. 
### Actual Behavior

When running `terraform apply` the following error is produced. 
 Error: Error creating Volume: POST https://api.digitalocean.com/v2/volumes: 403 (request "51693f34-b786-44e3-8fe1-8ac3dcbe7e1a") failed to create volume: volume allocation operations not allowed in this region at this time
### Steps to Reproduce

**Terraform Configuration Files**
Here is an example of the Droplet I am creating at the same time using "NYC3" as the region instead of "nyc3". This works as expected.

```
terraform {
  required_providers {
    digitalocean = {
      source = "digitalocean/digitalocean"
      version = "~> 2.0"
    }
  }
}
variable "do_token" {}

provider "digitalocean" {
  token = var.do_token
}
resource "digitalocean_droplet" "www-1" {
    image = "ubuntu-20-04-x64"
    name = "www-1"
    region = "NYC3"
    size = "s-1vcpu-1gb"
}

```

During the same plan and apply, below is the configuration file for the volume I am creating with "NYC3" as the region instead of "nyc3".
```
resource "digitalocean_volume" "foobar" {
  region                  = "NYC3"
  name                    = "baz"
  size                    = 100
  initial_filesystem_type = "ext4"
  description             = "an example volume"
}
```
**Terraform version**

I replicated this issue on 1.3.4 & 1.3.5

**Debug Output**

https://gist.github.com/nixc0/0864906078831e9a65bf6ce7c8f16f31
**Panic Output**

## Additional context

If I'm not mistaken, this line https://github.com/digitalocean/terraform-provider-digitalocean/blob/2a413c8b36d0f875c19f4cd1bafb26a769929417/digitalocean/resource_digitalocean_volume.go#L32 should be converting this "NYC3" to "nyc3" in order to work with the all lowercase requirements of the DO API.

andrewsomething:
[...]

PR #1074, opened 2023-11-10
Fixing Case Handling for Volume Resource

T-jegou:
Hello, this Pull Request is intended to address the issue https://github.com/digitalocean/terraform-provider-digitalocean/issues/901. However, I've observed that the API V2 accepts, for certain types of resources, the region to be provided in uppercase.

In my Pull Request, I've implemented a solution for the volume resource, involving converting the region string to lowercase when it's added to the `opts` structure in the `resourceDigitalOceanVolumeCreate` function, to ensure that the provider calls godo with the region field in lowercase.

However, I find that this approach lacks consistency since the API handles the case of a region field in uppercase in the context of the droplet resource (and probably others).

Perhaps the fix should be applied directly to the API.

Nevertheless, strictly adhering to the API documentation, the region slugs are only in lowercase. In the case where a region is indicated in uppercase, either in the resource schema, we apply a condition to ensure the region is only in lowercase, and the documentation should state "lowercase only" for region slugs, or we accept and handle the formatting to lowercase for storage in the state (with StateFunc) and for sending data to godo in `opts`.


andrewsomething (approved):
Hey @T-jegou! Thanks for the PR. I agree that it would be better if the API handled this consistently, and we can share that feedback internally.  Though I think this change is helpful for Terraform users in the meantime.

Issue #1086, opened 2023-11-30
`digitalocean_cdn` resource with custom domain stopped working


itssimon:
# Bug Report

## Describe the bug
Using a `digitalocean_cdn` resource with a custom domain within a Terraform configuration makes all `terraform` commands fail with the following error:

Error: GET https://api.digitalocean.com/v2/certificates/needs-cloudflare-cert: 404 (request "00000000-0000-0000-0000-000000000000") certificate not found

### Affected Resource(s)
- digitalocean_cdn

### Expected Behavior
No errors

### Actual Behavior
Error

### Steps to Reproduce
1. `terraform import terraform_cdn.example 00000000-0000-0000-0000-000000000000`

**Terraform Configuration Files**
```
resource "digitalocean_cdn" "example" {
  origin           = digitalocean_spaces_bucket.example.bucket_domain_name
  custom_domain    = "assets.example.com"
}
```

**Terraform version**
```
Terraform v1.5.4
on darwin_arm64
+ provider registry.terraform.io/digitalocean/digitalocean v2.32.0
```

## Additional context
I've been in touch with the DigitalOcean support team and they have indicated that there has recently been an upgrade to Spaces that enabled users to configure a custom domain for the CDN (with DNS managed elsewhere) without having to provide their own certificate.

I suspect this upgrade has broken the Terraform provider, as now they set the `certificate_id` to `needs-cloudflare-cert`, which doesn't seem to actually be a valid ID.

danaelhe:
Thanks for bringing this to our attention. We've connected with the DO Support team and the DO engineering team that owns this and working this through with them.

What I _think_ is happening is that in the terraform provider, we've deprecated `certificate_id` attribute in favor of `certificate_name` for `digitalocean_cdn` because when the certificate type is lets_encrypt, the certificate ID will change on renewal (after 3 months).

What I think is the problem in the Terraform provider is when `certificate_name` is passed in, we use the name to look up the id and then set that id to `certificate_id`.....hence the `certificate_id` will not be `needs-cloudflare-cert` [here](https://github.com/digitalocean/terraform-provider-digitalocean/blob/main/digitalocean/cdn/resource_cdn.go#L142-L151):
```
	if name, nameOk := d.GetOk("certificate_name"); nameOk {
		certName := name.(string)
		if certName != "" {
			cert, err := certificate.FindCertificateByName(client, certName)
			if err != nil {
				return diag.FromErr(err)
			}

[...]

PR #1089, opened 2023-12-04
CDN: Add Support for "needs-cloudflare-cert"

danaelhe:
Address #1086 

In order to have Cloudflare generate a certificate, `certificate_id` must be set to the value "needs-cloudflare-cert" in our API.

In the Terraform provider `certificate_name` is passed in instead of a `certificate_id` _(we've deprecated certificate_id attribute in favor of certificate_name for digitalocean_cdn because when the certificate type is lets_encrypt, the certificate ID will change on renewal (after 3 months)_), we use the name to look up the id and then set that id to `certificate_id`.....hence the `certificate_id` will not be `needs-cloudflare-cert` [here](https://github.com/digitalocean/terraform-provider-digitalocean/blob/main/digitalocean/cdn/resource_cdn.go#L142-L151):
```
	if name, nameOk := d.GetOk("certificate_name"); nameOk {
		certName := name.(string)
		if certName != "" {
			cert, err := certificate.FindCertificateByName(client, certName)
			if err != nil {
				return diag.FromErr(err)
			}

			cdnRequest.CertificateID = cert.ID
		}
```

Added logic to check if `needs-cloudflare-cert`is passed into `certificate_name` to set that as the `certificate_id`

### DIFF
--- a/digitalocean/cdn/resource_cdn_test.go
+++ b/digitalocean/cdn/resource_cdn_test.go
@@ -39,6 +39,33 @@ func TestAccDigitalOceanCDN_Create(t *testing.T) {
 	})
 }
 
+func TestAccDigitalOceanCDN_CreateWithNeedsCloudflareCert(t *testing.T) {
+
+	bucketName := generateBucketName()
+	cdnCreateConfig := fmt.Sprintf(testAccCheckDigitalOceanCDNConfig_CreateWithNeedsCloudflareCert, bucketName)
+
+	expectedOrigin := bucketName + originSuffix
+	expectedTTL := "3600"
+
+	resource.ParallelTest(t, resource.TestCase{
+		PreCheck:          func() { acceptance.TestAccPreCheck(t) },
+		ProviderFactories: acceptance.TestAccProviderFactories,
+		CheckDestroy:      testAccCheckDigitalOceanCDNDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: cdnCreateConfig,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckDigitalOceanCDNExists("digitalocean_cdn.foobar"),
+					resource.TestCheckResourceAttr(
+						"digitalocean_cdn.foobar", "origin", expectedOrigin),
+					resource.TestCheckResourceAttr("digitalocean_cdn.foobar", "ttl", expectedTTL),
+					resource.TestCheckResourceAttr("digitalocean_cdn.foobar", "certificate_name", "need-cloudflare-cert"),
+				),
+			},
+		},
+	})
+}
+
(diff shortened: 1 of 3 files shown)


danaelhe:
Hey @itssimon - could you give this a try?

andrewsomething (approved):
:+1:  Think this is a great workaround for the moment.

digitalocean/cdn/resource_cdn_test.go
```diff
@@ -39,6 +39,33 @@ func TestAccDigitalOceanCDN_Create(t *testing.T) {
...
+		CheckDestroy:      testAccCheckDigitalOceanCDNDestroy,
+		Steps: []resource.TestStep{
+			{
+				Config: cdnCreateConfig,
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckDigitalOceanCDNExists("digitalocean_cdn.foobar"),
+					resource.TestCheckResourceAttr(
+						"digitalocean_cdn.foobar", "origin", expectedOrigin),
+					resource.TestCheckResourceAttr("digitalocean_cdn.foobar", "ttl", expectedTTL),
+					resource.TestCheckResourceAttr("digitalocean_cdn.foobar", "certificate_name", "need-cloudflare-cert"),
```
andrewsomething, line 62:
```suggestion
					resource.TestCheckResourceAttr("digitalocean_cdn.foobar", "certificate_name", "needs-cloudflare-cert"),
```

One tiny change needed here

andrewsomething (changes requested):

digitalocean/cdn/resource_cdn_test.go
andrewsomething: According to the const, it is `needs-cloudflare-cert`, right?

digitalocean/cdn/resource_cdn_test.go
```diff
@@ -207,6 +234,18 @@ resource "digitalocean_cdn" "foobar" {
...
+const testAccCheckDigitalOceanCDNConfig_CreateWithNeedCloudflareCert = `
+resource "digitalocean_spaces_bucket" "bucket" {
+  name   = "%s"
+  region = "ams3"
+  acl    = "public-read"
+}
+
+resource "digitalocean_cdn" "foobar" {
+  origin           = digitalocean_spaces_bucket.bucket.bucket_domain_name
+  certificate_name = "need-cloudflare-cert"
```
andrewsomething, line 246:
```suggestion
  certificate_name = "needs-cloudflare-cert"
```

digitalocean/cdn/resource_cdn_test.go
danaelhe: oh yes 🤦‍♀️

andrewsomething (approved):
:rocket:

PR #1095, opened 2023-12-18
cdn: handle 'needs-cloudflare-cert' case in read func.

andrewsomething:
This addresses the continued issue called out in https://github.com/digitalocean/terraform-provider-digitalocean/issues/1086. We need to handle the 'needs-cloudflare-cert' case in read func as well as create. 

The acceptance test did not catch this. In order to fully test the flow, we would need an active domain. We can do this in our own testing, but the acceptance test suite must be usable by external contributors and not depend on external resources. So the updated test is skipped if `DO_TEST_SUBDOMAIN` is not set to an active DO manage sub domain.


danaelhe (approved):
good catch!

Issue #1104, opened 2024-01-26
Disabling ipv6 on a droplet does not actually disable it


danktec:
# Bug Report
Enabling ipv6 works fine. Disabling says that it happened but it did not, rerunning yields the same result forever.

### Affected Resource(s)
- digitalocean_droplet

### Expected Behavior
When setting 
`  ipv6 =  false`
on the digitalocean_droplet resource, if ipv6 was previously enabled it should now be disabled.

### Actual Behavior
An update-in-place is detected
`      ~ ipv6                 = true -> false`
but the API call does not make the change.

Re-running apply yields the same result forever.

### Steps to Reproduce
enable ipv6, run apply
disable ipv6, run apply (says it works but does not make any changes to the cloud state)

**Terraform version**
```
Terraform v1.7.1
on darwin_arm64
+ provider registry.terraform.io/digitalocean/digitalocean v2.34.1
```

danktec:
I don't think the API supports disabling v6 so perhaps the provider just needs to register this change as a "delete/replace" on the resource.

danktec:
Fixed pushed here https://github.com/digitalocean/terraform-provider-digitalocean/pull/1105

andrewsomething:
:pray:  Thanks @danktec! 

Fixed via https://github.com/digitalocean/terraform-provider-digitalocean/pull/1105

PR #1105, opened 2024-01-26
Error out with a message if the user tries to disable IPv6 on a droplet

danktec:
It appears the API does not support disablement of IPv6, only enablement. If ipv6 is set to false after being set to true then nothing changes on the cloud provider and the diff keeps happening forever #1104

This update uses forces a replacement if the user is trying to disable ipv6


andrewsomething (approved):
:+1:  LGTM!

Thanks for flagging this issue and the fix!

Issue #1117, opened 2024-03-01
Bug: "produced an unexpected new value: Root resource was present, but now absent." + Question: Adding an "App Platform" into a "Project" 


mwyld:
# Bug Report

I'm new to terraform and digital ocean but is there a reason why you can't move an app platform into a project? [digitalocean_project](https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/resources/project)

I'm also getting the following message when trying to add a domain or space into a project,

```
const spaces_bucket = new SpacesBucket(this, `my-spaces-bucket`, spaces-bucket_config)

new ProjectResources(this, `my-project-resource`, {
          project: `my-project`,
          resources: [
            Token.asString(spaces_bucket.urn),
          ],
});
```

                                │ Error: Provider produced inconsistent result after apply
                                │
                                │ When applying changes to
                                │ digitalocean_project_resources.my-app-project-resource,
                                │ provider "provider[\"registry.terraform.io/digitalocean/digitalocean\"]"
                                │ produced an unexpected new value: Root resource was present, but now absent.
                                │
                                │ This is a bug in the provider, which should be reported in the provider's own
                                │ issue tracker.

andrewsomething:
Hi @mwyld,

Is that Terraform CDK? In `project: my-project`, what is the value of `my-project`? It should be the ID of the project. Here's a working example in HCL that creates a new Space and puts into my existing project named "Test":

```hcl
data "digitalocean_project" "test" {
  name = "Test"
}

resource "digitalocean_spaces_bucket" "foo" {
  name   = "new-test-bucket-with-a-unique-name"
  region = "nyc3"
}

resource "digitalocean_project_resources" "foo" {
  project = data.digitalocean_project.test.id
  resources = [
    digitalocean_spaces_bucket.foo.urn
  ]
}
```

> I'm new to terraform and digital ocean but is there a reason why you can't move an app platform into a project?

You can place an app into a project. It looks like we're missing it in the documentation. Thanks for pointing that out!

PR #1120, opened 2024-03-01
docs: apps can be added to projects.

andrewsomething:
As mentioned in https://github.com/digitalocean/terraform-provider-digitalocean/issues/1117, the docs don't mention that apps can be added to projects when they can:

https://docs.digitalocean.com/reference/api/api-reference/#tag/Project-Resources

(Also went all in on plural)


danaelhe (approved):
👍

PR #1127, opened 2024-03-14
Load Balancer: Deprecate "algorithm" attribute

danaelhe:
This field has been deprecated. You can no longer specify an algorithm for load balancers.

https://docs.digitalocean.com/reference/api/api-reference/#operation/loadBalancers_create


andrewsomething (approved):
:+1:  LGTM

Issue #1121, opened 2024-03-05
digitalocean_database_cluster kafka missing attributes


SkeLLLa:
# Bug Report

---

## Describe the bug

Resource `digitalocean_database_cluster` for kafka cluster missing many attributes like:
```
"port": 0,
"private_uri": "",
"uri": "",
```

### Expected Behavior

All resources include port, uri and private_uri should be present

### Actual Behavior

Port attribute is wrong, other attributes missing

### Steps to Reproduce

1. Use example from docs:
```
resource "digitalocean_database_cluster" "kafka-example" {
  name       = "example-kafka-cluster"
  engine     = "kafka"
  version    = "3.5"
  size       = "db-s-2vcpu-2gb"
  region     = "nyc1"
  node_count = 3
}
```

2. See terraform state.

danaelhe:
Thank you for bringing this to our attention. 

For Kafka, we do not expect those attributes and the provider shouldn't have returned them at all. I've created a ticket internally for our team to ensure the provider only returns the attributes expected from Kafka. I'll keep this issue open until it is addressed.

SkeLLLa:
If the provider doesn't return them, how, for example port will be known? At least it should be returned, otherwise it will require hardconding it on a client side and if there will be changed in DO kafka deployment (e.g. "private port" will change or so) it will introduce a risk and will require extra testing.

PR #1131, opened 2024-03-20
Databases: Set Kafka Connection Attributes (port, uri, private_uri)

danaelhe:
Addresses #1121

Database team is working on a ticket to address this in the API layer, but in the meantime this sets the port, uri, and private uri at the terraform layer so users can easily collect connection details for their kafka clusters.

Default port will be the public SASL port, 25073. This is consistent with the UI.

### DIFF
--- a/digitalocean/database/resource_database_cluster.go
+++ b/digitalocean/database/resource_database_cluster.go
@@ -22,6 +22,10 @@ const (
 	mongoDBEngineSlug = "mongodb"
 	mysqlDBEngineSlug = "mysql"
 	redisDBEngineSlug = "redis"
+	kafkaDBEngineSlug = "kafka"
+
+	kafkaPublicSSLPort  = 25073
+	kafkaPrivateSSLPort = 25080
 )
 
 func ResourceDigitalOceanDatabaseCluster() *schema.Resource {
@@ -307,7 +311,7 @@ func resourceDigitalOceanDatabaseClusterCreate(ctx context.Context, d *schema.Re
 
 	// MongoDB clusters only return the password in response to the initial POST.
 	// We need to set it here before any subsequent GETs.
-	if database.EngineSlug == mongoDBEngineSlug {
+	if database.EngineSlug == mongoDBEngineSlug || database.EngineSlug == kafkaDBEngineSlug {
 		err = setDatabaseConnectionInfo(database, d)
 		if err != nil {
 			return diag.Errorf("Error setting connection info for database cluster: %s", err)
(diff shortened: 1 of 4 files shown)


digitalocean/database/resource_database_cluster.go
```diff
@@ -609,8 +613,12 @@ func flattenMaintWindowOpts(opts godo.DatabaseMaintenanceWindow) []map[string]in
 func setDatabaseConnectionInfo(database *godo.Database, d *schema.ResourceData) error {
 	if database.Connection != nil {
 		d.Set("host", database.Connection.Host)
-		d.Set("port", database.Connection.Port)
-		d.Set("uri", database.Connection.URI)
+		if database.EngineSlug == kafkaDBEngineSlug {
+			// default for kafka will be Public SASL port, consistent with UI
+			d.Set("port", kafkaPublicSSLPort)
```
andrewsomething, line 618: Are the ports guaranteed to be stable?  Maybe we should grab it from the response? Something like:

```go
const kafkaSASL = "sasl"
if database.Connection.ApplicationPorts != nil {
	if saslPort, ok := database.Connection.ApplicationPorts[kafkaSASL]; ok {
		d.Set("port", saslPort)
	}
}
```

digitalocean/database/resource_database_cluster.go
```diff
@@ -661,6 +675,16 @@ func buildMongoDBConnectionURI(conn *godo.DatabaseConnection, d *schema.Resource
 	return uri.String(), nil
 }
 
+// DO API returns null values for uri and private uri for Kafka clusters.
+// buildKafkaConnectionURI sets the uri and private uri using connection's host.
+func buildKafkaConnectionURI(conn *godo.DatabaseConnection, port int) string {
+	host := conn.Host
+
+	uri := fmt.Sprintf("%s:%d", host, port)
```
andrewsomething, line 683: Lets make sure that this is consistent with what the DBaaS team will eventually return from the API. Should there be a scheme? The other engines all seem to use the format `scheme://user:pass@host:port`

danaelhe: > Lets make sure that this is consistent with what the DBaaS team will eventually return from the API. Should there be a scheme? The other engines all seem to use the format `scheme://user:pass@host:port`

Converting this PR to a draft as dbaas said they might actually get to it this week or next!

andrewsomething: Sounds good!

loosla:
Thanks for the implementation! 
I’ve created a different PR since the initial code underwent some refactoring, along with changes from the Database team regarding Kafka (like the port). A new PR inspired by this one is here: https://github.com/digitalocean/terraform-provider-digitalocean/pull/1246

PR #1133, opened 2024-04-03
Add config options for global load balancers

asaha2:


asaha2:
cc @andrewsomething for reviews 👀

digitalocean/loadbalancer/resource_loadbalancer.go
```diff
@@ -432,13 +433,110 @@ func resourceDigitalOceanLoadBalancerV0() *schema.Resource {
...
+							Optional:    true,
+							Description: "list of domain verification errors",
+						},
+						"ssl_validation_error_reasons": {
+							Type:        schema.TypeList,
+							Computed:    true,
+							Elem:        &schema.Schema{Type: schema.TypeString},
+							Optional:    true,
+							Description: "list of domain SSL validation errors",
+						},
```
asaha2, line 484: The `*_error_reasons` fields are read-only fields and non-configurable by the user: should these be added as resource flags here?

digitalocean/loadbalancer/resource_loadbalancer.go
```diff
@@ -204,9 +204,10 @@ func resourceDigitalOceanLoadBalancerV0() *schema.Resource {
 			},
 
 			"forwarding_rule": {
-				Type:     schema.TypeSet,
-				Required: true,
-				MinItems: 1,
+				Type:          schema.TypeSet,
+				Optional:      true,
+				ConflictsWith: []string{"glb_settings"},
+				MinItems:      1,
```
asaha2, line 210: Forwarding rules aren't supported for GLB, instead `glb_settings` embed target rules as per requirement. Having `forwarding_rule` set as `required` is conflicting with the integration tests, hence changing this to optional. Should we update the resources readme text to reflect this change as well?

digitalocean/loadbalancer/resource_loadbalancer.go
```diff
@@ -432,13 +433,110 @@ func resourceDigitalOceanLoadBalancerV0() *schema.Resource {
...
+							Optional:    true,
+							Default:     false,
+							Description: "flag indicating if domain is managed by DigitalOcean",
+						},
+						"certificate_id": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							ValidateFunc: validation.NoZeroValues,
+							Description:  "certificate ID for TLS handshaking",
+						},
```
andrewsomething, line 470: Does this experience the same issue as with the certificates used for forwarding rules?

We had a lot of pain around Let's Encrypt certificates as their IDs change when they are renewed. This causes Terraform to prompt the user to reconcile state and attempt to reapply the old ID. So we had to migrate users from `certificate_id` to `certificate_name` being the source of truth. Check out how we approach it in `expandForwardingRules`.

https://github.com/digitalocean/terraform-provider-digitalocean/blob/007c65e4893779497d2235816a78ccc067605b84/digitalocean/loadbalancer/loadbalancer.go#L103-L113

digitalocean/loadbalancer/resource_loadbalancer.go
asaha2: Agreed 💯 taking note of it for later update once we GA.

digitalocean/loadbalancer/resource_loadbalancer.go
asaha2: Managing certificates for domain would be identical to how we do the same for forwarding rules currently. cc @bbassingthwaite to correct if I'm wrong.

andrewsomething (approved):
:rocket:

PR #1144, opened 2024-04-23
chore: remove go.mod replace for old unused pkg

yordis:
## Context

I am working on [Crossplane DigitalOcean No Fork Upjet Provider](https://github.com/straw-hat-team/provider-digitalocean/pull/4) that requires me to add this package as a dependency. I noticed this old package being pinned to an old version, so I had to deal with adding the `replace` statement to then realize that this package does not seem to be used.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -70,6 +70,4 @@ require (
 
 replace git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999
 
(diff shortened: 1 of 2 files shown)


andrewsomething (approved):
:+1:  Thanks for the PR @yordis!

I did push an additional commit after running `go mod vendor` to update `vendor/modules.txt`

yordis:
@andrewsomething thank you so much! Is there any opportunity to cut a release so that I can use it in the Crossplane Provider? I pinned the deps to my own repo but I would love to remove that

andrewsomething:
@yordis Sure thing! I've just released [v2.38.0](https://github.com/digitalocean/terraform-provider-digitalocean/releases/tag/v2.38.0) with this change.

PR #1145, opened 2024-04-23
chore: fix header of changelog

yordis:
## Context

Just making the file aligned with previous format (or maybe we should change the previous format)


andrewsomething (approved):
:+1: Thanks!

PR #1160, opened 2024-05-17
[VPC-3361] Add VPC Peering resource

apinonformoso:
VPC Peering is a closed beta feature and not available for public use.

[Testing doc](https://docs.google.com/document/d/1FmoRbDxcZiZGDkZ2Tllcgqbls2ytREEJN6jnRNWf4mQ/edit?usp=sharing)


digitalocean/vpcpeering/resource_vpcpeering.go
```diff
@@ -0,0 +1,204 @@
...
+	for _, id := range val {
+		vpcID := id
+		if _, ok := vpcIDMap[vpcID]; ok {
+			return fmt.Errorf("'vpc_ids' must contain unique VPC IDs, duplicate found: %s", vpcID)
+		}
+		vpcIDMap[vpcID] = struct{}{}
+	}
+
+	return nil
+}
```
ddatta-do, line 184: not sure if we need this here since we already have serverside check..

digitalocean/vpcpeering/resource_vpcpeering.go
```diff
@@ -0,0 +1,204 @@
...
+			return fmt.Errorf("error checking VPC Peering status: %s", err)
+		}
+
+		if time.Now().After(deadline) {
+			return fmt.Errorf("timed out waiting for VPC Peering to be deleted")
+		}
+
+		time.Sleep(2 * time.Second)
+	}
+}
```
ddatta-do, line 204: probably better to do it this way https://github.com/digitalocean/terraform-provider-digitalocean/blob/main/digitalocean/loadbalancer/resource_loadbalancer.go#L698

digitalocean/vpcpeering/datasource_vpcpeering.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+	if len(results) == 1 {
+		return results[0], nil
+	} else if len(results) == 0 {
+		return nil, fmt.Errorf("no VPC Peerings found with name %s", name)
+	}
+
+	return nil, fmt.Errorf("too many VPC Peerings found with name %s (found %d, expected 1)", name, len(results))
+}
```
ddatta-do, line 135: just curious .. why do we need to expose this via tf ?

digitalocean/vpcpeering/resource_vpcpeering.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+
+	log.Printf("[DEBUG] VPC Peering create request: %#v", vpcPeeringRequest)
+	vpcPeering, _, err := client.VPCs.CreateVPCPeering(context.Background(), vpcPeeringRequest)
+	if err != nil {
+		return diag.Errorf("Error creating VPC Peering: %s", err)
+	}
+
+	d.SetId(vpcPeering.ID)
+	log.Printf("[INFO] VPC Peering created, ID: %s", d.Id())
```
ddatta-do, line 95: probably need to wait before proceeding similar to Delete

digitalocean/vpcpeering/resource_vpcpeering.go
apinonformoso: I added this to spot this issue before the request with wrong parameters is done, but probably not necessary. As you mentioned it is already handled serverside.

apinonformoso: I will remove this validation

ddatta-do: tbh... I am also not sure .. I'll let other's decide

apinonformoso: I have validated and imo this check isn't necessary. Since `TypeSet` can't have a `ValidateFunc` we will get this error from serverside. I mean, for other fields where we have the `ValidateFunc` it is for running `terraform validate` and return the errors before making an apply, but I think this can't be done for `TypeSet` or `TypeList`

digitalocean/vpcpeering/datasource_vpcpeering.go
apinonformoso: In case a user has already created a vpc peering, it can be retrieved either by id or name

digitalocean/vpcpeering/sweep.go
```diff
@@ -0,0 +1,54 @@
...
+	client := meta.(*config.CombinedConfig).GodoClient()
+
+	opt := &godo.ListOptions{PerPage: 200}
+	vpcPeerings, _, err := client.VPCs.ListVPCPeerings(context.Background(), opt)
+	if err != nil {
+		return err
+	}
+
+	for _, v := range vpcPeerings {
+		if strings.HasPrefix(v.Name, sweep.TestNamePrefix) {
```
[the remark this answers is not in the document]
ddatta-do: or is it used only for testing ?

digitalocean/vpcpeering/resource_vpcpeering.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	return resourceDigitalOceanVPCPeeringRead(ctx, d, meta)
+}
+
+func resourceDigitalOceanVPCPeeringDelete(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	client := meta.(*config.CombinedConfig).GodoClient()
+	vpcPeeringID := d.Id()
+
+	err := retry.RetryContext(ctx, d.Timeout(schema.TimeoutDelete), func() *retry.RetryError {
```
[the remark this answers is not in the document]
apinonformoso: No particular reason, I implemented the CRUD by looking at how `vpc` resource was implemented but I can add the retry logic

digitalocean/vpcpeering/sweep.go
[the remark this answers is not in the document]
apinonformoso: Yes, this is correct. `sweep.go` is used to cleanup all resources created during testing.

digitalocean/vpcpeering/sweep.go
```diff
@@ -0,0 +1,54 @@
...
+			log.Printf("[DEBUG] Destroying VPC Peering %s", v.Name)
+			resp, err := client.VPCs.DeleteVPCPeering(context.Background(), v.ID)
```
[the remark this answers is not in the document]
apinonformoso: Yes I think I need to add a wait here, also looking at the `Dependencies` field on the init function I saw it says `Dependencies list the const names of other Sweeper functions that must be ran prior to running this Sweeper` so I should remove the `digitalocean_vpc` dependency and add `digitalocean_vpcpeering` to the vpc resource sweep.go

digitalocean/vpcpeering/resource_vpcpeering.go
```diff
@@ -0,0 +1,199 @@
...
+	vpcPeering, _, err := client.VPCs.CreateVPCPeering(context.Background(), vpcPeeringRequest)
+	if err != nil {
+		return diag.Errorf("Error creating VPC Peering: %s", err)
+	}
+
+	d.SetId(vpcPeering.ID)
+
+	log.Printf("[DEBUG] Waiting for VPC Peering (%s) to become active", d.Get("name"))
+	stateConf := &retry.StateChangeConf{
+		Delay:      10 * time.Millisecond,
```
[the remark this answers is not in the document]
apinonformoso: I added here an initial delay because if it fetches the state of the peering immediately it will get a not found error and fail the creation. I observed that 10 milliseconds is a good delay time so the peering is added to the db, but if you feel it is too tight we can go for 30 milliseconds. I think 5 seconds is a lot of unnecessary wait time that will make creation take more time than necessary due to state validation

digitalocean/vpcpeering/resource_vpcpeering.go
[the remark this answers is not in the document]
apinonformoso: added

digitalocean/vpcpeering/resource_vpcpeering.go
```diff
@@ -0,0 +1,212 @@
...
+
+		d.SetId(vpcPeering.ID)
+
+		log.Printf("[DEBUG] Waiting for VPC Peering (%s) to become active", d.Get("name"))
+		stateConf := &retry.StateChangeConf{
+			Delay:      5 * time.Second,
+			Pending:    []string{"PROVISIONING"},
+			Target:     []string{"ACTIVE"},
+			Refresh:    vpcPeeringStateRefreshFunc(client, d.Id()),
+			Timeout:    10 * time.Minute,
```
ddatta-do, line 104: I think this timeout should be less than or equal to https://github.com/digitalocean/terraform-provider-digitalocean/pull/1160/files#diff-9f6d953dd67504f834d1166d111e56b71895731f5d87fe7fcfbcb6f94a5fe373R63 this one

digitalocean/vpcpeering/resource_vpcpeering.go
```diff
@@ -0,0 +1,212 @@
...
+			}
+			return retry.NonRetryableError(fmt.Errorf("error deleting VPC Peering: %s", err))
+		}
+
+		log.Printf("[DEBUG] Waiting for VPC Peering (%s) to be deleted", d.Get("name"))
+		stateConf := &retry.StateChangeConf{
+			Pending:    []string{"DELETING"},
+			Target:     []string{http.StatusText(http.StatusNotFound)},
+			Refresh:    vpcPeeringStateRefreshFunc(client, d.Id()),
+			Timeout:    10 * time.Minute,
```
ddatta-do, line 159: same

digitalocean/vpcpeering/resource_vpcpeering.go
```diff
@@ -0,0 +1,212 @@
...
+		if err != nil {
+			if resp != nil && resp.StatusCode == http.StatusForbidden {
+				return retry.RetryableError(err)
+			}
+			return retry.NonRetryableError(fmt.Errorf("error deleting VPC Peering: %s", err))
+		}
+
+		log.Printf("[DEBUG] Waiting for VPC Peering (%s) to be deleted", d.Get("name"))
+		stateConf := &retry.StateChangeConf{
+			Pending:    []string{"DELETING"},
```
ddatta-do, line 156: probably add the delay here also ?

digitalocean/vpcpeering/resource_vpcpeering_test.go
```diff
@@ -0,0 +1,133 @@
...
+						"digitalocean_vpcpeering.foobar", "created_at"),
+					resource.TestCheckResourceAttrSet(
+						"digitalocean_vpcpeering.foobar", "status"),
+				),
+			},
+			{
+				Config: vpcPeeringUpdateConfig,
+				PreConfig: func() {
+					// Wait a few seconds for the VPC Peeing to become active
+					time.Sleep(3 * time.Second)
```
ddatta-do, line 51: probably not needed anymore ?

ddatta-do (approved):

digitalocean/vpcpeering/datasource_vpcpeering.go
```diff
@@ -0,0 +1,139 @@
...
+		vpcPeering, err := findVPCPeeringByName(vpcPeerings, name.(string))
+		if err != nil {
+			return diag.Errorf("Error retrieving VPC Peering: %s", err)
+		}
+
+		foundVPCPeering = vpcPeering
+	}
+
+	if foundVPCPeering == nil {
+		return diag.Errorf("Error retrieving VPC Peering: %s", fmt.Errorf("vpc peering not found"))
```
lfundaro-do, line 82: we could be more precise with the error here. I think you will reach this branch if neither "name" nor "id" was given, for that case, it's not a 404 but more like a bad request (401?) .

digitalocean/vpcpeering/datasource_vpcpeering.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func findVPCPeeringByName(vpcPeerings []*godo.VPCPeering, name string) (*godo.VPCPeering, error) {
+	results := make([]*godo.VPCPeering, 0)
+	for _, v := range vpcPeerings {
+		if v.Name == name {
+			results = append(results, v)
+		}
+	}
+	if len(results) == 1 {
```
lfundaro-do, line 132: would there be a case where a list of peerings will have a duplicate name ? afaik, peering name must be unique. I think you can refactor this. It doesn't seem you need a slice for this.

digitalocean/vpcpeering/resource_vpcpeering.go
```diff
@@ -0,0 +1,213 @@
...
+		VPCIDs: vpcIDsString,
+	}
+
+	log.Printf("[DEBUG] VPC Peering create request: %#v", vpcPeeringRequest)
+
+	err := retry.RetryContext(ctx, d.Timeout(schema.TimeoutCreate), func() *retry.RetryError {
+		vpcPeering, resp, err := client.VPCs.CreateVPCPeering(context.Background(), vpcPeeringRequest)
+		if err != nil {
+			if resp != nil && resp.StatusCode == http.StatusForbidden {
+				return retry.RetryableError(err)
```
lfundaro-do, line 91: if status is forbidden, are we sure we want to retry ?

danaelhe:
Super small request 🤏 Could you also change the filenames to be `vpc_peering` as well?

danaelhe (approved):
LGTM 🚀

PR #1166, opened 2024-05-23
Databases: Save DB password from initial POST request

razum90:
For all databases. Use password from subsequent GET requests in case it's not empty.

Fixes #1165.

PR #1169, opened 2024-05-29
dbaas:  ensure password is always retained in connection URIs, replicas, and connection pools.

andrewsomething:
This builds on the work in https://github.com/digitalocean/terraform-provider-digitalocean/pull/1166 to also make sure the password is not overwritten in connection URIs, replicas, and connection pools.


danaelhe (approved):
👍

Issue #1185, opened 2024-07-25
Imported database replica attempting to resize to current configuration


cphalpin:
# Bug Report

---

## Describe the bug

I've imported an existing database replica into Terraform Cloud. New run plans now say they will add a size of "db-s-1vcpu-2gb", which I believe is the current size of the replica, and when I try to apply it I get the error

```
Error: Error resizing database replica: PUT https://api.digitalocean.com/v2/databases/.../resize: 422 (request "...") cannot scale to same configuration
```

### Affected Resource(s)

digitalocean_database_replica

### Expected Behavior

I expect no change to be needed here since the replica should already be the correct size.

### Actual Behavior

Terraform is trying to call the resize API to set the replica to its current size.

### Steps to Reproduce

1. Create a new database replica through the DO UI.
2. Add matching configuration to terraform
3. Import existing replica with `terraform import`
4. Try to make any kind of changes with `terraform apply`

**Terraform Configuration Files**

config.tf
```hcl
locals {
  region = "lon1"

  environments = [
    "dev",
    "staging",
    "prod",
  ]

  database_server_size = "db-s-1vcpu-2gb"
  database_server_has_standby_node_default = false

  database_server_has_standby_node_overrides = {
  }

  database_servers = {
    for environment in local.default_environments : environment => {
      size = local.database_server_size,
      has_standby_node = lookup(local.database_server_has_standby_node_overrides, environment, local.database_server_has_standby_node_default),
    }
  }
}
```

databases.tf
```hcl
locals {
  sql_modes = [
    "ERROR_FOR_DIVISION_BY_ZERO",
    "IGNORE_SPACE",
    "NO_ENGINE_SUBSTITUTION",
    "NO_ZERO_DATE",
    "NO_ZERO_IN_DATE",
    "ONLY_FULL_GROUP_BY",
    "PIPES_AS_CONCAT",
    "REAL_AS_FLOAT",
    "STRICT_ALL_TABLES",
  ]
}

resource "digitalocean_database_cluster" "database_servers" {
  for_each = local.database_servers

  name = each.key
  engine = "mysql"
  version = "8"
  size = each.value.size
  region = local.region
  node_count = each.value.has_standby_node ? 2 : 1
  sql_mode = join(",", local.sql_modes)
  tags = [
    "env:${each.key}",
    "client:ubs",
  ]
}

resource "digitalocean_database_replica" "production_database_replica" {
[...]

PR #1186, opened 2024-07-26
databases: set replica size to state on read.

andrewsomething:
Now that `size` is available in the `GET /v2/databases/{uuid}/replicas/{name}` response, we can set it to state on read.

Resolves: https://github.com/digitalocean/terraform-provider-digitalocean/issues/1185


danaelhe (approved):
LGTM 🚀

Issue #1181, opened 2024-07-23
digitalocean_app: Service Autoscaling


moritzbruder:
### Is your feature request related to a problem? Please describe.
App Platform has an autoscaling feature that's not supported by the terraform provider as far as I can tell.

It's configurable via UI and app spec but already but not terraform:
https://www.digitalocean.com/blog/introducing-cpu-based-autoscaling-app-platform

### Describe the solution you'd like
Support for `autoscaling` properties of the `digitalocean_app.service` resources
 
### Describe alternatives you've considered
no usable alternatives really except building my own scaling or not using app platform, autoscaling was part of the reason for choosing app platform for me

PR #1189, opened 2024-07-30
digitalocean_app: Service Autoscaling

moritzbruder:
Adding autoscaling spec for `service` resources in `digitalocean_app`

Resolves https://github.com/digitalocean/terraform-provider-digitalocean/issues/1181

> 👉  This required the removal of the default value of `1` for the `instance_count` service property. API calls were failing because `instance_count` cannot be used at the same time as `autoscaling`. To my understanding this shouldn't be a breaking change though, since the API applies a default of 1 anyways as specified [here](https://docs.digitalocean.com/reference/api/api-reference/#operation/apps_create)

### DIFF
--- a/digitalocean/app/app_spec.go
+++ b/digitalocean/app/app_spec.go
@@ -449,13 +449,13 @@ func appSpecAutoscalingSchema() map[string]*schema.Schema {
 	return map[string]*schema.Schema{
 		"min_instance_count": {
 			Type:         schema.TypeInt,
-			Optional:     false,
+			Required:     true,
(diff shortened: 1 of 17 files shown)


andrewsomething (approved):
Hi @moritzbruder!

:pray:  Thanks so much for this great contribution!

I did push a couple additional commits to get this over the line. The main one being appending to the result in `flattenAppAutoscaling`. Without that, the app was successfully created with autoscaling enabled, but Terraform always saw a diff as it was not written to state. I also separated out the acceptance test as the config you changed was shared by a number of other tests requiring additional changes.

PR #1193, opened 2024-08-02
Fix Postgres Config TimescaleDB configuration

linuxluigi:
Fix Postgres Config TimescaleDB configuration. Until now it expect a list of objects but the godo library expect it as a single object. Also, the fixing the attribute key for TimescaleDB configuration.

I'm not sure if I implemented this correct, since I'm not really sure how to test this.


danaelhe:
Oh yes, good catch! In the [documentation](https://docs.digitalocean.com/reference/api/api-reference/#operation/databases_patch_config) it is suppose to be `max_background_workers`. Added a test for it and will merge on green.

Thank you so much!

danaelhe:
Tested it out locally and we are good to go! Thanks again.

danaelhe (approved):
🚀

Issue #1175, opened 2024-06-26
Add OpenSearch Documentation


ngfgrant:
### Is your feature request related to a problem? Please describe.

 
I think it would be useful to include some examples for:

1) Setting up cluster (such as:  https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/resources/database_cluster#create-a-new-postgresql-database-cluster) 
2) OpenSearch config (such as https://registry.terraform.io/providers/digitalocean/digitalocean/latest/docs/resources/database_mysql_config)
2) Adding/Configuring Managed Database log sink for each available. 
3) Adding/Configuring Log Forwarding Destination for App Platform in the App Platform docs.

### Describe alternatives you've considered

I am not sure there are any alternatives, this should be documented. 

### Additional context

I am happy to assist with some PRs for this.

danaelhe:
Addressed in #1198 🫶 

Thank you for opening the issue!

GiovanniColonni:
I only see [this file ](https://github.com/digitalocean/terraform-provider-digitalocean/pull/1198/files) but it's already merged so maybe I don't understand the request

PR #1198, opened 2024-08-12
Add opensearch to cluster reference

seano-vs:
Since this is now in use, it might be good to call it out int he docs.


seano-vs:
Should work towards #1175

danaelhe:
Tested it out and all looks great 💯 will merge on green. Thank you SO much for your contribution 🫶

danaelhe (approved):
LGTM

PR #1200, opened 2024-08-21
Added support for opensearch configuration on log_destination block

GiovanniColonni:
WIth this PR we would like to add support for OpenShift using the **log_destination** block. 
based on the spec of digital ocean we should have something like this:
```
log_destination: 
  -name: app-staging-opensearch
  open_search:
   basic_auth:
          user: username
   cluster_name: cluster_name
   ndex_name: logs
 ```
 
 In the specification created setting the forward from digital ocean from the UI there is no field for `password` in the `basic_auth` and im not sure if i have to include it or not

### DIFF
--- a/digitalocean/app/app_spec.go
+++ b/digitalocean/app/app_spec.go
@@ -1392,8 +1392,8 @@ func expandAppOpensearchBasicAuth(config []interface{}) *godo.OpenSearchBasicAut
 	basicAuthConfig := config[0].(map[string]interface{})
 
 	basicAuth := &godo.OpenSearchBasicAuth{
-		User:     string(basicAuthConfig["user"].(string)),
-		Password: string(basicAuthConfig["password"].(string)),
+		User:     basicAuthConfig["user"].(string),
+		Password: basicAuthConfig["password"].(string),
 	}
 
 	return basicAuth
--- a/digitalocean/app/app_spec.go
+++ b/digitalocean/app/app_spec.go
@@ -875,6 +875,46 @@ func appSpecLogDestinations() *schema.Resource {
 				Required:    true,
 				Description: "Name of the log destination",
 			},
+			"open_search": {
+				Type:        schema.TypeList,
+				Optional:    true,
+				MaxItems:    1,
+				Description: "OpenSearch configuration.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"endpoint": {
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "OpenSearch endpoint.",
+						},
+						"basic_auth": {
+							Type:        schema.TypeList,
+							Required:    true,
+							Description: "Basic authentication details.",
+							MaxItems:    1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"username": {
+										Type:        schema.TypeString,
+										Required:    true,
+										Description: "Username for basic authentication.",
+									},
+									"password": {
+										Type:        schema.TypeString,
+										Required:    true,
+										Sensitive:   true,
+										Description: "Password for basic authentication.",
+									},
+								},
+							},
+						},
+						"index_name":{
+							Type:        schema.TypeString,
+							Required:    true,
+							Description: "OpenSearch index name.",
+						}
+					}
+			}},
 			"papertrail": {
 				Type:        schema.TypeList,
 				Optional:    true,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by danaelhe** [comment]
#1175

**Comment by GiovanniColonni** [comment]
I see the PR for the documentation but the block for defining the destination is still not available to use in DO with terraform

**Comment by danaelhe** [comment]
> I see the PR for the documentation but the block for defining the destination is still not available to use in DO with terraform

Yes, you're right! Usually we include documentation in the same PR as the implemmentation. Could you add that in this PR if possible? Thank you SO much for your contribution!

**Comment by danaelhe** [comment]
> I only see [this file ](https://github.com/digitalocean/terraform-provider-digitalocean/pull/1198/files) but it's already merged so maybe I don't understand the request

Oh apologies- you're good then. The documentation you have in this PR already will suffice.

File: digitalocean/app/app_spec.go [hunk 1]
```diff
@@ -1,12 +1,11 @@
...
 
 import (
-	"log"
-	"net/http"
-
 	"github.com/digitalocean/godo"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+	"log"
+	"net/http"
```
danaelhe, line 8:
```suggestion
	"log"
	"net/http"

	"github.com/digitalocean/godo"
	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
```

^^ should fix lint ci issue!

digitalocean/app/app_spec.go
```diff
@@ -1243,6 +1289,38 @@ func expandAppLogDestinations(config []interface{}) []*godo.AppLogDestinationSpe
...
+				}
+				d.OpenSearch = &godo.AppLogDestinationSpecOpenSearch{
+					BasicAuth: &godo.OpenSearchBasicAuth{
+						User: (openSearchConfig["basic_auth"].([]interface{})[0].(map[string]interface{})["user"].(string)),
+					},
+					IndexName:   (openSearchConfig["index_name"].(string)),
+					ClusterName: (openSearchConfig["cluster_name"].(string)),
+				}
+			}
+		}
```
danaelhe, line 1322:
```suggestion
		if len(open_search) > 0 {
			openSearchConfig := open_search[0].(map[string]interface{})
			d.OpenSearch = &godo.AppLogDestinationSpecOpenSearch{
				Endpoint:    (openSearchConfig["endpoint"].(string)),
				BasicAuth:   expandAppOpensearchBasicAuth(openSearchConfig["basic_auth"].([]interface{})),
				IndexName:   (openSearchConfig["index_name"].(string)),
				ClusterName: (openSearchConfig["cluster_name"].(string)),
			}
		}
```
The API will do these checks for us :)

digitalocean/app/app_spec.go
```diff
@@ -875,6 +875,52 @@ func appSpecLogDestinations() *schema.Resource {
...
+						"basic_auth": {
+							Type:        schema.TypeList,
+							Required:    true,
+							Description: "Basic authentication details.",
+							MaxItems:    1,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"user": {
+										Type:        schema.TypeString,
+										Required:    true,
```
danaelhe, line 899:
```suggestion
										Optional:    true,
```
This is also optional:

> Username to authenticate with. Only required when endpoint is set. Defaults to doadmin when cluster_name is set.

digitalocean/app/app_spec.go
```diff
@@ -1307,6 +1385,42 @@ func flattenAppLogDestinations(destinations []*godo.AppLogDestinationSpec) []map
...
+					"basic_auth": []interface{}{
+						map[string]string{
+							"user": d.OpenSearch.BasicAuth.User,
+						},
+					},
+				}
+			}
+
+			r["opensearch"] = openSearch
+		}
```
danaelhe, line 1422:
```suggestion
		if d.OpenSearch != nil {
			openSearch := make([]interface{}, 1)

			openSearch[0] = map[string]interface{}{
				"endpoint":     d.OpenSearch.Endpoint,
				"cluster_name": d.OpenSearch.ClusterName,
				"index_name":   d.OpenSearch.IndexName,
				"basic_auth": []interface{}{
					map[string]string{
						"user":     d.OpenSearch.BasicAuth.User,
						"password": d.OpenSearch.BasicAuth.Password,
					},
				},
			}

			r["open_search"] = openSearch
		}
```
The API will perform these checks for us :)

digitalocean/app/app_spec.go
danaelhe: Also included this function 🤖 
```
func expandAppOpensearchBasicAuth(config []interface{}) *godo.OpenSearchBasicAuth {
	basicAuthConfig := config[0].(map[string]interface{})

	basicAuth := &godo.OpenSearchBasicAuth{
		User:     string(basicAuthConfig["user"].(string)),
		Password: string(basicAuthConfig["password"].(string)),
	}

	return basicAuth
}
```

danaelhe (changes requested):
Looks great!! Added a few changes. Tested locally and seems to work well:

```
resource "digitalocean_app" "golang-sample" {
  spec {
    name   = "golang-sample"
    region = "ams"

    service {
      name               = "go-service"
      environment_slug   = "go"
      instance_count     = 1
      instance_size_slug = "professional-xs"

      git {
        repo_clone_url = "https://github.com/digitalocean/sample-golang.git"
        branch         = "main"
      }

      log_destination {
        name = "MyLogs"
        open_search {
          endpoint = "https://something:1234"
          basic_auth {
            user = "user"
            password = "hi"
          }
        }
      }
    }
  }
}
```

GiovanniColonni (commented):
looks like all good at this point, thanks for helping

danaelhe (approved):
Looks great! Thank you so much for your contribution and help 🫶 

Added an example in docs in last commit to address #1175

PR #1201, opened 2024-09-01
Point docs to data attribute that actually exists

Mycobee:
This is a simple docs update to call an attribute that actually exists for the kubernetes cluster version

### DIFF
--- a/docs/resources/kubernetes_cluster.md
+++ b/docs/resources/kubernetes_cluster.md
@@ -61,14 +61,14 @@ For example:
 
 ```hcl
 data "digitalocean_kubernetes_versions" "example" {
-  version_prefix = "1.22."
+  version_prefix = "1.30."
 }
 
 resource "digitalocean_kubernetes_cluster" "foo" {
   name         = "foo"
   region       = "nyc1"
   auto_upgrade = true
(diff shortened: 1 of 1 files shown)


Mycobee:
@danaelhe may I get a review on this please?

danaelhe (approved):
Good catch👍🏽 Thank you so much! Will merge on green✅

Mycobee:
actually lets not do that. I think there are things under the hood that happen with `latest_version` that I misunderstood. I realize now that the original thing worked SMH

cc @danaelhe

Mycobee:
for some reason, after testing...this code is perfectly valid
```
data "digitalocean_kubernetes_versions" "current" {
  version_prefix = var.cluster_version
}

resource "digitalocean_kubernetes_cluster" "primary" {
  name    = var.cluster_name
  region  = var.cluster_region
  version = data.digitalocean_kubernetes_versions.current.latest_version

  node_pool {
    name       = "default"
    size       = var.worker_size
    node_count = var.worker_count
  }
}
```

Mycobee:
I suspect `latest_version` somehow gets the minor patch and adds it on to the `version_prefix`. Regardless, it works

danaelhe:
Ah I see. I’ll take a closer look on Tuesday.

PR #1202, opened 2024-09-02
Add support for Spaces in LON1

zoispag:
Support for Spaces in London was added 3 days ago.

https://www.digitalocean.com/blog/digitalocean-spaces-now-available-in-london


andrewsomething (approved):
:+1:  Good catch. Thanks for the PR!

PR #1203, opened 2024-09-05
dbaas: Use standardized name prefix in tests.

andrewsomething:
Logging into our acceptance test account showed that we had a handful of old test databases laying around from failed tests. The account sweeper failed to delete them since they were not using the standardized name prefix for acceptance tests:

https://github.com/digitalocean/terraform-provider-digitalocean/blob/78596c3e994acc58d452635ef0cdf5f1d47b75cc/digitalocean/database/sweep.go#L37

I've cleaned them up manually, but this should prevent more stranded resources in the future.


loosla (approved):
:+1:

Issue #1163, opened 2024-05-22
digitalocean_ssh_key race condition (roduced an unexpected new value: Root object was present, but now absent)


speedythesnail:
# Bug Report

---

## Describe the bug
When creating SSH keys, it appears there is a race condition where the SSH key is created but Terraform provides the below error message: 
```
│ Error: Provider produced inconsistent result after apply
│ 
│ When applying changes to module.ssh.digitalocean_ssh_key.personal_ssh_key,
│ provider
│ "module.ssh.provider[\"registry.terraform.io/digitalocean/digitalocean\"]"
│ produced an unexpected new value: Root object was present, but now absent.
│ 
│ This is a bug in the provider, which should be reported in the provider's
│ own issue tracker.
```
When I set the TF_LOG level to trace, I see that the SSH key is in fact created, and immediately after, the provider performs a GET request to get the key that was created, resulting in a 404 error. I manually performed a curl and see the key exists and it shows up in the Digital Ocean control panel.

I suspect there is a race condition, where the key was in fact created, but we are trying to retrieve information by performing the GET request too soon. See the below excerpt from the Terraform log:
```
2024-05-22T09:15:03.692-0400 [TRACE] checkPlannedChange: Verifying that actual change (action Create) matches planned change (action Create)
module.ssh.digitalocean_ssh_key.personal_ssh_key: Creating...
2024-05-22T09:15:03.692-0400 [TRACE] terraform.contextPlugins: Schema for provider "registry.terraform.io/digitalocean/digitalocean" is in the global cache
2024-05-22T09:15:03.692-0400 [INFO]  Starting apply for module.ssh.digitalocean_ssh_key.personal_ssh_key
2024-05-22T09:15:03.692-0400 [TRACE] terraform.contextPlugins: Schema for provider "registry.terraform.io/digitalocean/digitalocean" is in the global cache
2024-05-22T09:15:03.692-0400 [TRACE] terraform.contextPlugins: Schema for provider "registry.terraform.io/digitalocean/digitalocean" is in the global cache
2024-05-22T09:15:03.692-0400 [DEBUG] module.ssh.digitalocean_ssh_key.personal_ssh_key: applying the planned Create change
2024-05-22T09:15:03.692-0400 [TRACE] GRPCProvider: ApplyResourceChange
2024-05-22T09:15:03.692-0400 [TRACE] GRPCProvider: GetProviderSchema
[...]

Issue #923, opened 2023-01-31
Race condition when creating SSH keys


pielgrzym:
# Bug Report

---

## Describe the bug
Digital Ocean seems to have a race condition in their API - if you add an ssh-key and immediately query `/v2/account/keys/<NEW_KEY_ID>` your will get a 404, however the key exists. Provider does not handle this and fails.

### Affected Resource(s)
- digitalocean_ssh_key

### Expected Behavior
If the provider gets a 404 from DO API it should be able to retry after a delay.

### Actual Behavior
The provider marks the resource as absent.

### Steps to Reproduce

1. Create a tls_private_key in TF
2. Create digitalocean_ssh_key resource
3. Create droplet with digitalocean_ssh_key from previous point

Digitalocean has a bad day and after successfull ssh-key adding (`HTTP/2.0 201 Created` and proper json in trace) when the provider queries `/v2/account/keys/<NEW_KEY_ID>` it gets a 404. Clearly a mishap on DO side - since manual query returns a valid key - we have a race condition. Unfortunately the provider has no mechanism to handle this and will mark the resource as not created with a message: "Root resource was present, but now absent."

User will have to manually import the resoruce.

**Terraform Configuration Files**

**Terraform version**

**Debug Output**

**Panic Output**

## Additional context

**Important Factoids**

**References**

PR #1204, opened 2024-09-05
ssh: protect against database race condition in create

andrewsomething:
I've never been able to reproduce this issue, but as we have received multiple reports we should protect against it. While we retry 5xx errors, we do not generally retry 404s. This adds a retryable GET after creating an SSH key that blocks until the key is found. By default, we'll retry for 2 minutes. I think that is probably more than enough, but I've made it configurable as well.

Resolves: https://github.com/digitalocean/terraform-provider-digitalocean/issues/1163
Resolves: https://github.com/digitalocean/terraform-provider-digitalocean/issues/923

### DIFF
--- a/digitalocean/sshkey/resource_ssh_key.go
+++ b/digitalocean/sshkey/resource_ssh_key.go
@@ -3,12 +3,16 @@ package sshkey
 import (
 	"context"
 	"log"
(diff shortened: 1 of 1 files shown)


danaelhe (approved):
Nice!

Issue #1206, opened 2024-09-06
What is `environment_slug` in `digitalocean_app.service`?


bartekpacia:
Basically the title.

I couldn't find any information on it, and in examples I've only seen it assigned the `go` value:

```hcl
resource "digitalocean_app" "app" {
  spec {
    name   = "bee-ci-2"
    region = "sfo"

    service {
      name               = "backend"
      environment_slug   = "go" # <--- Where this comes from?
      instance_count     = 1
      instance_size_slug = "apps-s-1vcpu-0.5gb"
    }
   
    # ...
  }
}
```

Does Go mean it's a Golang env? Or what?

andrewsomething:
Hi @bartekpacia,

This recently came up over in a discussion thread: https://github.com/digitalocean/terraform-provider-digitalocean/discussions/1190

Environment slugs were a hint about the type of app, but they no longer have any material impact on your app. You can exclude them completely from your app spec. They will be deprecated at some point in the future.

We should probably go ahead and remove them from the examples to prevent confusion.

PR #1208, opened 2024-09-09
apps: update examples to remove environment_slug

andrewsomething:
We've seen a few questions about `environment_slug` recently. As it has no real effect anymore and will be deprecated at some point in the future, this removes it from the examples. While I'm in here, I also updated the example `instance_size_slug` to use the new style slugs.

Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/1206


danaelhe (approved):
👍

PR #1212, opened 2024-09-17
[databases]: add support for MongoDB advanced configuration

loosla:


digitalocean/database/resource_database_mongodb_config.go
```diff
@@ -0,0 +1,174 @@
...
+			State: resourceDigitalOceanDatabaseMongoDBConfigImport,
+		},
+		Schema: map[string]*schema.Schema{
+			"cluster_id": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.NoZeroValues,
+			},
+			"connect_default_read_concern": {
```
andrewsomething, line 31: Looks like a typo here:

```suggestion
			"default_read_concern": {
```

digitalocean/database/resource_database_mongodb_config.go
```diff
@@ -0,0 +1,174 @@
...
+				Optional: true,
+				Computed: true,
+				ValidateFunc: validation.StringInSlice(
+					[]string{
+						"local",
+						"available",
+						"majority",
+					},
+					false,
+				),
```
andrewsomething, line 42: The API seems to always return this field in uppercase which can cause unexpected diffs like:

```

  # digitalocean_database_mongodb_config.example will be updated in-place
  ~ resource "digitalocean_database_mongodb_config" "example" {
      ~ default_read_concern               = "MAJORITY" -> "majority"
        id                                 = "a12903dc-13bf-4ee3-8995-3d0507a4310e/mongodb-config"
        # (5 unchanged attributes hidden)
    }
```

We can workaround this by seeing `ignoreCase` to `true` in the `ValidateFunc` and by using an `DiffSuppressFunc` to ignore the case when comparing the local and remote values.

https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#diffsuppressfunc

```suggestion
				ValidateFunc: validation.StringInSlice(
					[]string{
						"local",
						"available",
						"majority",
					},
					true,
				),
				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
					return strings.EqualFold(old, new)
				},
```

andrewsomething (changes requested):
Awesome work!

I did notice a couple small issue. I left suggestion in-line,

digitalocean/database/resource_database_mongodb_config.go
loosla: `DiffSuppressFunc` definitely helps :+1: 
Thanks

andrewsomething (approved):
:rocket:

PR #1214, opened 2024-09-18
Update workflows for Go 1.22.

andrewsomething:
Following on https://github.com/digitalocean/terraform-provider-digitalocean/pull/1213 where the Go version was bumped to Go 1.22, this updates the workflows for the new version.

I've also updated to a version of golangci-lint with support for Go 1.23. This brought in some new checks that we are not passing. I've disabled them for now, but we should work through resolving them separately.


.golangci.yml
```diff
@@ -23,7 +23,6 @@ linters:
     - govet
     - typecheck
     - unconvert
-    - varcheck
```
andrewsomething, line 26: `varcheck` was deprecated.

```
ERRO [linters_context] varcheck: This linter is fully inactivated: it will not produce any reports. 
```

loosla (approved):
Thank you :rocket:

PR #1252, opened 2024-10-18
[release]: prep v2.43.0

loosla:
Release v2.43.0


andrewsomething (approved):
:rocket: 

(I pushed a commit with a tiny change moving the revert commit to the MISC section since they were reverting things that were never released in the first place.)

PR #1262, opened 2024-11-04
[databases]: update a description for redis_ssl in redis advanced config

loosla:
This PR adds information on which ports to use depending on `redis_ssl` value.

> When SSL is disabled, the cluster opens port 25060 for non-ssl connection while port 25061 still will be available for SSL connection.


andrewsomething (approved):
:+1:

Issue #1266, opened 2024-11-15
digitalocean_database_opensearch_config creation fails


brianhelba:
# Bug Report

## Describe the bug
Creation of a new `digitalocean_database_opensearch_config` fails.

### Affected Resource(s)
- `digitalocean_database_opensearch_config`

### Steps to Reproduce

1. With a resource definition like:
    ```terraform
    resource "digitalocean_database_opensearch_config" "elasticsearch" {
      cluster_id = digitalocean_database_cluster.elasticsearch.id
    
      # Adding additional optional arguments does not change the failing behavior
    }
    ```
    creation via `terraform apply` fails with the error (unique identifiers are redacted with `xxxx`):
    ```
    │ Error: Error updating Opensearch configuration: PATCH https://api.digitalocean.com/v2/databases/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/config: 422 (request "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx") the mask must be set with which fields are being updated
    ```

2. Running `TF_LOG=DEBUG terraform apply` shows that the HTTP API request to DigitialOcean is:
    ```
    ---[ REQUEST ]---------------------------------------
    PATCH /v2/databases/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/config HTTP/1.1
    Host: api.digitalocean.com
    User-Agent: Terraform/1.9.8 godo/1.126.0
    Content-Length: 14
    Accept: application/json
    Content-Type: application/json
    Accept-Encoding: gzip
    
    {
     "config": {}
    }
    ````
    with a response (some non-essential headers redacted):
    ```
    ---[ RESPONSE ]--------------------------------------
    HTTP/2.0 422 Unprocessable Entity
    Content-Length: 151
    Content-Type: application/json; charset=utf-8
    Server: cloudflare
    SameSite=None
    
    {
     "message": "the mask must be set with which fields are being updated",
     "id": "unprocessable_entity",
     "request_id": "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
    }
    ```
3. As noted above, adding additional optional arguements to the `digitalocean_database_opensearch_config` resource does not change the behavior: the underlying HTTP request still sends a PATCH with an empty `config` in the body.

### Workaround
[...]

PR #1268, opened 2024-11-15
opensearch_config: only send PATCH on create if values set.

andrewsomething:
As reported in https://github.com/digitalocean/terraform-provider-digitalocean/issues/1266, "creating" a `digitalocean_database_opensearch_config` without any setting leads to an error. This change protects against that by preventing sending the PATCH on create if no other values beside `cluster_id` are set.

Additionally, I ran into some issue running the acceptance test. One is a bug. We need to use `GetOkExists` when checking the value for a bool. If not, it is not correctly detected when being set to `false`.

There are also a few changes to the test that do not impact functionality which I will note inline.

### DIFF
--- a/digitalocean/database/resource_database_opensearch_config_test.go
+++ b/digitalocean/database/resource_database_opensearch_config_test.go
@@ -20,17 +20,21 @@ func TestAccDigitalOceanDatabaseOpensearchConfig_Basic(t *testing.T) {
 			{
 				Config: fmt.Sprintf(testAccCheckDigitalOceanDatabaseOpensearchConfigConfigBasic, dbConfig, true, 10, "1"),
 				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("digitalocean_database_opensearch_config.foobar", "enable_security_audit", "true"),
 					resource.TestCheckResourceAttr("digitalocean_database_opensearch_config.foobar", "ism_enabled", "true"),
+					resource.TestCheckResourceAttr("digitalocean_database_opensearch_config.foobar", "ism_history_enabled", "true"),
 					resource.TestCheckResourceAttr("digitalocean_database_opensearch_config.foobar", "ism_history_max_age_hours", "10"),
 					resource.TestCheckResourceAttr("digitalocean_database_opensearch_config.foobar", "ism_history_max_docs", "1"),
 				),
 			},
 			{
-				Config: fmt.Sprintf(testAccCheckDigitalOceanDatabaseOpensearchConfigConfigBasic, dbConfig, false, 1, "9223372036854775807"),
+				Config: fmt.Sprintf(testAccCheckDigitalOceanDatabaseOpensearchConfigConfigBasic, dbConfig, false, 1, "1000000000000000000"),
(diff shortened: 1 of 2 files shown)


digitalocean/database/resource_database_opensearch_config_test.go
```diff
@@ -20,17 +20,21 @@ func TestAccDigitalOceanDatabaseOpensearchConfig_Basic(t *testing.T) {
...
+					resource.TestCheckResourceAttr("digitalocean_database_opensearch_config.foobar", "enable_security_audit", "true"),
 					resource.TestCheckResourceAttr("digitalocean_database_opensearch_config.foobar", "ism_enabled", "true"),
+					resource.TestCheckResourceAttr("digitalocean_database_opensearch_config.foobar", "ism_history_enabled", "true"),
 					resource.TestCheckResourceAttr("digitalocean_database_opensearch_config.foobar", "ism_history_max_age_hours", "10"),
 					resource.TestCheckResourceAttr("digitalocean_database_opensearch_config.foobar", "ism_history_max_docs", "1"),
 				),
 			},
 			{
-				Config: fmt.Sprintf(testAccCheckDigitalOceanDatabaseOpensearchConfigConfigBasic, dbConfig, false, 1, "9223372036854775807"),
+				Config: fmt.Sprintf(testAccCheckDigitalOceanDatabaseOpensearchConfigConfigBasic, dbConfig, false, 1, "1000000000000000000"),
```
andrewsomething, line 31: We're hitting a bug with upstream Terraform here. I've change this to another value to allow us to tests the broader functionality despite this.

https://github.com/hashicorp/terraform-plugin-sdk/issues/1215

digitalocean/database/resource_database_opensearch_config_test.go
```diff
@@ -20,17 +20,21 @@ func TestAccDigitalOceanDatabaseOpensearchConfig_Basic(t *testing.T) {
 			{
 				Config: fmt.Sprintf(testAccCheckDigitalOceanDatabaseOpensearchConfigConfigBasic, dbConfig, true, 10, "1"),
 				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("digitalocean_database_opensearch_config.foobar", "enable_security_audit", "true"),
 					resource.TestCheckResourceAttr("digitalocean_database_opensearch_config.foobar", "ism_enabled", "true"),
+					resource.TestCheckResourceAttr("digitalocean_database_opensearch_config.foobar", "ism_history_enabled", "true"),
```
andrewsomething, line 25: Apparently both `ism_enabled` and `ism_history_enabled` both must be true when setting the other `ism_history` settings or the API returns errors, e.g.

```
        Error: Error updating Opensearch configuration: PATCH https://api.digitalocean.com/v2/databases/d6b65b4f-954d-4ee0-a29b-29eb2b4d228f/config: 422 (request "7f2461f9-6c49-4291-a304-364f1d5eb740") invalid 'user_config': invalid input for opensearch: 'ism_history_enabled' is a required property
```

I've added `enable_security_audit` so that we have another bool value to test changes. This exposed the other bug.

loosla (approved):
:rocket:

PR #1270, opened 2024-11-19
docs(droplet) details `user_data` behavior (resource forces recreate)

dduportal:
We [recently discovered](https://github.com/jenkins-infra/digitalocean/pull/211#issuecomment-2485170603) that `user_data` is immutable for a given Droplet and changing its value forces a resource re-creation.

It is a different behavior than other Terraform providers (Azure does not care, AWS triggers instance reboot, etc.) so it is worth documenting it.

This PR updates the `user_data` description by adding the link to official DigitalOcean documentation and specifying that changing this attribute forces a new resource.


andrewsomething (approved):
Thanks! Agreed that this is a useful clarification.

Issue #1272, opened 2024-11-20
digitalocean_database_opensearch_config touches unnecessary options on change


brianhelba:
# Bug Report

## Describe the bug
When performing change actions, `digitalocean_database_opensearch_config` sends HTTP requests with many unnecessary options.

This is especially impactful, since sending the particular option `override_main_response_version` triggers an unrelated upstream bug within DigitalOcean OpenSearch, whereby the database version is misreported, which disrupts usage of the related `digitalocean_database_cluster`.

### Affected Resource(s)
- `digitalocean_database_opensearch_config`
- `digitalocean_database_cluster` (indirectly)

### Expected Behavior
For the change from the resource:
```terraform
resource "digitalocean_database_opensearch_config" "elasticsearch" {
  cluster_id = digitalocean_database_cluster.elasticsearch.id
}
```
to
```terraform
resource "digitalocean_database_opensearch_config" "elasticsearch" {
  cluster_id = digitalocean_database_cluster.elasticsearch.id

  action_auto_create_index_enabled = false
}
```

I would expect an HTTP request like:
```
PATCH /v2/databases/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/config HTTP/1.1
Host: api.digitalocean.com
User-Agent: Terraform/1.9.8 godo/1.129.0

{
 "config": {
  "action_auto_create_index_enabled": false
 }
}
```

### Actual Behavior
Instead, in response to this change, `terraform apply` makes an HTTP request like:
```
PATCH /v2/databases/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/config HTTP/1.1
Host: api.digitalocean.com
User-Agent: Terraform/1.9.8 godo/1.129.0

{
 "config": {
  "http_max_content_length_bytes": 100000000,
  "http_max_header_size_bytes": 8192,
  "http_max_initial_line_length_bytes": 4096,
  "indices_query_bool_max_clause_count": 1024,
  "indices_memory_index_buffer_size_percentage": 10,
  "indices_memory_min_index_buffer_size_mb": 48,
  "indices_queries_cache_size_percentage": 10,
  "indices_recovery_max_mb_per_sec": 40,
  "indices_recovery_max_concurrent_file_chunks": 2,
  "ism_enabled": true,
  "ism_history_enabled": true,
  "ism_history_max_age_hours": 24,
  "ism_history_max_docs": 2500000,
[...]

PR #1273, opened 2024-11-21
opensearch_config: follow PATCH semantics

andrewsomething:
This change makes it so that we only send attributes that have changes for Opensearch configurations. The one exception is for the ISM-related settings. The API seems to require that they are sent together, e.g.:

```
 Error: Error updating Opensearch configuration: PATCH https://api.digitalocean.com/v2/databases/f3587545-9002-415b-9593-8473bc770a98/config: 422 (request "41520d54-cf5e-408d-ae19-e4fd45c310fa") invalid 'user_config': invalid input for opensearch: 'ism_history_enabled' is a required property
```

We may want to consider making similar changes for the other database engines. Though that is less pressing. The main motivator here is working around the bug triggered by sending `override_main_response_version`.

Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/1272

### DIFF
--- a/digitalocean/database/resource_database_opensearch_config.go
+++ b/digitalocean/database/resource_database_opensearch_config.go
@@ -280,165 +280,234 @@ func updateOpensearchConfig(ctx context.Context, d *schema.ResourceData, client 
 
 	opts := &godo.OpensearchConfig{}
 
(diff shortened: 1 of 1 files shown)


loosla (approved):
:rocket:  Thank you

PR #1279, opened 2024-11-27
LBAAS-3552: add lb ipv6 field

jvasilevsky:
add lb ipv6 field


loosla (approved):
:rocket: Looks good to me.

Issue #1289, opened 2024-12-17
`opensearch_acl` attempting to be set in digitalocean_database_user data, even when no settings block


iainsproat:
# Bug Report

Bumping from `digitalocean/digitalocean` `2.39.2` to `2.46.0`, with an existing DigitalOcean Postgres database cluster and user.

On running terraform plan, an error is reported related to `opensearch_acl`. I do not use opensearch and am not defining any opensearch resources in the terraform plan.

## Describe the bug

An error is thrown when updating digitalocean_database_user

```
Error: Error setting user settings: &errors.errorString{s:"Invalid address to set: []string{\"settings\", \"0\", \"opensearch_acl\"}"}
...
data "digitalocean_database_user" "doadmin" {
```

### Affected Resource(s)

- data digitalocean_database_user

### Expected Behavior

The plan should have run without error.

### Actual Behavior

```
Error: Error setting user settings: &errors.errorString{s:"Invalid address to set: []string{\"settings\", \"0\", \"opensearch_acl\"}"}
...
data "digitalocean_database_user" "doadmin" {
```

### Steps to Reproduce

1. Create a Postgres database cluster & database user (doadmin will suffice) in DigitalOcean UI
2. Create a Terraform data for `digitalocean_database_user`, referencing the ID of the existing resource.

```terraform
data "digitalocean_database_user" "doadmin" {
  cluster_id = "my-postgres-cluster-id"
  name       = "doadmin"
}
```

4. Run terraform plan with terraform version 2.46.0

**Terraform Configuration Files**

**Terraform version**

`v1.9.8`

**Debug Output**

**Panic Output**

## Additional context

**Important Factoids**

**References**

andrewsomething:
Hi @iainsproat, 

Thanks for flagging this issue! The change in https://github.com/digitalocean/terraform-provider-digitalocean/pull/1290 should resolve this. We just cut a new v2.46.1 release with it that should be available on the Terraform registry shortly.

PR #1215, opened 2024-09-20
Update Terraform provider to include new opensearch acl changes

bhardwajRahul:
This PR aims at adding support for opensearch acl changes to DO's Terraform Provider.

##### Testing instructions(for future references):
```
make testacc TESTARGS='-run=TestAccDigitalOceanDatabaseUser_OpenSearchACLs'

```

PR #1290, opened 2024-12-17
fix: add opensearch_acl to data.digitalocean_database_user schema

andrewsomething:
This bug was introduced when `opensearch_acl` support was added to the database cluster resource (https://github.com/digitalocean/terraform-provider-digitalocean/pull/1215). The resource and data source share the `flattenUserSettings` method. When the data source attempts to set the empty `opensearch_acl` to state, it errors as it is not part of the datasource schema. 

https://github.com/digitalocean/terraform-provider-digitalocean/blob/f27a1125229e865da2a20cb0682d57b38433ce82/digitalocean/database/resource_database_user.go#L348-L357

The acceptance test did not catch this as the user needs to have a `setting` block in the API response as the method checks it `settings` is `nil`. A brand new Postgres user does not, but a `doadmin` user does, e.g.

```
{
  "user": {
    "name": "doadmin",
    "role": "primary",
    "password": "<redacted>",
    "settings": {
      "pg_allow_replication": true
    }
  }
}
```

Updated test now passes, but before the change would fail with:

```
panic: Invalid address to set: []string{"settings", "0", "opensearch_acl"}
```

Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/1289

### DIFF
--- a/digitalocean/database/datasource_database_user.go
+++ b/digitalocean/database/datasource_database_user.go
@@ -56,6 +56,11 @@ func DataSourceDigitalOceanDatabaseUser() *schema.Resource {
 							Optional: true,
 							Elem:     userACLSchema(),
 						},
+						"opensearch_acl": {
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem:     userOpenSearchACLSchema(),
+						},
 					},
 				},
 			},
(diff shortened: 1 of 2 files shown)


loosla (approved):
:pray: Thank you

PR #1293, opened 2024-12-20
Add maintenance day validation for database and kubernetes clusters

moreinhardt:
The docs did not really tell me how to use this field.

### DIFF
--- a/digitalocean/kubernetes/resource_kubernetes_cluster.go
+++ b/digitalocean/kubernetes/resource_kubernetes_cluster.go
@@ -121,10 +121,20 @@ func ResourceDigitalOceanKubernetesCluster() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"day": {
-							Type:         schema.TypeString,
-							Optional:     true,
-							Computed:     true,
-							ValidateFunc: validation.StringInSlice([]string{"any", "monday", "tuesday", "wednesday", "thursday", "friday", "saturday", "sunday"}, true),
+							Type:     schema.TypeString,
+							Optional: true,
+							Computed: true,
+							ValidateFunc: validation.StringInSlice([]string{
+								"any",
+								"monday",
+								"tuesday",
+								"wednesday",
+								"thursday",
+								"friday",
+								"saturday",
+								"sunday"},
+								true,
+							),
 						},
 						"start_time": {
 							Type:     schema.TypeString,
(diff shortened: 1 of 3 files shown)


digitalocean/kubernetes/resource_kubernetes_cluster.go
```diff
@@ -121,9 +121,10 @@ func ResourceDigitalOceanKubernetesCluster() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"day": {
-							Type:     schema.TypeString,
-							Optional: true,
-							Computed: true,
+							Type:         schema.TypeString,
+							Optional:     true,
+							Computed:     true,
+							ValidateFunc: validation.IsDayOfTheWeek(true),
```
moreinhardt, line 127: I did not validate this also works with capitalized weekdays. The provider docs list lowercase examples but in the case of database clusters I tested and both were accepted by the DO API.

andrewsomething: I think ignoring case here is correct, but in addition to days of the week, DOKS maintenance policies can also set `any` if the user has no preference. 

`validation.IsDayOfTheWeek` is just a thin wrapper around `validation.StringInSlice`. So we can use that here it support `any` as well.

https://github.com/hashicorp/terraform-plugin-sdk/blob/00fc746ade8f31a96a9030b867bedcff3365afa5/helper/validation/time.go#L13-L24

andrewsomething (changes requested):
Thanks for another helpful contribution! 

I left a comment in line re: DOKS maintenance policies also supporting `any`.

digitalocean/kubernetes/resource_kubernetes_cluster.go
moreinhardt: Ah right, I missed that even though I saw it in the docs.

I updated it and amended my commit. Hope that's alright.

digitalocean/kubernetes/resource_kubernetes_cluster.go
```diff
@@ -121,9 +121,10 @@ func ResourceDigitalOceanKubernetesCluster() *schema.Resource {
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"day": {
-							Type:     schema.TypeString,
-							Optional: true,
-							Computed: true,
+							Type:         schema.TypeString,
+							Optional:     true,
+							Computed:     true,
+							ValidateFunc: validation.StringInSlice([]string{"any", "monday", "tuesday", "wednesday", "thursday", "friday", "saturday", "sunday"}, true),
```
moreinhardt, line 127: I don't have much experience with Go. I see that [Go has no line length limit](https://go.dev/doc/effective_go). So `go fmt` didn't complain or change this.

Should I wrap this to have every value on a separate line for readability?

andrewsomething: It's not something that is required in Go, but it is still often nice for readability. I've pushed a commit doing so.

andrewsomething (approved):
:+1:  LGTM!

Thanks again for the contribution! Sorry it took so long to land with the holidays. :pray:

Issue #1275, opened 2024-11-23
digitalocean_app: add termination block (drain_seconds and grace_period_seconds)


ddelange:
### Is your feature request related to a problem? Please describe.

Hi 👋 

I found a discrepancy between the tf provider and the [app spec](https://docs.digitalocean.com/products/app-platform/reference/app-spec/): the termination block (drain_seconds and grace_period_seconds) is missing

### Describe the solution you'd like

termination block (drain_seconds and grace_period_seconds) added to tf provider
 
### Describe alternatives you've considered

N/A

### Additional context

N/A

ddelange:
hi @andrewsomething :wave: 

is this one for you too?

PR #1297, opened 2024-12-30
apps: add support for termination configuration.

andrewsomething:
https://docs.digitalocean.com/products/app-platform/how-to/configure-termination/

Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/1275

### DIFF
--- a/digitalocean/app/app_spec.go
+++ b/digitalocean/app/app_spec.go
@@ -19,6 +19,11 @@ const (
 	functionComponent   appSpecComponentType = "function"
 )
 
+// AppSpecTermination is a type constraint for the termination attribute of an app component.
+type AppSpecTermination interface {
+	godo.AppServiceSpecTermination | godo.AppWorkerSpecTermination | godo.AppJobSpecTermination
+}
+
 // appSpecSchema returns map[string]*schema.Schema for the App Specification.
 // Set isResource to true in order to return a schema with additional attributes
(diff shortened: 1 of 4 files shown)


loosla:
Looks awesome :+1: 
Andrew, could you explain me how we apply the default values in Terraform provider? :pray: 

```code
service {
      name               = "image-service"
      instance_count     = 1
      instance_size_slug = "apps-s-1vcpu-1gb"
      termination {
      }
```

This configuration with empty `termination` applies zeros to both fields: `drain_seconds` and `grace_period_seconds` when I run `terraform state show` after the config above has been successfully applied. 

The curl GET for this app shows termination as empty {}.

I'm curious, how we can apply the defaults that we mention in the documentation.
Thank you.

loosla:
Also, noticed that the `instance_count` in `worker` is forced to be updated each time even if it's value hasn't been changed. 
Constantly see it in changes on `terraform plan`.
However, this is not related to Apps Termination feature.

ddelange (approved):
LGTM

are there maximum values for these? would be nice to add to the docs!

andrewsomething:
Rebased main to resolve conflicts. 

@loosla I don't think we want to hardcode the defaults in this case. This allows more flexibility on the API side. When it's enforced there, they can be updated without needing to rebuild/re-release the provider.

PR #1303, opened 2025-01-13
Bitbucket for App Platform

dbrian57:
Adds Bitbucket option to App Platform Terraform docs and updates an example to use Bitbucket.


andrewsomething:
Bitbucket is not currently supported by the Terraform provider. Looks like it should follow a similar pattern as GitHub and GitLab. I can bang this out.

loosla:
@andrewsomething Just in case, we also don't have Bitbucket in doct. They were only added to openapi and godo in December 2024.

andrewsomething:
Closing in favor of https://github.com/digitalocean/terraform-provider-digitalocean/pull/1304

PR #1305, opened 2025-01-14
There is no minimum on services

Xowap:
The schema validation forces to have at least one service, however it's not actually a requirement from DigitalOcean.

In this PR I'm lowering the minimum to zero, I'm not sure if there are more things to be done?


Xowap:
It's kind of the issue I'm running into (through Pulumi though, so not sure if this has any impact?): I'm trying to create an app with just one worker and I can't because it doesn't have a service. I managed to get around that by getting the container to pose as a service, but that's definitely not ideal.

andrewsomething (approved):

andrewsomething:
It seems like Pulumi must evaluate this differently somehow. The Terraform SDK says:

>  If the field Optional is set to true then MinItems is ignored and thus effectively zero.

https://github.com/hashicorp/terraform-plugin-sdk/blob/b11308dc8f0ff4a03f750ba012e7bc7d3974e895/helper/schema/schema.go#L239-L241

Either way, let's remove it as it's not needed.

PR #1321, opened 2025-01-30
kubernetes: add a regression test for control plane firewall

loosla:


danaelhe (approved):
Thanks for the additional test!

PR #1329, opened 2025-02-21
partner interconnect operations support

apinonformoso:
related PR https://github.com/digitalocean/godo/pull/793


digitalocean/partnerinterconnectattachment/resource_partner_interconnect_attachment.go
```diff
@@ -0,0 +1,289 @@
...
+			partnerInterconnectAttachmentRequest.BGP = bgp
+		}
+	}
+
+	log.Printf("[DEBUG] Partner Interconnect Attachment create request: %#v", partnerInterconnectAttachmentRequest)
+
+	err := retry.RetryContext(ctx, d.Timeout(schema.TimeoutCreate), func() *retry.RetryError {
+		partnerInterconnectAttachment, _, err := client.PartnerInterconnectAttachments.Create(context.Background(), partnerInterconnectAttachmentRequest)
+		if err != nil {
+			return retry.RetryableError(fmt.Errorf("error creating Partner Interconnect Attachment: %s", err))
```
andrewsomething, line 151: Can we be a bit more precise here? We shouldn't retry every possible returned from the API. For example, this would just retry a 422 error over and over for five minutes.

 It's worth noting that the godo client used retries 429 and 500-level errors  with an exponential backoff. So using `RetryContext` may not be needed at all.

digitalocean/partnerinterconnectattachment/resource_partner_interconnect_attachment.go
```diff
@@ -0,0 +1,289 @@
...
+			},
+			"connection_bandwidth_in_mbps": {
+				Type:         schema.TypeInt,
+				Required:     true,
+				Description:  "The connection bandwidth in Mbps",
+				ValidateFunc: validation.NoZeroValues,
+			},
+			"region": {
+				Type:         schema.TypeString,
+				Required:     true,
```
andrewsomething, line 48:
```suggestion
				Required:     true,
				ForceNew: true,
```

Generally, if an attribute is not updateable, it should be set to `ForceNew: true`. Probably want this on `connection_bandwidth_in_mbps` and `naas_provider` too.

digitalocean/partnerinterconnectattachment/resource_partner_interconnect_attachment.go
```diff
@@ -0,0 +1,289 @@
...
+
+		d.SetId(partnerInterconnectAttachment.ID)
+
+		log.Printf("[DEBUG] Waiting for Partner Interconnect Attachment (%s) to become active", d.Get("name"))
+		stateConf := &retry.StateChangeConf{
+			Delay:      5 * time.Second,
+			Pending:    []string{"CREATING"},
+			Target:     []string{"CREATED"},
+			Refresh:    partnerInterconnectAttachmentStateRefreshFunc(client, d.Id()),
+			Timeout:    2 * time.Minute,
```
andrewsomething, line 162:
```suggestion
			Timeout:    d.Timeout(schema.TimeoutCreate),
```

digitalocean/partnerinterconnectattachment/resource_partner_interconnect_attachment.go
```diff
@@ -0,0 +1,289 @@
...
+		return nil
+	})
+
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	log.Printf("[INFO] Partner Interconnect Attachment created, ID: %s", d.Id())
+
+	return nil
```
andrewsomething, line 177: After creating the resource, you never set any of its attributes to state. One common pattern is to return the read method here to set everything. Though that does add an additional HTTP request. For example:

https://github.com/digitalocean/terraform-provider-digitalocean/blob/da119aebd61f90fbdc3f5d22db5eb1de63f4709f/digitalocean/vpc/resource_vpc.go#L116

digitalocean/partnerinterconnectattachment/resource_partner_interconnect_attachment_test.go
```diff
@@ -0,0 +1,165 @@
...
+	"testing"
+
+	"github.com/digitalocean/godo"
+	"github.com/digitalocean/terraform-provider-digitalocean/digitalocean/acceptance"
+	"github.com/digitalocean/terraform-provider-digitalocean/digitalocean/config"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
+)
+
+func TestAccDigitalOceanPartnerInterconnectAttachment_Basic(t *testing.T) {
```
andrewsomething, line 15: This test fails for me with `missing BGP auth key`. Is there any way to run these tests without access to a third-part account? We need to keep the test suite runnable by community contributors. So if not, we'll need to do something like source the key from an environment variable and skip the test if empty. 

```
=== CONT  TestAccDigitalOceanPartnerInterconnectAttachment_Basic
    resource_partner_interconnect_attachment_test.go:24: Step 1/3 error: Error running apply: exit status 1
        
        Error: error creating Partner Interconnect Attachment: POST https://api.digitalocean.com/v2/partner_interconnect/attachments: 422 (request "ff5f8150-3012-441d-841d-7e13131ceb69") missing BGP auth key
        
          with digitalocean_partner_interconnect_attachment.foobar,
          on terraform_plugin_test.tf line 12, in resource "digitalocean_partner_interconnect_attachment" "foobar":
          12: resource "digitalocean_partner_interconnect_attachment" "foobar" {
        
--- FAIL: TestAccDigitalOceanPartnerInterconnectAttachment_Basic (302.85s)
```

digitalocean/partnerinterconnectattachment/resource_partner_interconnect_attachment_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+		*partnerInterconnectAttachment = *foundPartnerInterconnectAttachment
+
+		return nil
+	}
+}
+
+const testAccCheckDigitalOceanPartnerInterconnectAttachmentConfig_Basic = `
+resource "digitalocean_vpc" "vpc1" {
+  name   = "vpc1"
```
andrewsomething, line 113: The VPC names used in these test configs should also be randomly generated to prevent clashes with any existing VPCs on the account used for testing.

docs/resources/partner_interconnect_attachment.md
```diff
@@ -0,0 +1,54 @@
+---
+page_title: "DigitalOcean: digitalocean_partner_interconnect_attachment"
+subcategory: "Networking"
+---
+
+# digitalocean_partner_interconnect_attachment
+
+-> Partner Interconnect Attachment is currently in alpha. If you are not a member of the alpha group for this feature, you will not be able to use it until is has been more widely released. Please follow the official [DigitalOcean changelog](https://docs.digitalocean.com/release-notes/) for updates.
```
andrewsomething, line 8: Going forward, we've been instructed to not use terms like "alpha." Instead we need to use the product lifecycle nomenclature. In this case, we probably want "private preview."

https://docs.digitalocean.com/platform/product-lifecycle/

andrewsomething (changes requested):

digitalocean/partnerinterconnectattachment/resource_partner_interconnect_attachment_test.go
apinonformoso: I've fixed that error, but I noticed godo is returning 0 for `peer_router_asn` I'll take a look to godo to fix that because it will show an error here until that is fixed.

apinonformoso: Looks like issue is coming from the response of public API, not godo. Working on a fix

digitalocean/partnerinterconnectattachment/resource_partner_interconnect_attachment.go
danaelhe: What scenario are you thinking that we'd need to retry a create request for? For example, sometimes we have to retry a request if we know we have to wait for another resource to be in a ready state (waiting on new database to accept connections, or a new droplet to be accessible etc.) or if we want a resource to be a certain status before we can use it.

apinonformoso: I can think of only one scenario here, waiting for the VPCs that will be attached to the partner interconnect attachment to be created.
Partner interconnect attachment needs the VPC UUIDs and those are obtained once the VPCs are fully created.

danaelhe: So terraform is smart enough to plan out those dependencies. Terraform analyzes the configuration to identify dependencies between resources, ensuring that resources are created or updated in the correct order, automatically inferring most dependencies through expression references (implicit dependencies). Terraform builds a graph of resources and their dependencies to determine the proper sequence of operations during deployment. 

More on terrraform implicit dependency can be found [here](https://developer.hashicorp.com/terraform/tutorials/configuration-language/dependencies#manage-implicit-dependencies).

We usually retry when terraform thinks a resource is ready (creates a database, receives a 200 status), but isn't (database isn't ready to accept connections yet) sorta thing. 

So I actually don't think we need to retry here.

apinonformoso: Ok, I'll remove it

apinonformoso: Done 👍

danaelhe (approved):
lgtm 👍

PR #1333, opened 2025-03-10
APPS Allow email and slack to be added for alert destination

thearyanahmed:
This PR adds support for configuring email and slack webhook alert destinations. 

This terraform schema doesn't map directly to app's spec. However, I tried to be consistent with api format (DO REST api) and on the individual elements as well. 

Open to discussion. 

Test terraform

```hcl
terraform {
  required_providers {
    digitalocean = {
      source  = "digitalocean/digitalocean"
      version = ">= 2.0.0"
    }
  }
}

provider "digitalocean" {
  token = var.do_token
}

resource "digitalocean_app" "testapp" {
  spec {
    name   = "app-1"
    region = "nyc"

    alert {
      rule     = "DOMAIN_FAILED"
      disabled = false
      destinations {
        emails = ["user1@email.com", "user2@email.com"]
        slack_webhooks {
            channel = "@user1
            url = "https://hooks.slack.com/services/slack-url-0"
        }
      }
    }

    alert {
      rule     = "DEPLOYMENT_FAILED"
      disabled = false
      destinations {
        emails = ["user1@email.com", "user3@email.com"]
        slack_webhooks {
            channel = "@user2
            url = "https://hooks.slack.com/services/slack-url-1"
        }
        slack_webhooks {
            channel = "#channel-1"
            url = "https://hooks.slack.com/services/slack-url-2"
        }
      }
    }

    service {
      name               = "go-api"
      instance_count     = 1
      instance_size_slug = "apps-s-1vcpu-1gb"

      github {
        branch         = "main"
        deploy_on_push = true
        repo           = "digitalocean/sample-golang"
      }

      alert {
        value    = 75
        operator = "GREATER_THAN"
        window   = "TEN_MINUTES"
        rule     = "CPU_UTILIZATION"
        destinations {
            emails = ["user1@email.com", "user3@email.com"]
            slack_webhooks {
                channel = "#channel-1"
                url = "https://hooks.slack.com/services/slack-url-2"
            }
        }
      }
    }
  }
}

```


digitalocean/app/resource_app.go
```diff
@@ -314,3 +347,227 @@ func waitForAppDeployment(client *godo.Client, id string, timeout time.Duration)
...
+
+			alertDetails = append(alertDetails, appAlert)
+		}
+	}
+
+	return alertDetails
+}
+
+func extractEmailAndSlackNotifications(notifications []interface{}) ([]string, []*godo.AppAlertSlackWebhook) {
+	notification := notifications[0].(map[string]interface{})
```
ZachEddy, line 546: I see that callers of `extractEmailAndSlackNotifications` ensure the slice length is greater than 0 before calling it, but IMO it's a bit easier to follow the code if the slice check (`if len(notifications) > 0 { ... }`) resides above where the slice is accessed (`notifications[0].(map[string]interface{}`) so it's obvious the zero-length case is handled.

Only mentioning it because the unchecked (`notifications[0]`) stuck out to me.

digitalocean/app/resource_app.go
```diff
@@ -314,3 +347,223 @@ func waitForAppDeployment(client *godo.Client, id string, timeout time.Duration)
...
+		})
+		if err != nil {
+			return err
+		}
+	}
+
+	if hasComputeComponentAlerts {
+		err := resourceDigitalOceanAlertDestinationUpdate(appID, alerts, computeComponentAlerts, client, func(a, b *godo.AppAlert) bool {
+			return string(a.Spec.Rule) == string(b.Spec.Rule)
+		})
```
thearyanahmed, line 444: I'm a bit unsettled on this part. It is only checking the rule. However, an user may set multiple alerts for the same rule different values.

I could've checked values and other parameters, but then if someone changes those as well, then this match will fail. 
Open to alternatives.

thearyanahmed: One idea was to make compare all 3 values and document it. If they are updating alert destinations (only destinations), the user must make sure of this. 
Worst case the user would need to run one more deployment.

digitalocean/app/app_spec.go
```diff
@@ -189,6 +188,40 @@ func appSpecAppLevelAlerts() *schema.Resource {
 				Default:  false,
 				Optional: true,
 			},
+			"notifications": {
```
blesswinsamuel, line 191: We call this destinations in the HTTP API. It probably makes sense to use the same terminology here.

```suggestion
			"destinations": {
```

digitalocean/app/app_spec.go
```diff
@@ -189,6 +188,40 @@ func appSpecAppLevelAlerts() *schema.Resource {
...
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"email": {
```
blesswinsamuel, line 196: For consistency with HTTP API

```suggestion
						"emails": {
```

digitalocean/app/app_spec.go
```diff
@@ -189,6 +188,40 @@ func appSpecAppLevelAlerts() *schema.Resource {
...
+							Type:     schema.TypeList,
+							Optional: true,
+							Elem: &schema.Schema{
+								Type:         schema.TypeString,
+								ValidateFunc: validation.StringLenBetween(3, 100),
+							},
+						},
+						"slack": {
```
blesswinsamuel, line 204: For consistency with HTTP API

```suggestion
						"slack_webhooks": {
```

digitalocean/app/resource_app.go
```diff
@@ -314,3 +347,223 @@ func waitForAppDeployment(client *godo.Client, id string, timeout time.Duration)
...
+
+					if len(emails) == 0 && len(slackWebhooks) == 0 {
+						continue
+					}
+
+					alertDetail.Emails = emails
+					alertDetail.SlackWebhooks = slackWebhooks
+				}
+			}
+		}
```
blesswinsamuel, line 408: Should we skip the disabled alerts?

digitalocean/app/app_spec.go
```diff
@@ -189,6 +188,40 @@ func appSpecAppLevelAlerts() *schema.Resource {
 				Default:  false,
 				Optional: true,
 			},
+			"notifications": {
+				Type:     schema.TypeList,
```
blesswinsamuel, line 192: Should alert destinations for an alert be an object instead of a list, since we assume only one alert destination per alert spec?

digitalocean/app/resource_app.go
```diff
@@ -314,3 +347,223 @@ func waitForAppDeployment(client *godo.Client, id string, timeout time.Duration)
...
+	specList := spec.([]interface{})
+	if len(specList) == 0 {
+		return nil
+	}
+	specMap, ok := specList[0].(map[string]interface{})
+	if !ok {
+		return nil
+	}
+
+	service, ok := specMap["service"]
```
blesswinsamuel, line 468: We support component level alerts for workers, jobs and functions as well. So I think we may want to support the same in terraform as well.

digitalocean/app/app_spec.go
thearyanahmed: It's not like this. I missed our api with alert team's api. They use these.

digitalocean/app/resource_app.go
thearyanahmed: In the UI, and API, alerts can be update while they are disabled. And I would like to keep the same.

blesswinsamuel (commented):
This is the first time I'm looking at a terraform provider code. So I have some learning to do. A question for my understanding: If I change only the alert destination (adding an email for example), how does terraform know that the alert destination changed when generating the plan (are we doing a diff anywhere, or is that handled by terraform)? I see ListAlerts called only while executing the plan. So my understanding is terraform doesn't detect a change in the app spec, so the alert destinations are not updated when only the alert destination is updated while the app spec remains the same. (please correct me if I'm wrong in any of my understanding)

Also, my concern with this approach is that users might think that destinations is part of the app spec and have the expectation that they can also configure alert destinations via the app spec using our UpdateApp API. But we don't support specifying alert destinations via UpdateApp API.

I wonder if we should move alert destination configs outside of the spec, or maybe as a separate resource instead. Or perhaps, we should consider making alert destinations configurable via the app spec.

thearyanahmed:
I'm not certain how adding a separate resource def. fully helps. Cause in that case, part of spec would be in the resource def, part of be in the app spec. That could also create a friction/confusion. 

I can simply update the documentation to mention this bit, what say you? @blesswinsamuel

blesswinsamuel:
> I'm not certain how adding a separate resource def. fully helps. Cause in that case, part of spec would be in the resource def, part of be in the app spec. That could also create a friction/confusion.

Alert destination configs are not part of the app spec today. Our HTTP APIs have a separate endpoint for configuring the alert destinations. So I was thinking we do the same here. But yeah, it does create friction since there needs to be a way to refer to the alerts defined in the app spec, and I don't see a way to refer to the alerts by a stable ID, so the only way seems to be to repeat all of the alerts config again with alert destinations added.

The other option is to consider making alert destinations configurable via the app spec. This would require looking at the history to see why this was not part of the app spec.

My only concern with making alert destinations part of the app spec only in terraform is that users might expect the same when using our UpdateApp API (via CLI or HTTP API or godo). In the UpdateApp API, we don't allow specifying alert destinations part of the app spec. So do you think that would cause confusion among users?

I think yes, but I'm not entirely sure. I'd defer this question to the API engineering team.

Other than this and the unresolved comments, this looks good to me! Nice work!

thearyanahmed:
Sorry for not being clear. The destinations are separate of the app spec, but the alerts, that contains the destinations are part of the specs. 

> The other option is to consider making alert destinations configurable via the app spec. This would require looking at the history to see why this was not part of the app spec.

The email is encrypted in our DB. And I guess I don't expect people keeping their terraform files with plaintext emails. In that case, it should be okay to allow emails in terraform files, as they should be private.

And to echo your concern, yes, there will be confusion in the scenario you mentioned in the UpdateApp api. My added comment is, we have friction regardless either way we put. 

Thoughts @blesswinsamuel

blesswinsamuel:
> The email is encrypted in our DB. And I guess I don't expect people keeping their terraform files with plaintext emails. In that case, it should be okay to allow emails in terraform files, as they should be private.

The same goes for certain env vars, image registry creds, log destination api tokens, etc. The app spec is encrypted in our db, with a redacted copy.

> And to echo your concern, yes, there will be confusion in the scenario you mentioned in the UpdateApp api. My added comment is, we have friction regardless either way we put.

Yes, I agree. If we were able to move the alert destinations into the app spec, I think that would be ideal. I guess one downside is creating too many deployments when updating the alert destinations via UI.

But maybe having different APIs is okay. I'm good with this approach if #api-engineering is good with having different HTTP and terraform APIs.

thearyanahmed:
Pausing this for now.

Issue #1338, opened 2025-03-14
Importing a global load balancer doesnt import TYPE properly


romankor:
# Bug Report

I have imported our global LB into the state and TF wants to recreate it now because it seems it didn't get the type of it 

```
  # digitalocean_loadbalancer.relays_lb must be replaced
-/+ resource "digitalocean_loadbalancer" "relays_lb" {
      ~ droplet_ids                      = [] -> (known after apply)
      ~ http_idle_timeout_seconds        = 60 -> (known after apply)
      ~ id                               = "752f1474-c85f-465c-8440-0f3336a04694" -> (known after apply)
      + ip                               = (known after apply)
      + ipv6                             = (known after apply)
        name                             = "relays-prod"
      ~ project_id                       = "54fbf00f-e73b-4e7c-b181-c1398deaa995" -> (known after apply)
      + size_unit                        = 2
      ~ status                           = "active" -> (known after apply)
      + target_load_balancer_ids         = (known after apply)
      + type                             = "GLOBAL" # forces replacement
      ~ urn                              = "do:loadbalancer:752f1474-c85f-465c-8440-0f3336a04694" -> (known after apply)
      + vpc_uuid                         = (known after apply)
        # (7 unchanged attributes hidden)

      + domains {
          + certificate_name             = (known after apply)
          + is_managed                   = false
          + name                         = "relays-do.twingate.com"
          + ssl_validation_error_reasons = (known after apply)
          + verification_error_reasons   = (known after apply)
        }

      ~ firewall (known after apply)

      + glb_settings {
          + target_port     = 443
          + target_protocol = "https"
        }

      ~ sticky_sessions (known after apply)
      - sticky_sessions {
          - cookie_ttl_seconds = 0 -> null
          - type               = "none" -> null
            # (1 unchanged attribute hidden)
        }

        # (1 unchanged block hidden)
    }
```

this is a big blocker for us we use the LB in production 

i tried to actually ignore the property `type` but that doesn't really work also

### Steps to Reproduce

**Terraform Configuration Files**

**Terraform version**

**Debug Output**

**Panic Output**

## Additional context

**Important Factoids**

**References**

danaelhe:
Just so I understand, terraform always thinks there's a change in the `digitalocean_load_balancer` resource because it can't pick up the `type` attribute? So it tries to recreate the resource every time?

romankor:
yes it doesnt import actually a bunch of other properties like `glb_settings` and `domains`

PR #1340, opened 2025-03-17
load balancers: fix importing global LBs

andrewsomething:
This fixes importing global load balancers. We were not setting the GLB related attributes in the read method, so they were not set on import. This required setting the `type` attribute to `Computed`. There are also some related test fix ups as well as a new test for importing GLBs. Finally, I cleaned up some of the error messages (No need to preface them with `[DEBUG]`).

Fixes: #1338


loosla (approved):
:rocket: Looks good to me

PR #1341, opened 2025-03-18
Revert "partner interconnect operations support"

andrewsomething:
Reverts digitalocean/terraform-provider-digitalocean#1329

This temporarily reverts this change pending the incorporation of the feature rename in  https://github.com/digitalocean/godo/pull/802 Releasing this in its current form will just break when the API is updated. We can re-add it once the transition is complete. 

CC: @apinonformoso


loosla (approved):
:+1:

PR #1370, opened 2025-05-12
GLB fix domains certificate id/name on loadbalancers_resources

StephenVarela:
certificate_id is a read only attribute on the GLB domains. As such we should only be setting it on the datasources call to flatten domains as it doesnt exist in the resources schema.

### DIFF
--- a/digitalocean/loadbalancer/resource_loadbalancer.go
+++ b/digitalocean/loadbalancer/resource_loadbalancer.go
@@ -475,6 +475,11 @@ func resourceDigitalOceanLoadBalancerV0() *schema.Resource {
 							ValidateFunc: validation.NoZeroValues,
 							Description:  "name of certificate required for TLS handshaking",
 						},
+						"certificate_id": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "certificate ID for TLS handshaking",
+						},
 						"verification_error_reasons": {
 							Type:        schema.TypeList,
 							Computed:    true,
--- a/docs/resources/loadbalancer.md
+++ b/docs/resources/loadbalancer.md
@@ -161,7 +161,8 @@ the backend service. Default value is `false`.
 
 * `name` - (Required) The domain name to be used for ingressing traffic to a Global Load Balancer.
 * `is_managed` - (Optional) Control flag to specify whether the domain is managed by DigitalOcean.
-* `certificate_id` - (Optional) The certificate ID to be used for TLS handshaking.
+* `certificate_name` - (Optional) The certificate name to be used for TLS handshaking.
+* `certificate_id` - (Read Only) The certificate id associated with the domain used for TLS handshaking.
 
 `glb_settings` supports the following:
 


digitalocean/loadbalancer/resource_loadbalancer_test.go
```diff
@@ -1473,8 +1475,9 @@ resource "digitalocean_loadbalancer" "lorem" {
   }
 
   domains {
-    name       = "test-updated-2.github.io"
-    is_managed = false
+    name             = "test-updated-2.github.io"
+    is_managed       = false
+    certificate_name = "test-cert"
```
andrewsomething, line 1480: There is nothing in the test config creating this certificate causing the test to fail:

```
$ make testacc TESTARGS='-run=TestAccDigitalOceanGlobalLoadbalancer'

# snip

=== PAUSE TestAccDigitalOceanGlobalLoadbalancer
=== CONT  TestAccDigitalOceanGlobalLoadbalancer
    resource_loadbalancer_test.go:858: Step 2/2 error: Error running apply: exit status 1

        Error: certificate not found

          with digitalocean_loadbalancer.lorem,
          on terraform_plugin_test.tf line 34, in resource "digitalocean_loadbalancer" "lorem":
          34: resource "digitalocean_loadbalancer" "lorem" {

--- FAIL: TestAccDigitalOceanGlobalLoadbalancer (193.26s)
```

andrewsomething (changes requested):
I wonder if it wouldn't make sense to just add `certificate_id` to the schema as a `Computed` attribute? If knowing the current ID of the certificate is useful enough to be included in the data source, it's likely worthwhile to include in the resource as well.

StephenVarela:
> I wonder if it wouldn't make sense to just add `certificate_id` to the schema as a `Computed` attribute? If knowing the current ID of the certificate is useful enough to be included in the data source, it's likely worthwhile to include in the resource as well.

Hmm i actually dont know why we have the id being returned in the datasource either, it seems like we are overwriting it with the name anyway (I suspect it was for legacy reasons/not breaking existing clients)

Thoughts on actually removing certificate_id? instead of adding to the datasource?

digitalocean/loadbalancer/resource_loadbalancer_test.go
StephenVarela: Just backed out this change sorry the issue isnt really surfaced here it shouldnt have been added

andrewsomething:
> it seems like we are overwriting it with the name anyway 

:thinking:  hmmm... Good point. It may just have been cargo cult-ed from how we treat certificates in the forwarding rules where we originally used the ID and transitioned to using the name. Though it's probably safer to just keep it around at this point in case it is depended on by someone.

digitalocean/loadbalancer/resource_loadbalancer.go
```diff
@@ -475,6 +475,13 @@ func resourceDigitalOceanLoadBalancerV0() *schema.Resource {
 							ValidateFunc: validation.NoZeroValues,
 							Description:  "name of certificate required for TLS handshaking",
 						},
+						"certificate_id": {
+							Type:         schema.TypeString,
+							Optional:     true,
+							Computed:     true,
+							ValidateFunc: validation.NoZeroValues,
```
andrewsomething, line 482: If we do add this, we can remove `Optional` and `ValidateFunc` as it is read-only.

StephenVarela: done!

andrewsomething (approved):

Issue #1374, opened 2025-05-15
Terraform does not clean up docker credentials on destroy


nicvibe:
# Bug Report

Terraform does not remove created credentials for container registry on destroy.

---

## Describe the bug
Credentials are created, but are not removed on destroy.

### Affected Resource(s)
- digitalocean_container_registry_docker_credentials.

### Expected Behavior
Terraform delete the created access token

### Actual Behavior
Terraform logs says that the credentials were removed, but the access token is still available on the website (https://cloud.digitalocean.com/account/api/tokens)

### Steps to Reproduce
1. `terraform apply`
2. `terraform destroy`

**Terraform Configuration Files**
```terraform
resource "digitalocean_container_registry" "container_registry" {
  name                   = "my-container-registry-u712345"
  subscription_tier_slug = "basic"
  region                 = var.region
}

resource "digitalocean_container_registry_docker_credentials" "container_registry_credentials" {
  registry_name  = "my-container-registry-u712345"
  expiry_seconds = 365 * 24 * 60 * 60

  depends_on = [
    digitalocean_container_registry.container_registry
  ]
}
```

**Terraform version**
Terraform v1.10.4
on darwin_arm64

provider registry.terraform.io/digitalocean/digitalocean v2.50.0

**Debug Output**
```sh
$ terraform destroy -auto-approve
digitalocean_container_registry.container_registry: Refreshing state... [id=my-container-registry-u712345]
digitalocean_container_registry_docker_credentials.container_registry_credentials: Refreshing state... [id=my-container-registry-u712345]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  - destroy

Terraform will perform the following actions:

  # digitalocean_container_registry.container_registry will be destroyed
  - resource "digitalocean_container_registry" "container_registry" {
      - created_at             = "2025-05-15 21:06:49 +0000 UTC" -> null
      - endpoint               = "registry.digitalocean.com/my-container-registry-u712345" -> null
      - id                     = "my-container-registry-u712345" -> null
      - name                   = "my-container-registry-u712345" -> null
      - region                 = "ams3" -> null
      - server_url             = "registry.digitalocean.com" -> null
      - storage_usage_bytes    = 0 -> null
      - subscription_tier_slug = "basic" -> null
    }

  # digitalocean_container_registry_docker_credentials.container_registry_credentials will be destroyed
[...]

PR #1375, opened 2025-05-15
registry: fix credential revocation on destroy

andrewsomething:
The DOCR API changed the way API tokens are stored in the Docker credential response. Previously the token was used for both the username and password (`base64(token:token)`). Now it uses the account's email address as the username  (`base64(email:token)`). The token remains in the password field. 

Unfortunately we were using the one in the username field when we attempt to revoke the token. So this change means the token was left unrevoked. This change updates to using the password field.

It also updates the acceptance test actually validate that the token was revoked.

```
=== RUN   TestAccDigitalOceanContainerRegistryDockerCredentials_Basic
--- PASS: TestAccDigitalOceanContainerRegistryDockerCredentials_Basic (16.46s)
=== RUN   TestAccDigitalOceanContainerRegistryDockerCredentials_withExpiry
--- PASS: TestAccDigitalOceanContainerRegistryDockerCredentials_withExpiry (13.35s)
PASS
ok  	github.com/digitalocean/terraform-provider-digitalocean/digitalocean/registry	29.882s
```

Fixes: https://github.com/digitalocean/terraform-provider-digitalocean/issues/1374

### DIFF
--- a/digitalocean/registry/resource_container_registry_docker_credentials.go
+++ b/digitalocean/registry/resource_container_registry_docker_credentials.go
@@ -124,7 +124,7 @@ func resourceDigitalOceanContainerRegistryDockerCredentialsUpdate(ctx context.Co
 	return nil
 }
 
-type dockerConfig struct {
+type DockerConfig struct {
 	Auths struct {
 		Registry struct {
 			Auth string `json:"auth"`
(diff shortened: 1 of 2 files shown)


danaelhe (approved):
lgtm! Ty!

PR #1377, opened 2025-05-16
Update App docs to mention repository access

dbrian57:
Clarifies that customers need to provide App Platform access to their repository account.


andrewsomething (approved):
:+1:  Good one!

PR #1382, opened 2025-05-29
Add redundancy_zone field to Partner Network Connect

wez470:
Makes PNC redundancy_zone available in terraform.

I tested this locally and was able to set my desired redundancy zone.


harshmaru7 (approved):
LGTM !

PR #1390, opened 2025-06-16
Add app ingress authority match

domenipavec:
This adds authority matching to digitalocean app ingress config allowing routing based on domain.


josef-butler:
This would be incredibly useful!

domenipavec:
@SSharma-10 done

SSharma-10:
the test case is failing because file is not properly formatted. Pease format the file accordingly

domenipavec:
@SSharma-10 Fixed.

No idea how that happened, I have go fmt always run on save.

SSharma-10:
it seems the test case is failing due to an incorrectly formatted configuration file. Could you please try `make terrafmt` command?

domenipavec:
@SSharma-10 done

SSharma-10 (approved):

PR #1421, opened 2025-08-04
add support for new doks cluster options

m3co-code:
`amd_gpu_device_plugin` and `amd_gpu_device_metrics_exporter_plugin` being those options.


digitalocean/kubernetes/datasource_kubernetes_cluster_test.go
```diff
@@ -94,6 +96,14 @@ resource "digitalocean_kubernetes_cluster" "foo" {
   routing_agent {
     enabled = true
   }
+
+  amd_gpu_device_plugin {
+    enabled = true
+  }
+
+  amdGpuDeviceMetricsExporterPlugin {
```
anup-deka, line 104: Is should be `amd_gpu_device_metrics_exporter_plugin` right? 🤔

m3co-code: Oh yes, my bad. Updated.

m3co-code:
Fixed the test, could you restamp?

Issue #1368, opened 2025-05-11
Terraform destroy fails due to VPC in use


nicvibe:
# Bug Report

I'm unable to successfully delete VPCs in one go because there are resources attached to it.

---

## Describe the bug

I'm getting 409 status code when trying to delete a VPC that had resources attached to it. In my case, I created a kubernetes cluster and attached a VPC to it.

My guess is that kubernetes is being deleted asynchronously and is still attached to the VPC. As a result, when terraform tries to delete the VPC, it fails. Because of it, I have to keep retrying the destroy operation until all the attached resources are deleted.

### Affected Resource(s)
- digitalocean_vpc
- digitalocean_kubernetes_cluster

### Expected Behavior
I expect that either the that VPC gets immediately deleted or terraform waits until all dependent resources (in this case kubernetes cluster) get fully deleted before try to delete the VPC.

### Actual Behavior
The deletion happens flawless

### Steps to Reproduce
1. `terraform apply`
2. `terraform destroy`

**Terraform Configuration Files**
```tf
resource "digitalocean_kubernetes_cluster" "kube-cluster" {
  name   = "cluster"
  region = var.region
  version  = "1.32.2-do.1"
  vpc_uuid = digitalocean_vpc.private_network.id

  node_pool {
    name       = "worker-pool"
    size       = "s-2vcpu-2gb"
    node_count = 2
  }

  depends_on = [digitalocean_vpc.private_network]
}
resource "digitalocean_vpc" "private_network" {
  name     = "private-network"
  region   = var.region
  ip_range = "10.10.10.0/24"
}
```

**Terraform version**
Terraform v1.10.4
on darwin_arm64
+ provider registry.terraform.io/digitalocean/digitalocean v2.50.0

**Debug Output**
```sh
digitalocean_kubernetes_cluster.kube-cluster: Destroying... [id=8bdb4c48-3878-4bf9-932b-a3798754fb6a]
digitalocean_kubernetes_cluster.kube-cluster: Destruction complete after 1s
digitalocean_vpc.private_network: Destroying... [id=8f1b4120-3618-4cfc-8d1d-a20d320c9076]
╷
│ Error: Error deleting VPC: DELETE https://api.digitalocean.com/v2/vpcs/8f1b4120-3618-4cfc-8d1d-a20d320c9076: 409 (request "137424ab-243a-48e2-b127-c8851ad935b8") Can not delete VPC with members
│
│

```

[...]

PR #1424, opened 2025-08-11
(bugfix) retry delete VPC if it still used by other resources

AnhCaooo:
Catch 409 Conflict while `terraform destroy` tries to delete VPC. It seems that while detroy, user's VPC still used by other resources


SSharma-10:
Thank you for contributing to digital Ocean Open Source, and for identifying and fixing the issue. We truly appreciate your attention to detail and quick action.

SSharma-10 (approved):

AnhCaooo:
Btw, @SSharma-10 shall we close [this issue](https://github.com/digitalocean/terraform-provider-digitalocean/issues/1368) as this now resolved?

PR #1425, opened 2025-08-11
fix kubernetes cluster autoscaler configuration no-op updates and un-…

d-honeybadger:
…setting of customizations

In TF, updates are a PUT-like operation where removal of a field should generally be interpreted as un-setting that field.
But on the backend, in k8saas api, updates are done via PATCH requests.
I tried to capture that logic for cluster_autoscaler_configuration.expanders before, but didn't quite get it right, there were a couple problems:

1. On every `update`, even if existing cluster had `nil` expanders, and the cluster being applied also had `nil` expanders, it still sent an empty array of expanders. Since the expanders are behind a feature flipper, this led to validation errors for customers that didn't have the feature enabled. In this PR, I'm ensuring that if existing expanders are empty, and the desired expanders are empty (be it nil or `[]`), we send `nil` to the api, i.e. "no update".
2. If a user had expanders configured, and later unset them, which resulted in existing cluster having an empty array of expanders, then on every subsequent TF apply, TF saw a non-empty diff `expanders = [] -> null`

### DIFF
--- a/digitalocean/kubernetes/kubernetes.go
+++ b/digitalocean/kubernetes/kubernetes.go
@@ -442,6 +442,10 @@ func FilterTags(tags []string) []string {
 
 func expandCAConfigOpts(config []interface{}) *godo.KubernetesClusterAutoscalerConfiguration {
 	caConfig := &godo.KubernetesClusterAutoscalerConfiguration{}
+	if len(config) == 0 || config[0] == nil {
+		return caConfig
+	}
+
 	configMap := config[0].(map[string]interface{})
 
 	if v, ok := configMap["scale_down_utilization_threshold"]; ok {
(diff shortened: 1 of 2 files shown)


digitalocean/kubernetes/kubernetes.go
```diff
@@ -442,6 +442,10 @@ func FilterTags(tags []string) []string {
 
 func expandCAConfigOpts(config []interface{}) *godo.KubernetesClusterAutoscalerConfiguration {
 	caConfig := &godo.KubernetesClusterAutoscalerConfiguration{}
+	if len(config) == 0 || config[0] == nil {
```
d-honeybadger, line 445: One more drive-by fix: cluster with `cluster_autoscaler_configuration {}` (empty but non-nil) used to cause a panic before this fix when trying to type-assert `nil` as `configMap := config[0].(map[string]interface{})` in the code below.

SSharma-10 (approved):
LGTM

PR #1472, opened 2025-11-19
DOCR: Multiple registries support

rak16:
- Added support for multi-registry endpoints to existing registry and docker credentials resources
- Added a new registries resource to be used eventually, after we deprecate the existing one


anup-deka (approved):
LGTM

PR #1477, opened 2025-11-25
APP Platform: Fix Intermittent 524 error, Use PendingDeployment ID in WaitForAppDeployment

vhrahulkumar:
This PR addresses intermittent 524 timeout errors during App Platform deployments by optimizing the `WaitForAppDeployment` function to use the `PendingDeployment.ID` when available from the API response.

- Modified `WaitForAppDeployment`: Prioritize `PendingDeployment.ID` for direct polling instead of always listing deployments.
- Fallback Logic: If `PendingDeployment` is not set (e.g., during app creation), fall back to the existing behavior of listing deployments.
- Unit Tests: Added tests for both scenarios (using pending ID and fallback to listing).

### DIFF
--- a/digitalocean/app/resource_app.go
+++ b/digitalocean/app/resource_app.go
@@ -304,38 +304,43 @@ func WaitForAppDeployment(client *godo.Client, id string, timeout time.Duration,
 	n := 0
 
 	ticker := time.NewTicker(time.Duration(tickerInterval) * time.Second)
+	defer ticker.Stop() // Ensure ticker is stopped on exit
+
+	var deploymentID string
+	if pendingDeploymentID != "" {
+		deploymentID = pendingDeploymentID
+	}
+
 	for range ticker.C {
 		if n*tickerInterval > timeoutSeconds {
-			ticker.Stop()
 			break
 		}
 
 		var deployment *godo.Deployment
 		var err error
 
-		if pendingDeploymentID != "" {
-			deployment, _, err = client.Apps.GetDeployment(context.Background(), id, pendingDeploymentID)
+		if deploymentID != "" {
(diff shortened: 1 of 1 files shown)


digitalocean/app/resource_app.go
```diff
@@ -312,24 +330,16 @@ func waitForAppDeployment(client *godo.Client, id string, timeout time.Duration,
...
 			if err != nil {
+				ticker.Stop()
 				return fmt.Errorf("error trying to read app deployment state: %s", err)
 			}
 
-			// We choose the most recent deployment. Note that there is a possibility
-			// that the deployment has not been created yet. If that is true,
-			// we will do the wrong thing here and test the status of a previously
-			// completed deployment and exit. However there is no better way to
-			// correlate a deployment with the request that triggered it.
```
greeshmapill: Let's keep this comment as it is still relevant.

digitalocean/app/resource_app.go
```diff
@@ -288,20 +298,28 @@ func resourceDigitalOceanAppDelete(ctx context.Context, d *schema.ResourceData,
...
+		var deployment *godo.Deployment
+		var err error
+
+		if pendingDeploymentID != "" {
+			deployment, _, err = client.Apps.GetDeployment(context.Background(), id, pendingDeploymentID)
+			if err != nil {
+				ticker.Stop()
+				return fmt.Errorf("error trying to read app deployment state: %s", err)
+			}
+		} else {
```
greeshmapill: I think we still need to check `deployment != nil`, else for every poll it will try to list deployments when pendingDeploymentID is not set.

greeshmapill (approved):

kishlay-singh-DO:
Could you sync your fork and we should be good

PR #1500, opened 2026-01-31
Control plane firewalls are in private preview.

dallinb:
Added a note in the resource documentation that:
- Provides a link to the control plane firewall documentation <https://docs.digitalocean.com/products/kubernetes/how-to/add-control-plane-firewall/>.
- Clarifies that this feature is currently in private preview <https://docs.digitalocean.com/platform/product-lifecycle/>.

A preview of the suggested edit of the documentation can be seen at <https://github.com/cbdq-io/terraform-provider-digitalocean/blob/bugfix/DOKS-CP-FW-PP/docs/resources/kubernetes_cluster.md#argument-reference>.

### DIFF
--- a/digitalocean/app/app_spec.go
+++ b/digitalocean/app/app_spec.go
@@ -1311,6 +1311,26 @@ func appSpecIngressSchema() *schema.Resource {
 					},
 				},
 			},
+			"secure_header": {
+				Type:     schema.TypeList,
+				Optional: true,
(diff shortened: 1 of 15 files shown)


dallinb:
Looks like control plane firewall is now GA <https://docs.digitalocean.com/products/kubernetes/how-to/add-control-plane-firewall/#add-control-plane-firewall-terraform>.

PR #1502, opened 2026-02-05
Update godo to v1.173.0 to fix Workspace.created_by type mismatch

johnandersen777:
Fixes:

```
Error: json: cannot unmarshal string into Go struct field Workspace.agent.workspace.created_by of type uint64
```


SSharma-10:
Thanks John!

SSharma-10 (approved):

PR #1511, opened 2026-03-12
HA default true for 1.36.0 and later

pyadagiri-do:
Updates the `ha` schema default to nil for Kubernetes clusters. So that API based default can apply, if user doesn't provide it explicitly in the config

API based defaults:
Kubernetes ≥ 1.36.0: HA control plane enabled when ha is omitted
Kubernetes < 1.36.0: HA control plane disabled when ha is omitted

e2e Validations: https://docs.google.com/document/d/1MNy8df0iekNwVp5xsGZ1HuaXoWVerDViTVdnBRJeraE/edit?tab=t.yc0jg93jcl8l

### DIFF
--- a/digitalocean/kubernetes/resource_kubernetes_cluster.go
+++ b/digitalocean/kubernetes/resource_kubernetes_cluster.go
@@ -70,7 +70,7 @@ func ResourceDigitalOceanKubernetesCluster() *schema.Resource {
 			"ha": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  false,
+				Default:  true,
 			},
 
 			"registry_integration": {
--- a/digitalocean/kubernetes/resource_kubernetes_cluster.go
+++ b/digitalocean/kubernetes/resource_kubernetes_cluster.go
@@ -32,6 +32,15 @@ const (
 	rdmaSharedDevicePluginField            = "rdma_shared_device_plugin"
 )
 
+// ExpandHAFromConfig returns the HA value for the create request. When ha is not
+// specified in config, returns nil so the API applies its version-dependent default.
+func ExpandHAFromConfig(value interface{}, ok bool) *bool {
+	if !ok {
+		return nil
+	}
+	return godo.PtrTo(value.(bool))
+}
+
 func ResourceDigitalOceanKubernetesCluster() *schema.Resource {
 	return &schema.Resource{
 		CreateContext: resourceDigitalOceanKubernetesClusterCreate,
@@ -70,7 +79,7 @@ func ResourceDigitalOceanKubernetesCluster() *schema.Resource {
 			"ha": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  true,
+				// Default is version-dependent: true for DOKS >= 1.36, false otherwise (set in CustomizeDiff)
 			},
 
 			"registry_integration": {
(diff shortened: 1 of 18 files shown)


digitalocean/kubernetes/resource_kubernetes_cluster.go
```diff
@@ -427,10 +436,11 @@ func resourceDigitalOceanKubernetesClusterCreate(ctx context.Context, d *schema.
 		RegionSlug:   d.Get("region").(string),
 		VersionSlug:  d.Get("version").(string),
 		SurgeUpgrade: d.Get("surge_upgrade").(bool),
-		HA:           d.Get("ha").(bool),
 		Tags:         tag.ExpandTags(d.Get("tags").(*schema.Set).List()),
 		NodePools:    poolCreateRequests,
 	}
+	ha, haOk := d.GetOk("ha")
```
d-honeybadger: Have you tried e2e testing it, locally? (e.g. build the provider, and run `terraform apply` using the locall-built provider)
I had a similar case recently where I wanted Terraform to NOT default a bool to `false` when it's not set by the user, and the only thing that worked for me was the deprecated `GetOkExists` method, whereas `GetOk` was returning "OK" even when the field wasn't present in the config. 
But maybe your case is subtly different? If this actually works that's great and no change is necessary, I just wanted to check.

sjana-do: I agree with this comment, with `GetOk`, both the scenario (when user explicitly passes false, and when user omits the flag) produces `ok = false`. The API will enable HA even when user explicitly sets `ha: false`. Using `GetOkExists` might be a better option

digitalocean/kubernetes/resource_kubernetes_cluster.go
```diff
@@ -70,7 +79,7 @@ func ResourceDigitalOceanKubernetesCluster() *schema.Resource {
 			"ha": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  false,
+				// Default is version-dependent: true for DOKS >= 1.36, false otherwise (set in CustomizeDiff)
```
sjana-do: `ha` should be `Optional: true` + `Computed: true` (and no Default) so Terraform can accept the flag being unset in config while still storing the value returned from the API after create/read. Without `Computed`, omitting ha tends to produce a perpetual diff: config implies the zero value false while state reflects the API default (e.g. true on 1.36+).

digitalocean/kubernetes/resource_kubernetes_cluster.go
sjana-do, line 693: the ha will default to `false` when not passed, defeating the purpose of create flow being defaulted to `true`

sjana-do (changes requested):

digitalocean/kubernetes/resource_kubernetes_cluster.go
pyadagiri-do: so here during cluster update, does  d.Get("ha") come directly from tf config or does it use current state, config, diff etc. If so this should be ok I guess ?
d has both old and new values ?
correct me here

digitalocean/kubernetes/resource_kubernetes_cluster.go
pyadagiri-do: yes, was able to repro this issue now.
ran terraform apply with ha= false in my main.tf and using 1.36 version. ha was omitted in API request body, and API created cluster with ha=true.

digitalocean/kubernetes/resource_kubernetes_cluster.go
pyadagiri-do: ack added Computed: true

pyadagiri-do (commented):
addressed comments

digitalocean/kubernetes/resource_kubernetes_cluster.go
sjana-do: Agreed

sjana-do (approved):
LGTM

PR #1517, opened 2026-04-03
nfs reassign

v-amanjain-afk:


sgupta832 (approved):

digitalocean/nfs/resource_nfs_attachment.go
```diff
@@ -60,22 +60,15 @@ func resourceDigitalOceanNfsAttachmentCreate(ctx context.Context, d *schema.Reso
 		return diag.Errorf("Error retrieving share: %s", err)
 	}
 
-	// If share is attached to a different VPC, detach first
+	// If share is attached to a different VPC, use reassign
 	if len(share.VpcIDs) > 0 && share.VpcIDs[0] != vpcId {
-		log.Printf("[DEBUG] Detaching share (%s) from VPC (%s) before attaching to (%s)", shareId, share.VpcIDs[0], vpcId)
-		_, _, err := client.NfsActions.Detach(context.Background(), shareId, share.VpcIDs[0], region)
+		log.Printf("[DEBUG] Reassigning share (%s) from VPC (%s) to VPC (%s)", shareId, share.VpcIDs[0], vpcId)
+		err := reassignNfs(ctx, client, shareId, region, share.VpcIDs[0], vpcId)
```
sjana-do: Can we also have a retry logic here if the reassign call fails, like we have in the else if block below

sjana-do (changes requested):
Can we also add test cases for the new reassign method added.

digitalocean/nfs/resource_nfs_attachment.go
v-amanjain-afk: Done. Added retry logic

digitalocean/nfs/resource_nfs_attachment.go
```diff
@@ -60,22 +60,25 @@ func resourceDigitalOceanNfsAttachmentCreate(ctx context.Context, d *schema.Reso
...
+		err := retry.RetryContext(ctx, 5*time.Minute, func() *retry.RetryError {
+			log.Printf("[DEBUG] Reassigning share (%s) from VPC (%s) to VPC (%s)", shareId, share.VpcIDs[0], vpcId)
 
-		if err = waitForNfsDetach(ctx, client, shareId, region, share.VpcIDs[0]); err != nil {
-			return diag.Errorf("Error waiting for detach: %s", err)
-		}
-	}
+			err := reassignNfs(ctx, client, shareId, region, share.VpcIDs[0], vpcId)
+			if err != nil {
+				return retry.NonRetryableError(
```
sjana-do: Returning NonRetryableError everytime makes RetryContext wrapper a dead code. It would just come out of the retry loop on first error itself. Can we have a logic to retry (return RetryableError) for transient failures like 409 conflict, or timeout errors, and return the NonRetryableError for permanent failures.

sjana-do (changes requested):

digitalocean/nfs/resource_nfs_attachment.go
v-amanjain-afk: Added new retry logic based on error type

sjana-do (approved):
LGTM

PR #1519, opened 2026-04-09
Reverted softprops/action-gh-release@v6 to v2

sjana-do:


kishlay-singh-DO (approved):

sjana-do:
Reverted version of softprops/action-gh-release to v2 as v6 is not available.
[softprops/action-gh-release](https://github.com/softprops/action-gh-release/releases)

PR #1521, opened 2026-04-14
Update Request and Response schema for Dedicated Inference to support Custom model

divyanshgupta-dodeveloper:
Update Request and Response schema for Dedicated Inference to support Custom model

Jira: https://do-internal.atlassian.net/browse/NFRNC-702


digitalocean/dedicatedinference/resource_dedicated_inference.go
sjana-do, line 385: I see this method is being used in create/update flow, mapping terraform config to godo request payload, Shouldn't the provider_model_id field also passed to the payload?

digitalocean/dedicatedinference/datasource_dedicated_inferences.go
sjana-do, line 61: "provider_model_id" field is missing from the schema here.

sjana-do (changes requested):
Can we please update the docs about this new field.

sjana-do (approved):
LGTM

PR #1530, opened 2026-04-25
DOKS: per-cluster SSO config

d-honeybadger:
A few related changes for DOKS SSO. The project is not available to the public yet, so, breaking changes in the sso-specific flags are okay.

1. DOKS now expects per-cluster SSO configuration (i.e. each cluster gets an SSO client ID and issuer url). Adding the corresponding fields to the schema
2. There's no more reason to differentiate between "enabled" and "required" values not being set by the customer vs being set to "false". This makes it much less awkward to handle for terraform: no more `GetOkExists`

### DIFF
--- a/digitalocean/kubernetes/datasource_kubernetes_cluster.go
+++ b/digitalocean/kubernetes/datasource_kubernetes_cluster.go
@@ -225,24 +225,26 @@ func DataSourceDigitalOceanKubernetesCluster() *schema.Resource {
 
 			"sso": {
 				Type:     schema.TypeList,
+				Computed: true,
 				Optional: true,
+				MaxItems: 1,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"enabled": {
 							Type:     schema.TypeBool,
-							Required: true,
+							Computed: true,
 						},
 						"required": {
 							Type:     schema.TypeBool,
-							Optional: true,
+							Computed: true,
 						},
 						"issuer_url": {
 							Type:     schema.TypeString,
-							Optional: true,
+							Computed: true,
 						},
 						"client_id": {
 							Type:     schema.TypeString,
-							Optional: true,
+							Computed: true,
 						},
 					},
 				},
@@ -271,10 +273,6 @@ func DataSourceDigitalOceanKubernetesCluster() *schema.Resource {
 				Computed: true,
 			},
 
-			"kube_config_type": {
-				Type:     schema.TypeString,
-				Computed: true,
-			},
 			"kube_config": kubernetesConfigSchema(),
 
 			"auto_upgrade": {
--- a/digitalocean/kubernetes/datasource_kubernetes_cluster.go
+++ b/digitalocean/kubernetes/datasource_kubernetes_cluster.go
@@ -230,13 +230,19 @@ func DataSourceDigitalOceanKubernetesCluster() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 						"enabled": {
 							Type:     schema.TypeBool,
-							Optional: true,
-							Computed: true,
+							Required: true,
 						},
 						"required": {
 							Type:     schema.TypeBool,
 							Optional: true,
-							Computed: true,
+						},
+						"issuer_url": {
+							Type:     schema.TypeString,
+							Optional: true,
+						},
+						"client_id": {
+							Type:     schema.TypeString,
+							Optional: true,
 						},
 					},
 				},
@@ -265,6 +271,10 @@ func DataSourceDigitalOceanKubernetesCluster() *schema.Resource {
 				Computed: true,
 			},
 
+			"kube_config_type": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
 			"kube_config": kubernetesConfigSchema(),
 
 			"auto_upgrade": {
--- a/digitalocean/kubernetes/kubernetes.go
+++ b/digitalocean/kubernetes/kubernetes.go
@@ -639,16 +639,18 @@ func flattenCAConfigOpts(opts *godo.KubernetesClusterAutoscalerConfiguration) []
 	return result
 }
 
-func expandSSOOpts(d *schema.ResourceData) *godo.KubernetesClusterSSO {
+func expandSSOOpts(sso []interface{}) *godo.KubernetesClusterSSO {
 	ssoConfig := &godo.KubernetesClusterSSO{}
-	if _, isSet := d.GetOk("sso"); !isSet {
+	if len(sso) == 0 || sso[0] == nil {
 		return ssoConfig
 	}
 
-	ssoConfig.Enabled = d.Get("sso.0.enabled").(bool)
-	ssoConfig.Required = d.Get("sso.0.required").(bool)
-	ssoConfig.IssuerURL = d.Get("sso.0.issuer_url").(string)
-	ssoConfig.ClientID = d.Get("sso.0.client_id").(string)
+	ssoMap := sso[0].(map[string]interface{})
+
+	ssoConfig.Enabled = ssoMap["enabled"].(bool)
+	ssoConfig.Required = ssoMap["required"].(bool)
+	ssoConfig.IssuerURL = ssoMap["issuer_url"].(string)
+	ssoConfig.ClientID = ssoMap["client_id"].(string)
 
 	return ssoConfig
 }
--- a/digitalocean/kubernetes/kubernetes.go
+++ b/digitalocean/kubernetes/kubernetes.go
@@ -640,10 +640,11 @@ func flattenCAConfigOpts(opts *godo.KubernetesClusterAutoscalerConfiguration) []
 }
 
 func expandSSOOpts(sso []interface{}) *godo.KubernetesClusterSSO {
-	ssoConfig := &godo.KubernetesClusterSSO{}
 	if len(sso) == 0 || sso[0] == nil {
-		return ssoConfig
+		return nil
 	}
+
+	ssoConfig := &godo.KubernetesClusterSSO{}
 
 	ssoMap := sso[0].(map[string]interface{})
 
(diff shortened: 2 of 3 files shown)


digitalocean/kubernetes/datasource_kubernetes_cluster.go
```diff
@@ -230,13 +230,19 @@ func DataSourceDigitalOceanKubernetesCluster() *schema.Resource {
 					Schema: map[string]*schema.Schema{
```
DO-rrao, line 232: Can you please confirm that the changes are still not released to the public? 
If this is a breaking change, we would have to announce in the release.

digitalocean/kubernetes/resource_kubernetes_cluster.go
```diff
@@ -200,13 +200,19 @@ func ResourceDigitalOceanKubernetesCluster() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 						"enabled": {
 							Type:     schema.TypeBool,
-							Optional: true,
-							Computed: true,
+							Required: true,
```
DO-rrao, line 205: Missing `MaxItems: 1` on the `sso` block. Every other singleton TypeList in this file (`maintenance_policy`, `controlPlaneFirewallField`, `routingAgentField`, etc.) has this constraint.

Without it, Terraform accepts multiple `sso {}` blocks — confirmed with `terraform plan` — but `expandSSOOpts` only reads `sso.0.*`, so additional blocks are silently discarded.

Same issue exists in the datasource schema.

digitalocean/kubernetes/datasource_kubernetes_cluster.go
```diff
@@ -230,13 +230,19 @@ func DataSourceDigitalOceanKubernetesCluster() *schema.Resource {
...
 						"enabled": {
 							Type:     schema.TypeBool,
-							Optional: true,
-							Computed: true,
+							Required: true,
```
DO-rrao: The datasource `sso` block is marked `Optional: true` with `enabled` as `Required: true` inside. For a read-only datasource, this should be `Computed: true` (matching `node_pool`, `maintenance_policy`, etc.) with all sub-fields also `Computed: true`.

Confirmed via `terraform plan`: the `sso` attribute is completely absent from the datasource plan output (while `Computed` fields like `kube_config_type` correctly show `(known after apply)`). This means users cannot reliably reference `data.digitalocean_kubernetes_cluster.x.sso[0].enabled` at plan time.

DO-rrao (commented):
Review: a few schema/convention issues found and confirmed with `terraform plan`.

digitalocean/kubernetes/datasource_kubernetes_cluster.go
```diff
@@ -265,6 +271,10 @@ func DataSourceDigitalOceanKubernetesCluster() *schema.Resource {
 				Computed: true,
 			},
 
+			"kube_config_type": {
```
DO-rrao: `kube_config_type` is declared as `Computed` but:
1. The godo `KubernetesCluster` struct has no corresponding field
2. `digitaloceanKubernetesClusterRead` never calls `d.Set("kube_config_type", ...)`

This will always resolve to `""`. Should this be removed until the API/godo support is in place, or does it need a `d.Set` call added?

digitalocean/kubernetes/kubernetes.go
```diff
@@ -639,19 +639,16 @@ func flattenCAConfigOpts(opts *godo.KubernetesClusterAutoscalerConfiguration) []
 	return result
 }
 
-func expandSSOOptsIfSet(d *schema.ResourceData) *godo.KubernetesClusterSSO {
+func expandSSOOpts(d *schema.ResourceData) *godo.KubernetesClusterSSO {
```
DO-rrao: Nit: convention. Every other expand function in this file takes `raw []interface{}` and leaves the `d.Get(...)` to the caller (e.g. `expandControlPlaneFirewallOpts`, `expandRoutingAgentOpts`). `expandSSOOpts` takes `*schema.ResourceData` instead. Not a correctness issue, but an inconsistency.

digitalocean/kubernetes/datasource_kubernetes_cluster.go
d-honeybadger: Apologies, this change shouldn't have even been left in the PR (I started adding support for multiple kubeconfig types but then realized sso kubeconfig, which requires interactive login, shouldn't be used with terraform at all)

digitalocean/kubernetes/datasource_kubernetes_cluster.go
d-honeybadger: Yes, SSO for DOKS is still behind a flipper and not available to customers.

digitalocean/kubernetes/datasource_kubernetes_cluster.go
d-honeybadger: I see, so, Computed: true only for the datasource whereas the kubernetes resource schema used to creating/updating does not get Computed: true, right?

I did have to keep Optional:true though for `known after apply` to show up and also for MaxItems: 1 setting to be allowed (so it's both Computed AND Optional now). I think it looks good during `plan` now. Tested enabling, disabling, omitting the whole section

digitalocean/kubernetes/kubernetes.go
d-honeybadger: Yeah I wasn't a huge fan of using empty interfaces as it seems nil and type issues are easy to miss. But I can see your point about consistency. Updated to use `[]interface{}` instead.

digitalocean/kubernetes/resource_kubernetes_cluster.go
d-honeybadger: Added, thank you!

digitalocean/kubernetes/kubernetes.go
```diff
@@ -639,19 +639,41 @@ func flattenCAConfigOpts(opts *godo.KubernetesClusterAutoscalerConfiguration) []
 	return result
 }
 
-func expandSSOOptsIfSet(d *schema.ResourceData) *godo.KubernetesClusterSSO {
-	if _, isSet := d.GetOk("sso"); !isSet {
-		return nil
+func expandSSOOpts(sso []interface{}) *godo.KubernetesClusterSSO {
+	ssoConfig := &godo.KubernetesClusterSSO{}
+	if len(sso) == 0 || sso[0] == nil {
+		return ssoConfig
```
DO-rrao: **Bug: `expandSSOOpts` sends SSO payload for every cluster, breaking non-SSO accounts**

When no `sso` block is in the config, `expandSSOOpts` returns `&godo.KubernetesClusterSSO{}` (non-nil zero struct). Because the parent field is `*KubernetesClusterSSO` with `json:"sso,omitempty"`, a non-nil pointer is **not** omitted — the API receives:

```json
{"sso": {"enabled": false, "required": false}}
```

This breaks cluster create/update for **every account without the SSO flipper**:

```
POST .../kubernetes/clusters: 422 validation error: SSO is not available for this account
```

Confirmed by building the provider from this branch and running `terraform apply` with a cluster that has no `sso` block.

**Fix:** return `nil` instead of the zero struct:

```go
func expandSSOOpts(sso []interface{}) *godo.KubernetesClusterSSO {
	if len(sso) == 0 || sso[0] == nil {
		return nil
	}

	ssoConfig := &godo.KubernetesClusterSSO{}
	ssoMap := sso[0].(map[string]interface{})
	// ...
```

Note: this differs from `expandControlPlaneFirewallOpts` and similar functions which return a non-nil zero struct. The difference is that the API validates the presence of `sso` against the feature flag, whereas `control_plane_firewall` with `{enabled: false}` is benign. Each expand function's empty-return value needs to match what the API expects.

digitalocean/kubernetes/kubernetes.go
```diff
@@ -639,19 +639,41 @@ func flattenCAConfigOpts(opts *godo.KubernetesClusterAutoscalerConfiguration) []
...
-		ssoConfig.Enabled = v.(bool)
-	}
-	if v, isSet := d.GetOkExists("sso.0.required"); isSet {
-		ssoConfig.Required = v.(bool)
-	}
+	ssoConfig.Enabled = ssoMap["enabled"].(bool)
+	ssoConfig.Required = ssoMap["required"].(bool)
+	ssoConfig.IssuerURL = ssoMap["issuer_url"].(string)
+	ssoConfig.ClientID = ssoMap["client_id"].(string)
+
```
DO-rrao, line 655: **`issuer_url` and `client_id` cannot be cleared once set (godo `omitempty`)**

The godo struct has:

```go
type KubernetesClusterSSO struct {
	Enabled   bool   `json:"enabled"`
	Required  bool   `json:"required"`
	IssuerURL string `json:"issuer_url,omitempty"`
	ClientID  string `json:"client_id,omitempty"`
}
```

When a user removes `issuer_url` from their config (or sets it to `""`), `expandSSOOpts` produces `IssuerURL: ""`. Due to `omitempty`, the empty string is dropped from the JSON body — the API never receives a signal to clear the field. The old value persists silently.

This requires a godo change (e.g., `*string` or a separate "clear" mechanism) to fix properly. Worth documenting as a known limitation or addressing in godo before GA.

d-honeybadger: I think this works just fine. When DOKS http API receives an SSO config, it acts as though it's a PUT-like operation on the sso config (rather than PATCH-ike). Meaning, if it receives json like this
```
{
  "enabled": false,
  "required": false
}
```
That will get parsed to a go struct (in k8saasapi) with `IssuerURL` and `ClientID` being empty strings.
Then when the actual update happens in the database, these empty strings are persisted, so, issuerURL and clientID do get unset.

digitalocean/kubernetes/kubernetes.go
d-honeybadger: Gah thank you!!! Fixed as you suggested, tested a create and update being not on the flipper

DO-rrao (approved):
LGTM
Thanks for addressing the coments

PR #1534, opened 2026-05-12
APICLI-4364 - loadbalancer, certificate: wait for GLB cert binding + retry cert delete on rotation

sjana-do:
https://do-internal.atlassian.net/browse/APICLI-3455
Terraform apply was failing on Global Load Balancer custom-certificate rotation: Terraform tried to delete the old certificate before the LB control plane had finished swapping the binding to the new one, so the delete returned 403 "certificate in use" and the apply aborted.
* digitalocean_loadbalancer: after Create (and after Update when domains change) poll LoadBalancers.Get until each custom GLB domain reports the configured certificate_name. Short-circuits for REGIONAL load balancers, managed domains, and domains without a certificate_name so non-GLB paths are unaffected.
* digitalocean_certificate: add explicit Create (10m) and Delete (15m) timeouts and a retry loop with exponential backoff (2s -> 30s) that retries only the "certificate in use" 403 from the API; all other errors fail fast. Honors the user-supplied timeouts block and propagates ctx so terraform apply cancellation is respected. Adds unit tests for the error classifier, timeout defaults, the desired-binding extractor, and the no-op short-circuit paths. Acceptance test for the end-to-end rotation will be added in a follow-up so it does not run on every CI build.
Docs updated for both resources.

### DIFF
--- a/examples/loadbalancer/glb_certificate_rotation_two_phase/README.md
+++ b/examples/loadbalancer/glb_certificate_rotation_two_phase/README.md
@@ -36,4 +36,4 @@ Or use the value from `terraform output attach_this_certificate_name_to_glb_doma
 
 ## If you start from a single certificate resource
 
-Use `terraform state mv` to split one resource into `old` / `new` addresses, or add a second resource and import — pair with support on the first migration if needed. See `docs/design/glb-certificate-terraform-rotation.md` section **Immediate safe process**.
+Use `terraform state mv` to split one resource into `old` / `new` addresses, or add a second resource and import — pair with support on the first migration if needed.
--- a/examples/loadbalancer/glb_certificate_rotation_two_phase/README.md
+++ b/examples/loadbalancer/glb_certificate_rotation_two_phase/README.md
@@ -1,0 +1,39 @@
+# Global Load Balancer: two-phase custom certificate rotation
+
+This example shows the **safe immediate process** for rotating a **custom** TLS certificate on a **Global** load balancer when a single `terraform apply` can hit API errors deleting the old certificate while it is still referenced.
+
+It is a **structural template** only: you must supply real PEM material and merge the `domains` change into your **existing** Global Load Balancer configuration.
+
+## Prerequisites
+
+- A working **`digitalocean_loadbalancer`** with `type = "GLOBAL"` and `domains` using a custom certificate today.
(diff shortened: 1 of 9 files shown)


examples/loadbalancer/glb_certificate_rotation_two_phase/main.tf
```diff
@@ -0,0 +1,90 @@
...
+#   then terraform apply (no destroy of old cert).
+#
+# Phase 2: remove the digitalocean_certificate.old block from this file (and
+#   its variables); terraform apply again to delete the old cert only.
+
+terraform {
+  required_providers {
+    digitalocean = {
+      source  = "digitalocean/digitalocean"
+      version = ">= 2.50.0"
```
harshmaru7: just want to confirm 2.50.0 is the correct minimum version that will include this feature ?

examples/loadbalancer/glb_certificate_rotation_two_phase/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Phase 2 — delete old cert only
+
+1. Remove the `digitalocean_certificate.old` resource block (and any references) from your configuration.
+2. Run `terraform plan` — you should see **destroy** for the old cert only.
+3. Run `terraform apply`.
+
+## If you start from a single certificate resource
+
+Use `terraform state mv` to split one resource into `old` / `new` addresses, or add a second resource and import — pair with support on the first migration if needed. See `docs/design/glb-certificate-terraform-rotation.md` section **Immediate safe process**.
```
harshmaru7: minor nit - can we update this See `docs/design/glb-certificate-terraform-rotation.md` section, references an incorrect file

examples/loadbalancer/glb_certificate_rotation_two_phase/main.tf
sjana-do: Thanks, updated the version to 2.86.0

examples/loadbalancer/glb_certificate_rotation_two_phase/README.md
sjana-do: removed the reference, thanks

harshmaru7 (approved):

PR #1538, opened 2026-05-19
Add GradientAI custom model resource and data

venkatranabothu:


digitalocean/gradientai/resource_custom_model.go
```diff
@@ -0,0 +1,423 @@
...
+						"hf_token": {
+							Type:        schema.TypeString,
+							Optional:    true,
+							Sensitive:   true,
+							Description: "HuggingFace token used to access ACCESS_TYPE_PRIVATE or ACCESS_TYPE_GATED repositories. Write-only.",
+						},
+					},
+				},
+			},
+			"preferred_gpu_region": {
```
anup-deka, line 122: Will this ensure the model to be created in regions we provide? I don't see models regions changing when I provide NYC2

venkatranabothu: This is just a preference. We currently support `tor1` and `atl1` only and will honour the preference if it falls within those.

anup-deka (approved):

PR #1544, opened 2026-05-28
add worker_subnet_uuid field to k8s cluster resource and datasource

d-honeybadger:
New field is available in the DOKS API.

I modeled it after `vpc_uuid` field since it behaves in the same ways:
- can only be set in a create request but not in an update request
- is always present in the cluster response, even if not set (in which case the default worker subnet uuid will be used)

### DIFF
--- a/digitalocean/kubernetes/datasource_kubernetes_cluster.go
+++ b/digitalocean/kubernetes/datasource_kubernetes_cluster.go
@@ -156,6 +156,11 @@ func DataSourceDigitalOceanKubernetesCluster() *schema.Resource {
 				Computed: true,
 			},
 
+			"worker_subnet_uuid": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
 			"cluster_subnet": {
 				Type:     schema.TypeString,
 				Computed: true,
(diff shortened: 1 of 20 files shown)


digitalocean/kubernetes/datasource_kubernetes_cluster.go
```diff
@@ -156,6 +156,11 @@ func DataSourceDigitalOceanKubernetesCluster() *schema.Resource {
 				Computed: true,
 			},
```
DO-rrao, line 158: Hey Dasha, had the following comments - 
1. Add docs for worker_subnet_uuid in docs/resources/kubernetes_cluster.md (Argument Reference) and docs/data-sources/kubernetes_cluster.md (Attributes Reference). This is required for parity with every comparable field.
2. (Optional) Beef up the acceptance test with a case that sets worker_subnet_uuid explicitly so the create branch is exercised.

d-honeybadger: 1. Added, thank you! I failed to add other fields to this doc in the past, so, drive-by added them in this PR.
2. If that's okay, I'll skip the extra test. Given each TestAcc creates a separate cluster, it seems an overkill for a simple field setting. Also, there's no way (yet?) to create a VPC subnet using terraform, so the test would have to deal with creating a subnet via godo.

DO-rrao (approved):

Issue #1557, opened 2026-06-23
BYOIP reserved IPs 404 on the action-status poll while DO-allocated ones succeed


waw-saas:
# Bug Report

## TL;DR

**`digitalocean_reserved_ip_assignment` fails with a 404 on the action-status poll for reserved IPs that come from a BYOIP prefix — but works correctly for DO-allocated reserved IPs.** The assign/unassign actually completes on DigitalOcean's side; only the provider's status poll 404s, leaving Terraform state out of sync. A DO-allocated reserved IP run side-by-side with identical config never hits this.

## Describe the bug

`digitalocean_reserved_ip_assignment` fails on **assign / unassign** with a 404 while polling the reserved-IP **action status**, even though the assign/unassign action actually **completes successfully** on DigitalOcean's side.

**The failure is specific to BYOIP-range reserved IPs.** A reserved IP allocated by DigitalOcean behaves correctly; a reserved IP from a customer **BYOIP prefix** consistently 404s on the action-status poll. We isolated this by running both cases in the same config (see Case A vs Case B below): Case A (DO-allocated) succeeds, Case B (BYOIP) fails — same Terraform code, same operation, only the address origin differs.

The provider creates the assign (or unassign) action, gets back an `action_id`, and immediately starts polling:

```
GET /v2/reserved_ips/{ip}/actions/{action_id}
```

That `GET` returns `404 ... The resource you requested could not be found` because the action id is not yet queryable (eventual consistency on the actions endpoint). The provider treats the 404 as a hard failure and aborts:

```
Error: Error waiting for reserved IP (192.0.2.10) to be Assigned:
Error retrieving reserved IP (192.0.2.10) ActionId (3000000000):
GET https://api.digitalocean.com/v2/reserved_ips/192.0.2.10/actions/3000000000: 404 (request "…")
The resource you requested could not be found.

  with digitalocean_reserved_ip_assignment.byoip
```

The damaging part: the assignment **did happen** on DO (the IP is attached to the new droplet), but because the apply errored, the `digitalocean_reserved_ip_assignment` resource is **missing from Terraform state**. A subsequent `terraform apply` then tries to assign *again* and 404s *again*, so the config never converges without manual `terraform import` recovery.

[...]

PR #1558, opened 2026-06-24
Added retry on 404 for get actions

imaskm:
Fixes the Issue [1557](https://github.com/digitalocean/terraform-provider-digitalocean/issues/1557)

### DIFF
--- a/digitalocean/reservedip/reserved_ip_actions.go
+++ b/digitalocean/reservedip/reserved_ip_actions.go
@@ -1,0 +1,67 @@
+package reservedip
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"net/http"
+
+	"github.com/digitalocean/godo"
+	"github.com/digitalocean/terraform-provider-digitalocean/digitalocean/config"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func isReservedIPActionNotFound(resp *godo.Response, err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if resp != nil && resp.StatusCode == http.StatusNotFound {
+		return true
+	}
+
+	if godoErr, ok := err.(*godo.ErrorResponse); ok && godoErr.Response != nil && godoErr.Response.StatusCode == http.StatusNotFound {
+		return true
+	}
+
+	return false
+}
+
+func newReservedIPActionStateRefreshFunc(client *godo.Client, ipAddress string, actionID int) retry.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		action, resp, err := client.ReservedIPActions.Get(context.Background(), ipAddress, actionID)
+		if isReservedIPActionNotFound(resp, err) {
+			// The assign/unassign call returns an action ID immediately, but the
+			// action may not be queryable yet (notably for BYOIP-prefix reserved
+			// IPs). Treat 404 as a transient condition and keep polling.
+			log.Printf("[DEBUG] Reserved IP (%s) action (%d) not found (404), will retry", ipAddress, actionID)
+			return nil, "", nil
+		}
+		if err != nil {
+			return nil, "", fmt.Errorf("Error retrieving reserved IP (%s) ActionId (%d): %s", ipAddress, actionID, err)
+		}
+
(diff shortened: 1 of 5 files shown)


digitalocean/reservedip/reserved_ip_actions_test.go
```diff
@@ -0,0 +1,62 @@
...
+			name: "non-404 error response",
+			err: &godo.ErrorResponse{
+				Response: &http.Response{StatusCode: http.StatusForbidden},
+			},
+			want: false,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
arkahood: This repo we have go 1.23.0 so no need to do this!

digitalocean/reservedip/reserved_ip_actions.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func isReservedIPActionNotFound(resp *godo.Response, err error) bool {
+	if err == nil {
+		return false
+	}
+
+	if resp != nil && resp.StatusCode == http.StatusNotFound {
```
arkahood: we need to add `resp.Response != nil` as well I believe otherwise resp.StatusCode could panic.

arkahood (approved):

arkahood (approved):

digitalocean/reservedip/reserved_ip_actions.go
```diff
@@ -0,0 +1,99 @@
...
+func newReservedIPActionStateRefreshFunc(client *godo.Client, ipAddress string, actionID int, op reservedIPActionOperation, dropletID int) retry.StateRefreshFunc {
+	return func() (interface{}, string, error) {
+		action, resp, err := client.ReservedIPActions.Get(context.Background(), ipAddress, actionID)
+		if isReservedIPActionNotFound(resp, err) {
+			// The assign/unassign call returns an action ID immediately, but the
+			// action may not be queryable yet (notably for BYOIP-prefix reserved
+			// IPs). Some BYOIP actions never appear on the actions endpoint even
+			// after the assignment completes, so fall back to reading the IP.
+			reservedIP, _, ipErr := client.ReservedIPs.Get(context.Background(), ipAddress)
+			if ipErr == nil && reservedIPActionComplete(reservedIP, op, dropletID) {
```
DO-rrao: Possible nil dereference: reservedIPActionComplete dereferences reservedIP.Droplet, but we only guard on ipErr == nil here, not on reservedIP != nil. godo normally returns a non-nil IP when err == nil, so this is defensive, but worth hardening:

digitalocean/reservedip/reserved_ip_actions.go
```diff
@@ -0,0 +1,99 @@
...
+		if err != nil {
+			return nil, "", fmt.Errorf("Error retrieving reserved IP (%s) ActionId (%d): %s", ipAddress, actionID, err)
+		}
+
+		log.Printf("[INFO] The reserved IP Action Status is %s", action.Status)
+		return &action, action.Status, nil
+	}
+}
+
+func newReservedIPAssignmentActionStateRefreshFunc(d *schema.ResourceData, meta interface{}, actionID int, op reservedIPActionOperation) retry.StateRefreshFunc {
```
DO-rrao, line 75: newReservedIPAssignmentActionStateRefreshFunc always reads droplet_id, while newReservedIPResourceActionStateRefreshFunc (below) only reads it when op == reservedIPActionAssign. The unassign path here doesn't use dropletID, so it's harmless today, but the divergence is a trap for future readers. Suggest matching the resource version's op == reservedIPActionAssign guard, or adding a short comment explaining why the difference is fine.

digitalocean/reservedip/reserved_ip_actions.go
```diff
@@ -0,0 +1,99 @@
+package reservedip
```
DO-rrao, line 1: I am not able to recreate the issue on my side as i dont have a BYOIP reserved IP. 
Can you please confirm that we have tested the bug? Thanks

imaskm: I don't have it either but customer has tested the change and it's working fine for them

DO-rrao (approved):

PR #1560, opened 2026-06-25
nfs access point

v-amanjain-afk:


digitalocean/nfs/access_point_client.go
```diff
@@ -0,0 +1,126 @@
...
+	return root.AccessPoint, resp, nil
+}
+
+func listNfsAccessPoints(ctx context.Context, client *godo.Client, shareID string, status string, vpcID string) ([]*nfsAccessPoint, *godo.Response, error) {
+	path := fmt.Sprintf("v2/nfs/shares/%s/access_points", shareID)
+	values := url.Values{}
+	if status != "" {
+		values.Set("status", status)
+	}
+	if vpcID != "" {
```
rak16: This isn't needed 

Ref - https://docs.digitalocean.com/reference/api/reference/nfs/#nfs_list_access_points

rak16 (changes requested):

digitalocean/nfs/access_point_client.go
v-amanjain-afk: done

rak16 (approved):

digitalocean/nfs/access_point_client.go
```diff
@@ -0,0 +1,119 @@
+package nfs
```
DO-rrao: Pls resolve merge conflicts

v-amanjain-afk: done

digitalocean/nfs/access_point_client.go
```diff
@@ -0,0 +1,119 @@
...
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+
+	"github.com/digitalocean/godo"
+)
+
+type nfsAccessPointPolicy struct {
```
DO-rrao: Do we need this whole file? It duplicates what godo already vendors:
- `client.Nfs.CreateAccessPoint`, `GetAccessPoint`, `ListAccessPoints`, `DeleteAccessPoint`
- the `nfsAccessPoint` / `nfsAccessPointPolicy` types and root structs

Nothing else in the provider builds requests by hand like this. Suggest deleting the file and calling godo directly. If the acc tests need a helper, keep a thin `GetNfsAccessPoint` that just delegates to `client.Nfs.GetAccessPoint`.

digitalocean/nfs/access_point_client.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil, resp, err
+	}
+
+	return root.AccessPoint, resp, nil
+}
+
+func listNfsAccessPoints(ctx context.Context, client *godo.Client, shareID string, vpcID string) ([]*nfsAccessPoint, *godo.Response, error) {
+	path := fmt.Sprintf("v2/nfs/shares/%s/access_points", shareID)
+	if vpcID != "" {
+		path = path + "?" + url.Values{"vpc_id": {vpcID}}.Encode()
```
DO-rrao: Is `vpc_id` actually a supported list filter?
- godo's `NfsListAccessPointsOptions` only models `status`
- the data source already filters by VPC client-side

So this looks either ignored or redundant — drop it or confirm the endpoint supports it.

digitalocean/nfs/resource_nfs_access_point.go
```diff
@@ -0,0 +1,214 @@
...
+
+	shareID := d.Get("share_id").(string)
+	req := &nfsCreateAccessPointRequest{
+		Name:         d.Get("name").(string),
+		Path:         d.Get("path").(string),
+		VpcID:        d.Get("vpc_id").(string),
+		AccessPolicy: expandNfsAccessPointPolicy(d.Get("access_policy")),
+	}
+
+	accessPoint, _, err := createNfsAccessPoint(ctx, client, shareID, req)
```
DO-rrao: Create doesn't wait for the AP to become active, so apply returns with `status = ACCESS_POINT_CREATING`. The response includes an action we can wait on. Using godo:

```go
resp, _, err := client.Nfs.CreateAccessPoint(ctx, shareID, req)
if err != nil {
    return diag.Errorf("Error creating NFS access point: %s", err)
}
d.SetId(resp.AccessPoint.ID)
if resp.Action != nil {
    if err := util.WaitForAction(client, resp.Action); err != nil {
        return diag.FromErr(err)
    }
}
```

(Or a `retry.StateChangeConf` polling `ACCESS_POINT_CREATING` -> `ACCESS_POINT_ACTIVE` with the usual 15m/5s/15s timeouts.) Note the acc test only passes today because the separate `IsActive` check polls.

digitalocean/nfs/resource_nfs_access_point.go
```diff
@@ -0,0 +1,214 @@
...
+	d.Set("vpc_id", accessPoint.VpcID)
+	d.Set("access_policy", flattenNfsAccessPointPolicy(accessPoint.AccessPolicy))
+
+	return nil
+}
+
+func resourceDigitalOceanNfsAccessPointDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*config.CombinedConfig).GodoClient()
+
+	_, _, err := deleteNfsAccessPoint(ctx, client, d.Id())
```
DO-rrao: Delete is async (returns an action; status goes to `ACCESS_POINT_DELETED`) but we `SetId("")` immediately without waiting. Other resources wait until the object is gone (`Target: http.StatusText(http.StatusNotFound)`). Please wait on the returned action / poll until deleted, otherwise destroy can race and CheckDestroy gets flaky.

digitalocean/nfs/resource_nfs_access_point.go
```diff
@@ -0,0 +1,214 @@
...
+	d.SetId(accessPoint.ID)
+	return resourceDigitalOceanNfsAccessPointRead(ctx, d, meta)
+}
+
+func resourceDigitalOceanNfsAccessPointRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*config.CombinedConfig).GodoClient()
+
+	accessPoint, resp, err := getNfsAccessPoint(ctx, client, d.Id())
+	if err != nil {
+		if resp != nil && resp.StatusCode == 404 {
```
DO-rrao, line 145: Since delete is a soft-delete, a GET can return the object with `status: ACCESS_POINT_DELETED` instead of a 404, which would leave a dead AP in state. Suggest treating that status the same as not-found and clearing the ID (the attachment resource does something similar with `share.Status != "ACTIVE"`).

digitalocean/nfs/datasource_nfs_access_point.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+func DataSourceDigitalOceanNfsAccessPoint() *schema.Resource {
+	return &schema.Resource{
+		ReadContext: dataSourceDigitalOceanNfsAccessPointRead,
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:          schema.TypeString,
+				Optional:      true,
+				ConflictsWith: []string{"name", "share_id"},
```
DO-rrao: Minor: the runtime check at the bottom handles it, but it'd be clearer to express at the schema level that `name` and `share_id` go together (and both conflict with `id`).

DO-rrao (commented):
Thanks for adding NFS access points. The schemas match the API payloads and the unit tests pass. A few things to resolve before merge, mostly around following the provider's conventions for async actions and reusing the godo client we already vendor.

Blocking:
- Use the vendored godo client instead of the hand-rolled `access_point_client.go`. godo already exposes `client.Nfs.CreateAccessPoint/GetAccessPoint/ListAccessPoints/DeleteAccessPoint`, and every other resource uses the godo client directly.
- Create and Delete don't wait for the action to finish. The API returns `action: IN_PROGRESS` and `status: ACCESS_POINT_CREATING`/`ACCESS_POINT_DELETED`, but we return immediately, so apply finishes with the AP still creating. Convention is `retry.StateChangeConf` (see vpcnatgateway, loadbalancer, vpcpeering) or `util.WaitForAction(client, action)` for action-returning endpoints.

Non-blocking:
- Read only handles a 404; soft-deleted APs can come back as `ACCESS_POINT_DELETED` and should also be cleared from state.
- The `vpc_id` list query param doesn't look like a real filter (godo only models `status`) and we already filter by VPC client-side.
- Missing CHANGELOG entry for the new resource + data source.
- The diff seems to include unrelated vendor churn (secrets.go, gradientai, database_vector) — probably merge noise, but worth confirming the base branch.

digitalocean/nfs/access_point_client.go
v-amanjain-afk: removed this file

digitalocean/nfs/access_point_client.go
v-amanjain-afk: removed this file

digitalocean/nfs/resource_nfs_access_point.go
v-amanjain-afk: done

digitalocean/nfs/resource_nfs_access_point.go
v-amanjain-afk: done

digitalocean/nfs/resource_nfs_access_point.go
v-amanjain-afk: done

digitalocean/nfs/datasource_nfs_access_point.go
v-amanjain-afk: done

DO-rrao (approved):
lgtm

PR #1561, opened 2026-06-26
Support subnet_uuid in vpcnatgateway

johannaratliff:
subnet uuid in vpcnatgw after godo updated to 1.98.0 to support.


digitalocean/vpcnatgateway/resource_vpc_nat_gateway.go
```diff
@@ -68,6 +68,12 @@ func ResourceDigitalOceanVPCNATGateway() *schema.Resource {
 							Description: "ID of the ingress VPC",
 							ForceNew:    true,
 						},
+						"subnet_uuid": {
+							Type:        schema.TypeString,
+							Optional:    true,
```
SSharma-10, line 73: here resource schema likely needs `Computed: true`

SSharma-10:
also there is no test that actually sets subnet_uuid. Can you please add it ?

SSharma-10 (approved):
LGTM

PR #1563, opened 2026-07-06
fix: prevent panic in kafka topic create when API response omits the topic

ab0utbla-k:
`digitalocean_database_kafka_topic` creation panics with a nil pointer dereference when the API creates the topic but responds without a topic payload:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation]
github.com/digitalocean/terraform-provider-digitalocean/digitalocean/database.resourceDigitalOceanDatabaseKafkaTopicCreate
    digitalocean/database/resource_database_kafka_topic.go:318
```

## Root cause

Kafka topic creation is asynchronous server-side. When the topic is not provisioned in time, the API responds `202 Accepted` with an acknowledgment body instead of the documented `201` with the topic object. Captured from a live cluster (the create handler blocked for ~20s first, `do-upstream-service-time: 20745`):

```
HTTP/2 202
{"message":"topic cannot be performed. Reason: topic is still provisioning","id":"accepted","request_id":"..."}
```

The topic *is* created and reaches `active` shortly after. But godo decodes this body into a `databaseTopicRoot` with a nil `Topic` and no error, and the create function dereferences `topic.Name` unguarded. Because the panic kills the plugin before the resource ID is recorded, the topic never lands in state - every retry crashes again and orphans another topic. Reproduced consistently (2/2 CI applies + probes) on a real cluster: Kafka 3.8, `db-s-2vcpu-2gb` x3, region `fra1`, provider v2.86.0 (the create path is unchanged through v2.93.0/main). `GetTopic` returns the same payload-less body while the topic is provisioning.

This is the create-path sibling of #1553, which fixed the same class of panic in the read path and noted the API "commonly returns ... partially populated configs (e.g. immediately after topic creation)".

## Changes

- Create derives the resource ID from the request (`opts.Name`), which is known client-side, instead of trusting the response payload.
- Create then waits (`retry.RetryContext`, same pattern as `resource_database_replica`) until `GetTopic` actually returns the topic, so the first apply converges instead of tainting the resource while the topic provisions.
- Read guards against a nil topic from `GetTopic` and returns an error instead of panicking on `topic.State`.
- Regression test replaying the captured API behavior (202 acknowledgment on create, payload-less body on first read): it panics with the exact production stack trace before this change and passes with it.

## Validation

Built the patched provider and ran it via `dev_overrides` against the live cluster exhibiting the behavior: create succeeded in a single apply (`Creation complete`, ID and `state = "active"` in state, follow-up plan shows `No changes`), followed by a clean destroy. Unit suite for `digitalocean/database` passes.

### DIFF
--- a/digitalocean/database/resource_database_kafka_topic.go
+++ b/digitalocean/database/resource_database_kafka_topic.go
@@ -12,6 +12,7 @@ import (
 	"github.com/digitalocean/terraform-provider-digitalocean/digitalocean/config"
 	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/retry"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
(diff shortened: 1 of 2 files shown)


ab0utbla-k:
Fixes #1566

viktorrokhlov-source (approved):

ktrufanov-skycop:
Still reproducing on `v2.95.0` today, twice in a single CI run — two different Kafka clusters (one staging, one production), same topic name, same unguarded dereference:

```
Error: Plugin did not respond
  with digitalocean_database_kafka_topic.app["<app>.<topic>"]

panic: runtime error: invalid memory address or nil pointer dereference
github.com/digitalocean/terraform-provider-digitalocean/digitalocean/database.resourceDigitalOceanDatabaseKafkaTopicCreate
    digitalocean/database/resource_database_kafka_topic.go:318
Error: The terraform-provider-digitalocean_v2.95.0 plugin crashed!
```

This is the fourth time we have hit it on the same code path.

Worth adding to the case for merging: the crash is not just a failed apply. The create is accepted server-side before the panic, so the topic really is created — we confirmed it over the Kafka wire protocol, with the requested partition count and replication factor, all replicas in ISR — but the id never reaches state. The result is a permanent desync that gets worse on its own:

1. The topic exists on the cluster, absent from state.
2. Every later plan re-plans it as a *create*.
3. That apply fails with `422 AlreadyExists`.
4. Because the plan is all-creates, it reads as non-destructive, so any automation that gates only on destructive changes runs it unattended — and now every unrelated change to that environment fails until a human hand-writes an `import` block.

So one transient `202` costs an out-of-band adoption PR and blocks an entire environment's pipeline in the meantime. Upgrading is not an escape either: neither this PR nor #1577 is in `v2.99.0`, so every release so far carries the same dereference.

Either fix resolves it for us — this one or #1577. Happy to test a build of whichever branch a maintainer prefers.

SSharma-10 (approved):
Thanks for the PR! LGTM

PR #1565, opened 2026-07-08
Check droplet events before assign reserved_ip

imaskm:
When assigning a reserved_ip/byoip, we need to wait on droplet to finish any events which might be pending


digitalocean/reservedip/reserved_ip_actions.go
```diff
@@ -19,6 +20,52 @@ const (
...
+			if err != nil {
+				return nil, "", fmt.Errorf("Error retrieving droplet (%d) actions: %s", dropletID, err)
+			}
+
+			if dropletHasPendingEvent(actions) {
+				log.Printf("[DEBUG] Droplet (%d) has pending events, waiting before reserved IP assign", dropletID)
+				return nil, "waiting", nil
+			}
+
+			return nil, "ready", nil
```
anup-deka:
```suggestion
				return actions, "waiting", nil
			}

			return actions, "ready", nil
```

anup-deka (approved):
LGTM

PR #1567, opened 2026-07-10
added byoip support for load balancer and nat-gateway

rbhatia-code:


DO-rrao:
**Highest-risk item — LB `ip` is sent on every update**

`buildLoadBalancerRequest` is shared by both create and update:

```go
// digitalocean/loadbalancer/resource_loadbalancer.go
if v, ok := d.GetOk("ip"); ok {
    opts.IP = v.(string)
}
```

On update, `ip` is populated in state (computed), so `GetOk` returns it and `opts.IP` is included in the update payload. godo documents `IP` as "Create-only; not used on update." If the API silently ignores it, this is harmless; if it validates/rejects a create-only field, this could break **all** LB updates, not just BYOIP ones.

Since the field is already `ForceNew` (any real change recreates), I'd scope the assignment to creation only to be safe — e.g. guard with `d.IsNewResource()`, or set `opts.IP` in the create func rather than the shared builder. Worth confirming the API's update behavior before merge.

DO-rrao:
**Data source schema is now inconsistent with the shared elem**

`datasource_vpc_nat_gateway.go` still declares `egresses` / `public_gateways` as plain `Computed`, but it reuses `egressPublicGatewaysSchemaResource()`, which this PR changed to `Optional + Computed + ForceNew`.

For a data source, `Optional` / `ForceNew` are meaningless (harmless, not a bug), but the shared helper now carries resource-only semantics into the datasource. Not blocking — just flagging the coupling in case future edits assume the helper is datasource-safe.

digitalocean/vpcnatgateway/resource_vpc_nat_gateway.go
```diff
@@ -90,13 +90,19 @@ func ResourceDigitalOceanVPCNATGateway() *schema.Resource {
 			},
 			"egresses": {
 				Type:        schema.TypeList,
+				Optional:    true,
 				Computed:    true,
-				Description: "List of egresses",
+				ForceNew:    true,
+				MaxItems:    1,
```
DO-rrao, line 95: ### 1 — `MaxItems: 1` on `egresses` will break refresh if the API returns more than one gateway

This `MaxItems: 1` is inconsistent with how state is built. `flattenEgresses` loops over `egresses.PublicGateways` and appends **one `egresses` list entry per gateway**:

```go
for _, egress := range egresses.PublicGateways {
    ...
    result = append(result, map[string]interface{}{"public_gateways": gatewaySet})
}
```

`godo.Egresses.PublicGateways` is a slice, so if the API ever returns 2+ gateways, `d.Set("egresses", ...)` produces a list longer than 1 and violates `MaxItems: 1` → Terraform errors with "Provider produced inconsistent result" on read/import.

The intended model is a single `egresses` block whose `public_gateways` set holds all gateways. Please either:
- **(preferred)** rework `flattenEgresses` to emit one egress entry and add every gateway to that entry's `public_gateways` set, or
- drop `MaxItems: 1` from `egresses`.

digitalocean/vpcnatgateway/resource_vpc_nat_gateway.go
```diff
@@ -90,13 +90,19 @@ func ResourceDigitalOceanVPCNATGateway() *schema.Resource {
...
+				Description: "Optional egress configuration. Set public_gateways.ipv4 to assign a BYOIP / reserved IP on create. When omitted, a system-allocated reserved IP is provisioned.",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"public_gateways": {
 							Type:        schema.TypeSet,
+							Optional:    true,
 							Computed:    true,
+							ForceNew:    true,
+							MaxItems:    1,
```
DO-rrao: ### 2 — `ForceNew` on the `Optional+Computed` nested set can trigger an unintended full NAT gateway replacement

`Optional + Computed + ForceNew` on a nested `TypeSet` is risky, because a `TypeSet` element's identity is its hash, and the hash here is derived from `ipv4`. That creates two ways to accidentally destroy and recreate the gateway:

1. **Empty block to request a system IP** — a user writes `public_gateways {}` (no `ipv4`). `expandEgresses` correctly skips it and the API allocates an IP, but on the next plan the config hash (empty `ipv4`) differs from the state hash (allocated `ipv4`). On a `ForceNew` block, that diff proposes a **destroy + recreate** of the whole NAT gateway.
2. **API normalizes the IP** — the BYOIP round-trip only stays diff-free if the API echoes back the exact same `ipv4` string; any reformatting flips the hash and triggers replacement.

Could you add an acceptance test asserting an empty plan in both cases: (a) create with a BYOIP `ipv4`, and (b) create with no `egresses` block at all? If either shows a diff, we'll need a `DiffSuppressFunc`/`CustomizeDiff`, or to reconsider `ForceNew` on the nested blocks.

DO-rrao (commented):
A few things to address before merge, mostly around the NAT gateway `egresses` schema and docs. Two inline comments below (issues 1 & 2). The third item can't be an inline comment because the docs files aren't part of this PR's diff, so I'm including it here:

### 3 — Docs still describe these fields as read-only; they're now inputs

These fields are now user-configurable inputs, but the docs still present them as read-only outputs. Please update:

- **`docs/resources/loadbalancer.md`** — `ip` is only listed under attributes today. Document it as an **optional, create-only (ForceNew)** BYOIP input: must be an unassigned BYOIP on the account in the same region; when omitted, a floating IP is auto-allocated. Also note BYOIP is IPv4-only (`ipv6` remains computed).
- **`docs/resources/vpc_nat_gateway.md`** — move `egresses` / `public_gateways.ipv4` into the arguments section as **optional, create-only** BYOIP inputs, and note that omitting them yields a system-allocated reserved IP.

A CHANGELOG entry for the new feature would be good to include as well.

DO-rrao (approved):
lgtm

PR #1570, opened 2026-07-16
MDROP-22:add terraform support for microdroplets

kwadhwa-source:


sgupta832:
Closing as superseded. MicroDroplet terraform support was merged in #1578, then reverted in #1584 after public registry docs were discovered. We're restoring onto the api-v2 contract in a new PR (resources only, docs withheld until PP). See MDROP-336 / MDROP-22.

PR #1572, opened 2026-07-20
Support subnet_uuid in loadbalancers

johannaratliff:
Support subnet_uuid in loadbalancers


digitalocean/loadbalancer/loadbalancer_unit_test.go
```diff
@@ -186,6 +186,36 @@ func TestWaitForGLBDomainCertificatesIfNeeded_NoOpPaths(t *testing.T) {
 // TestWaitForGLBDomainCertificateBindings_EmptyDesired makes sure the
 // underlying poller exits immediately when there is nothing to wait for,
 // without touching the godo client.
+func TestBuildLoadBalancerRequest_SubnetUUID(t *testing.T) {
```
DO-rrao, line 198: Small thing: this new test landed right in the middle of the comment and the function it belongs to. The `// TestWaitForGLBDomainCertificateBindings_EmptyDesired makes sure...` comment now sits above `TestBuildLoadBalancerRequest_SubnetUUID`, so it reads like it's describing the wrong test. Mind moving the new test just below the `EmptyDesired` function (or above the comment) so they line back up?

digitalocean/loadbalancer/resource_loadbalancer.go
```diff
@@ -386,6 +386,14 @@ func resourceDigitalOceanLoadBalancerV0() *schema.Resource {
 				ValidateFunc: validation.NoZeroValues,
 			},
 
+			"subnet_uuid": {
```
DO-rrao, line 389: The docs say a subnet requires `vpc_uuid` to be set, but we don't actually check that here — if someone sets `subnet_uuid` on its own, we'll just forward it and let the API reject it. Adding `RequiredWith: []string{"vpc_uuid"}` would catch it at plan time and keep the behavior in sync with what the docs promise. Not a blocker, just a nicer error for the user.

digitalocean/loadbalancer/resource_loadbalancer_test.go
```diff
@@ -61,6 +61,8 @@ func TestAccDigitalOceanLoadbalancer_Basic(t *testing.T) {
 						"digitalocean_loadbalancer.foobar", "droplet_ids.#", "1"),
 					resource.TestCheckResourceAttrSet(
 						"digitalocean_loadbalancer.foobar", "vpc_uuid"),
+					resource.TestCheckResourceAttr(
+						"digitalocean_loadbalancer.foobar", "subnet_uuid", ""),
```
DO-rrao, line 66: These only assert `subnet_uuid` is empty, so we're really just testing the "not set" case. The actual feature — creating an LB with a real subnet and having it come back correctly — isn't covered by any acceptance test; right now it only lives in the `buildLoadBalancerRequest` unit test, which doesn't hit the API. Could we add a case that sets `subnet_uuid` inside a VPC and checks it round-trips? That's the part I'd most want green before merging.

digitalocean/loadbalancer/datasource_loadbalancer_test.go
```diff
@@ -137,6 +139,8 @@ data "digitalocean_loadbalancer" "foobar" {
 						"data.digitalocean_loadbalancer.foobar", "urn", expectedURNRegEx),
```
DO-rrao, line 139: feature code looks good — schema, buildLoadBalancerRequest, read handlers, and docs are all correct, and the unit test passes locally. Two issues on the acceptance test:

1. t.Setenv will panic (blocking). resource.ParallelTest calls t.Parallel(), and t.Setenv panics in parallel tests ("cannot set environment variables in parallel tests"). So step 2's PreConfig will panic once step 1 succeeds. Suggest dropping the TF_VAR indirection and interpolating accTestSubnetUUID directly into the step-2 config via a closure, since it's known after step 1.

2. Cleanup ordering. The subnet is deleted in t.Cleanup, which runs after Terraform destroys the VPC. If DO won't delete a VPC that still has a subnet, CheckDestroy fails; if it cascades, the cleanup just 404s (harmless). Can you confirm? If the former, delete the subnet before the TF destroy.

Optional: subnet_uuid isn't in godo (hence the raw HTTP to /v2/vpcs/{id}/subnets) — worth confirming that path and eventually adding godo support so the test can wire the subnet in-graph. If the feature is allowlist-gated, a note on whether the test can run in CI would help.

anup-deka (approved):
LGTM

PR #1576, opened 2026-07-24
Add AMD GPU partition mode support to Droplet and sizes

DO-rrao:
## Summary

Bumps godo to **v1.201.0** and surfaces the new AMD GPU partition mode feature through the provider.

- **`digitalocean_droplet` (resource):** new optional, `ForceNew` `gpu_partition_mode` argument, validated against the supported enum values (`PARTITION_MODE_SPX_NPS1`, `PARTITION_MODE_DPX_NPS2`). Omit it to create a full GPU. The API only echoes the value back on **create** (not on read), so a refresh preserves the configured value rather than clearing it.
- **`digitalocean_sizes` (data source):** new computed `gpu_info` block exposing `count`, `model`, `vram { amount, unit }`, and `supported_partition_modes` for discovery. An empty `supported_partition_modes` means partition-mode selection is unavailable for that size/account.
- **`digitalocean_droplet` (data source):** exposes `gpu_partition_mode`. It is currently empty because read-back is not yet returned by the API.

Server remains authoritative: the provider does no gating and only validates the enum string; unsupported combinations are rejected by the API.

## Test plan

- [x] `go build ./...`
- [x] `go vet`, `gofmt`, lint clean
- [x] New unit tests pass:
  - `digitalocean/size`: `flattenGPUInfo` / `flattenDigitalOceanSize` incl. the absent-field → empty-list case
  - `digitalocean/droplet`: create-request wiring (`expandDropletCreateRequest`), attribute set with the no-read-back preservation guard, enum validation, and data-source flatten
- [x] godo `v1.201.0` checksum verified against the public transparency log (`sum.golang.org`)
- [x] Manually verified against a live account: `gpu-mi350x8-2304gb` returns `["PARTITION_MODE_DPX_NPS2","PARTITION_MODE_SPX_NPS1"]`; a non-partitionable size returns an empty list

## Notes

- Read-back of `gpu_partition_mode` on droplet GET is not delivered yet; the field is added now for forward compatibility and the resource preserves the configured value on refresh.
- Vendor churn includes unrelated godo changes (microdroplets, MySQL config, etc.) pulled in by the 1.197 → 1.201 bump.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 Changelog moved to Release Notes in [Github Releases](https://github.com/digitalocean/terraform-provider-digitalocean/releases)
 
+## Unreleased
+
+FEATURES:
+
+- MDROP-22: add `digitalocean_microdroplet` resource
+- MDROP-22: add `digitalocean_microdroplet_image` resource
+- MDROP-22: add `digitalocean_microdroplet` data source
+- MDROP-22: add `digitalocean_microdroplets` data source
+- MDROP-22: add `digitalocean_microdroplet_image` data source
+- MDROP-22: add `digitalocean_microdroplet_images` data source
+- MDROP-22: add `digitalocean_microdroplet_checkpoints` data source
+
 ## 2.48.2
 
 BUGFIXES:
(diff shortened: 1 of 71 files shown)


SSharma-10 (approved):
LGTM

PR #1578, opened 2026-07-27
MDROP-22: add terraform support for microdroplets

kwadhwa-source:
Add resources and data sources for DigitalOcean MicroDroplets and MicroDroplet images, plus a data source for reading MicroDroplet checkpoints.

New:
- digitalocean_microdroplet (resource)
- digitalocean_microdroplet_image (resource)
- digitalocean_microdroplet (data source)
- digitalocean_microdroplets (data source, plural)
- digitalocean_microdroplet_image (data source)
- digitalocean_microdroplet_images (data source, plural)
- digitalocean_microdroplet_checkpoints (data source, plural; read-only, since checkpoints are captured automatically by DO when a MicroDroplet pauses)

State transitions on the microdroplet resource use the dedicated pause / resume action endpoints exposed in godo v1.200.0, rather than a PATCH on the resource, matching the doctl implementation.

Bumps github.com/digitalocean/godo v1.163.0 -> v1.200.0 (required: the microdroplet client and MicroDropletCheckpoint type were introduced in v1.200.0). Follow-on fixes in unrelated packages caused by the bump:

- genai: client.GenAI -> client.GradientAI (godo renamed the field)
- app: AppIngressSpecRuleStringMatch.{Prefix,Exact} are now *string
- kubernetes: KubernetesClusterCreateRequest.HA is now *bool

### DIFF
--- a/digitalocean/microdroplet/datasource_microdroplet.go
+++ b/digitalocean/microdroplet/datasource_microdroplet.go
@@ -1,0 +1,91 @@
+package microdroplet
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/digitalocean/godo"
+	"github.com/digitalocean/terraform-provider-digitalocean/digitalocean/config"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+// DataSourceDigitalOceanMicroDroplet returns a data source that looks up a
+// MicroDroplet by `id` or by `name`.
+func DataSourceDigitalOceanMicroDroplet() *schema.Resource {
+	recordSchema := microDropletDataSourceSchema()
+	recordSchema["id"] = &schema.Schema{
+		Type:         schema.TypeString,
+		Optional:     true,
+		Computed:     true,
+		Description:  "MicroDroplet ID",
+		ValidateFunc: validation.NoZeroValues,
+		ExactlyOneOf: []string{"id", "name"},
+	}
+	recordSchema["name"] = &schema.Schema{
+		Type:         schema.TypeString,
+		Optional:     true,
+		Computed:     true,
+		Description:  "MicroDroplet name",
+		ValidateFunc: validation.NoZeroValues,
+		ExactlyOneOf: []string{"id", "name"},
+	}
+
+	return &schema.Resource{
+		ReadContext: dataSourceDigitalOceanMicroDropletRead,
+		Schema:      recordSchema,
+	}
+}
+
+func dataSourceDigitalOceanMicroDropletRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	client := meta.(*config.CombinedConfig).GodoClient()
+
+	var found *godo.MicroDroplet
+	if id, ok := d.GetOk("id"); ok {
+		m, _, err := client.MicroDroplets.Get(ctx, id.(string))
+		if err != nil {
+			return diag.Errorf("Error retrieving MicroDroplet: %s", err)
+		}
+		found = m
+	} else if name, ok := d.GetOk("name"); ok {
+		matches, err := listMicroDropletsByName(ctx, client, name.(string))
+		if err != nil {
+			return diag.Errorf("Error listing MicroDroplets: %s", err)
+		}
+		switch len(matches) {
+		case 0:
+			return diag.Errorf("no MicroDroplet found with name %s", name.(string))
+		case 1:
+			found = &matches[0]
+		default:
+			return diag.Errorf("too many MicroDroplets found with name %s (found %d, expected 1)", name.(string), len(matches))
+		}
+	}
+
+	if err := setMicroDropletAttributes(d, found); err != nil {
+		return diag.FromErr(err)
+	}
+	return nil
+}
+
+func listMicroDropletsByName(ctx context.Context, client *godo.Client, name string) ([]godo.MicroDroplet, error) {
+	opts := &godo.ListOptions{Page: 1, PerPage: 200}
+	var out []godo.MicroDroplet
+	for {
+		batch, resp, err := client.MicroDroplets.ListByName(ctx, name, opts)
+		if err != nil {
+			return nil, err
+		}
+		out = append(out, batch...)
+		if resp == nil || resp.Links == nil || resp.Links.IsLastPage() {
+			break
+		}
+		page, err := resp.Links.CurrentPage()
+		if err != nil {
+			return nil, fmt.Errorf("error paging MicroDroplets: %w", err)
+		}
+		opts.Page = page + 1
+	}
+	return out, nil
+}
--- a/digitalocean/microdroplet/microdroplet.go
+++ b/digitalocean/microdroplet/microdroplet.go
@@ -1,0 +1,494 @@
+package microdroplet
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/digitalocean/godo"
+	"github.com/digitalocean/terraform-provider-digitalocean/digitalocean/config"
+	"github.com/digitalocean/terraform-provider-digitalocean/digitalocean/tag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+// stateValues is the closed set accepted by the settable `state` attribute on
+// digitalocean_microdroplet. The API also exposes transient values (creating,
+// pausing, resuming, ...) but users should never set those directly.
+var stateValues = []string{
+	string(godo.MicroDropletStateRunning),
+	string(godo.MicroDropletStatePaused),
+}
+
+// networkingValues enumerates the accepted values for the `networking`
+// attribute on digitalocean_microdroplet.
+var networkingValues = []string{
+	string(godo.MicroDropletNetworkingPublic),
+	string(godo.MicroDropletNetworkingVPC),
+}
+
+// httpProtocolValues enumerates the accepted values for the `http_protocol`
+// attribute on digitalocean_microdroplet. The control plane accepts only
+// `http` (HTTP/1.1) and `http2`; `https` is not a valid platform value.
+var httpProtocolValues = []string{
+	string(godo.MicroDropletHTTPProtocolHTTP),
+	string(godo.MicroDropletHTTPProtocolHTTP2),
+}
+
+// tagsSchemaForceNew returns tag.TagsSchema() with ForceNew set. Tags are
+// accepted by MicroDropletCreateRequest but the MicroDroplets API exposes no
+// endpoint to mutate them afterwards, so any change has to recreate the
+// resource. Marking ForceNew keeps Terraform's plan honest — without it,
+// changes would silently no-op on apply.
+func tagsSchemaForceNew() *schema.Schema {
+	s := tag.TagsSchema()
+	s.ForceNew = true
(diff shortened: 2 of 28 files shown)


digitalocean/microdroplet/microdroplet.go
```diff
@@ -0,0 +1,474 @@
...
+var networkingValues = []string{
+	string(godo.MicroDropletNetworkingPublic),
+	string(godo.MicroDropletNetworkingVPC),
+}
+
+// httpProtocolValues enumerates the accepted values for the `http_protocol`
+// attribute on digitalocean_microdroplet.
+var httpProtocolValues = []string{
+	string(godo.MicroDropletHTTPProtocolHTTP),
+	string(godo.MicroDropletHTTPProtocolHTTPS),
```
apinonformoso: HTTPS isn't a supported protocol. The control plane accepts exactly `http1` and `http2`, mapping onto the platform enum in the data path.

digitalocean/microdroplet/microdroplet.go
```diff
@@ -0,0 +1,474 @@
...
+			Optional:     true,
+			ForceNew:     true,
+			Description:  "Port the MicroDroplet exposes over HTTP",
+			ValidateFunc: validation.IntBetween(1, 65535),
+		},
+		"http_protocol": {
+			Type:         schema.TypeString,
+			Optional:     true,
+			ForceNew:     true,
+			Description:  "HTTP protocol: 'http', 'https', or 'http2'",
```
apinonformoso:
```suggestion
			Description:  "HTTP protocol: 'http', or 'http2'",
```

docs/resources/microdroplet.md
```diff
@@ -0,0 +1,86 @@
...
+The following arguments are supported:
+
+* `name` - (Required) Name of the MicroDroplet. Forces recreation on change.
+* `region` - (Required) DigitalOcean region slug for the MicroDroplet's location. Forces recreation on change.
+* `size` - (Required) MicroDroplet size slug. Forces recreation on change.
+* `image` - (Required) MicroDroplet image reference (UUID or URN). Forces recreation on change.
+* `networking` - (Optional) Networking mode: `public` (default) or `vpc`. Forces recreation on change.
+* `vpc_uuid` - (Optional) UUID of a VPC to attach to. Only valid when `networking = "vpc"`. Forces recreation on change.
+* `http_port` - (Optional) HTTP port to expose. Forces recreation on change.
+* `http_protocol` - (Optional) HTTP protocol: `http`, `https`, or `http2`. Forces recreation on change.
```
apinonformoso:
```suggestion
* `http_protocol` - (Optional) HTTP protocol: `http`, or `http2`. Forces recreation on change.
```

docs/data-sources/microdroplet.md
```diff
@@ -0,0 +1,33 @@
+---
+page_title: "DigitalOcean: digitalocean_microdroplet"
+subcategory: "MicroDroplets"
+---
+
+# digitalocean_microdroplet
+
+Fetches a single [DigitalOcean MicroDroplet](https://docs.digitalocean.com/products/microdroplets/) by `id` or by `name`.
```
apinonformoso, line 8: This URL doesn't exists validate with PDOCs team, what will be the documentation URL

apinonformoso (approved):

digitalocean/microdroplet/resource_microdroplet.go
```diff
@@ -0,0 +1,218 @@
...
+
+	if d.HasChange("state") {
+		target := godo.MicroDropletState(d.Get("state").(string))
+		if err := transitionMicroDropletState(ctx, client, d.Id(), target, d.Timeout(schema.TimeoutUpdate)); err != nil {
+			return diag.FromErr(err)
+		}
+	}
+
+	return resourceDigitalOceanMicroDropletRead(ctx, d, meta)
+}
```
DO-rrao, line 146: Update silently drops auto_pause / auto_resume / tags changes

Update only reacts to state, but auto_pause, auto_resume, and tags are not ForceNew in microDropletResourceSchema(), and godo's MicroDropletsService exposes no update endpoint for them. Net effect: changing any of these produces a clean plan and a "1 changed" apply that makes no API call, then Read overwrites state — silent drift with no error.

Concrete impact: a user lowering auto_pause.idle_timeout from 30m to 5m to cut cost sees a successful apply, but the platform keeps the old timeout and keeps billing the extra idle time. Same for disabling auto_resume on a webhook-backed instance (keeps waking/billing).

Please either mark auto_pause/auto_resume as ForceNew (honest, since the API can't mutate them in place), or implement real update handling. For tags, the provider already has tag.SetTags — please wire that in on update rather than no-op'ing.

digitalocean/microdroplet/resource_microdroplet.go
```diff
@@ -0,0 +1,218 @@
...
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(10 * time.Minute),
+			Update: schema.DefaultTimeout(5 * time.Minute),
+			Delete: schema.DefaultTimeout(5 * time.Minute),
+		},
+		Schema: microDropletResourceSchema(),
+	}
```
DO-rrao, line 35: If we go the ForceNew route for issue in Update, auto_pause (and auto_resume) should declare it explicitly:

```
"auto_pause": {
    Type:        schema.TypeList,
    Optional:    true,
    Computed:    true,
    ForceNew:    true,   // API has no in-place update path
    MaxItems:    1,
    ...
}
```
Without this, changes are neither recreated nor applied (see Update comment).

digitalocean/microdroplet/resource_microdroplet_test.go
```diff
@@ -0,0 +1,298 @@
...
+					testAccCheckMicroDropletExists(resourceName),
+					resource.TestCheckResourceAttr(resourceName, "current_state", string(godo.MicroDropletStateRunning)),
+				),
+			},
+		},
+	})
+}
+
+func TestAccDigitalOceanMicroDroplet_UpdateAutoPause(t *testing.T) {
+	name := acceptance.RandomTestName()
```
DO-rrao, line 145: The comment here claims auto_pause is ForceNew "via the wider schema", but it isn't set ForceNew in microDropletResourceSchema(). As written, step 2 hits the no-op Update path, Read reconciles from the API (which never received the change), and the auto_pause.0.enabled == "true" check fails — or the step fails with a non-empty plan.

This means the test is either red against a real account, or falsely green against a mock that echoes config back (which would hide the Update bug). Please align the test with whatever we decide for the Update/ForceNew fix, and drop/rewrite the misleading comment.

digitalocean/microdroplet/datasource_microdroplet.go
```diff
@@ -0,0 +1,91 @@
...
+			return diag.Errorf("no MicroDroplet found with name %s", name.(string))
+		case 1:
+			found = &matches[0]
+		default:
+			return diag.Errorf("too many MicroDroplets found with name %s (found %d, expected 1)", name.(string), len(matches))
+		}
+	}
+
+	if err := setMicroDropletAttributes(d, found); err != nil {
+		return diag.FromErr(err)
```
DO-rrao, line 67: tags never round-trip. godo.MicroDroplet has no Tags field and this function doesn't set tags, while tag.TagsSchema() is Optional (not Computed). So tags are effectively create-only: they can't be verified against the API, won't populate on import, and (combined with the Update issue) changes are silently dropped.

Real impact: an org adds tags = ["team:payments","env:prod"] for cost allocation, apply "succeeds," but the instance stays untagged in DO — their billing/ownership reports silently exclude it.

Also note flattenMicroDroplet omits tags, so data.digitalocean_microdroplets.*.tags is always empty and a filter { key = "tags" } returns nothing. Please reconcile tags on read (and in the flatten map), or at minimum document the limitation prominently.

digitalocean/microdroplet/resource_microdroplet.go
```diff
@@ -0,0 +1,218 @@
...
+	if _, err := waitForMicroDropletState(ctx, client, id, target, pending, timeout); err != nil {
+		return fmt.Errorf("error waiting for MicroDroplet (%s) to reach state %s: %w", id, target, err)
+	}
+	return nil
+}
+
+// waitForMicroDropletState polls MicroDroplets.Get until the observed state
+// matches target, or the timeout elapses. `pending` should list every state
+// that is legally a step on the way to `target`.
+func waitForMicroDropletState(ctx context.Context, client *godo.Client, id string, target godo.MicroDropletState, pending []godo.MicroDropletState, timeout time.Duration) (interface{}, error) {
```
DO-rrao, line 203: Create waiter can fail on a transient state. 
Pending: [creating], Target: [running] means any other state godo returns — notably unknown, which the API can emit briefly right after POST — is treated as terminal by retry.StateChangeConf:

`unexpected state 'unknown', wanted target 'running'`
This shows up as intermittent create failures (e.g. provisioning many MicroDroplets via for_each, leaving a partially-applied state) and flaky CI that passes on re-run. Suggest adding unknown to the create/transition Pending set:

`[]godo.MicroDropletState{godo.MicroDropletStateUnknown, godo.MicroDropletStateCreating}`
The image waiter in resource_microdroplet_image.go has the same exposure if Get returns an empty/unknown status momentarily — worth guarding there too.

digitalocean/microdroplet/microdroplet.go
```diff
@@ -0,0 +1,494 @@
...
+						Required:    true,
+						Description: "Whether auto-pause is enabled",
+					},
+					"idle_timeout": {
+						Type:         schema.TypeString,
+						Optional:     true,
+						Computed:     true,
+						Description:  "Idle timeout as a Go duration string (e.g. '5m', '30s')",
+						ValidateFunc: validation.NoZeroValues,
+					},
```
DO-rrao, line 140: **`auto_pause` can be added, but never edited or disabled**

`ForceNew` on the `auto_pause` list only fires when the block is added/removed (`auto_pause.#` changes). I verified the plan with `schema.Resource.Diff` against this commit — edits *inside* an existing block are not `ForceNew`, and since `Update` only handles `state`, they silently no-op:

| Change | Plan result |
|---|---|
| `idle_timeout` `5m` → `10m` | in-place diff, `RequiresNew=false` → `Update` ignores → **perpetual diff / no-op** |
| `enabled` `true` → `false` | in-place diff, `RequiresNew=false` → **silent no-op** |
| remove the `auto_pause` block | **`NO DIFF`** — `Optional+Computed` retains the old value, so removal is ignored |

Net effect: once auto-pause is enabled, there is **no working way to change `idle_timeout` or to turn it off** via Terraform — the in-place edit, `enabled = false`, and removing the block are all swallowed. This is a common day-2 operation (tuning/disabling auto-pause), so users will hit a plan that never converges. The `RecreateOnAutoPauseChange` test only covers the `0 → 1` add case, so it doesn't catch this.

Suggested fix (verified this flips the edit cases to `RequiresNew=true`): mark the nested fields `ForceNew`.

```go
"enabled": {
    Type:        schema.TypeBool,
    Required:    true,
    ForceNew:    true,
    Description: "Whether auto-pause is enabled",
},
"idle_timeout": {
    Optional:     true,
    Computed:     true,
    ForceNew:     true,
    Description:  "Idle timeout as a Go duration string (e.g. '5m', '30s')",
},
```

That covers edits and `enabled = false`. Disabling via *block removal* still won't be detected while `auto_pause` is `Computed` — if that path should work too, consider a `CustomizeDiff` (e.g. `customdiff.ForceNewIfChange`) on `auto_pause`, or dropping `Computed`. A test that exercises `enabled true → false` and block `1 → 0` would guard this.

DO-rrao (approved):
lgtm

PR #1579, opened 2026-07-28
Add DOKS DRA plugins and GPU partition mode

thearyanahmed:
- Add `nvidia_gpu_dra_driver` and `amd_gpu_dra_driver` to `digitalocean_kubernetes_cluster` (and datasource).
- Add `gpu_partition_mode` to kubernetes node pools (`AMD_PARTITION_MODE_SPX_NPS1` / `AMD_PARTITION_MODE_DPX_NPS2`, create-only).
- Update docs and acceptance test coverage.
- Depends on godo v1.202.0 (minor after current main, this PR to be precise https://github.com/digitalocean/godo/pull/1068 ). CI should fail before that, which is expected.

### DIFF
--- a/digitalocean/kubernetes/kubernetes.go
+++ b/digitalocean/kubernetes/kubernetes.go
@@ -210,6 +210,7 @@ func expandNodePools(nodePools []interface{}) []*godo.KubernetesNodePool {
 	expandedNodePools := make([]*godo.KubernetesNodePool, 0, len(nodePools))
 	for _, rawPool := range nodePools {
 		pool := rawPool.(map[string]interface{})
+		gpuPartitionMode, _ := pool["gpu_partition_mode"].(string)
 		cr := &godo.KubernetesNodePool{
 			ID:               pool["id"].(string),
 			Name:             pool["name"].(string),
@@ -222,7 +223,7 @@ func expandNodePools(nodePools []interface{}) []*godo.KubernetesNodePool {
 			Labels:           expandLabels(pool["labels"].(map[string]interface{})),
 			Nodes:            expandNodes(pool["nodes"].([]interface{})),
 			Taints:           expandNodePoolTaints(pool["taint"].(*schema.Set).List()),
-			GPUPartitionMode: pool["gpu_partition_mode"].(string),
+			GPUPartitionMode: gpuPartitionMode,
 		}
 
 		expandedNodePools = append(expandedNodePools, cr)
--- a/digitalocean/kubernetes/kubernetes.go
+++ b/digitalocean/kubernetes/kubernetes.go
@@ -97,6 +97,16 @@ func nodePoolSchema(isResource bool) map[string]*schema.Schema {
 			Optional: true,
 			Elem:     nodePoolTaintSchema(),
 		},
+
+		"gpu_partition_mode": {
+			Type:     schema.TypeString,
+			Optional: true,
+			ForceNew: true,
+			ValidateFunc: validation.StringInSlice([]string{
+				godo.KubernetesAMDPartitionModeSPXNPS1,
+				godo.KubernetesAMDPartitionModeDPXNPS2,
+			}, false),
+		},
 	}
 
 	if isResource {
@@ -201,17 +211,18 @@ func expandNodePools(nodePools []interface{}) []*godo.KubernetesNodePool {
 	for _, rawPool := range nodePools {
 		pool := rawPool.(map[string]interface{})
 		cr := &godo.KubernetesNodePool{
-			ID:        pool["id"].(string),
-			Name:      pool["name"].(string),
-			Size:      pool["size"].(string),
-			Count:     pool["node_count"].(int),
-			AutoScale: pool["auto_scale"].(bool),
-			MinNodes:  pool["min_nodes"].(int),
-			MaxNodes:  pool["max_nodes"].(int),
-			Tags:      tag.ExpandTags(pool["tags"].(*schema.Set).List()),
-			Labels:    expandLabels(pool["labels"].(map[string]interface{})),
-			Nodes:     expandNodes(pool["nodes"].([]interface{})),
-			Taints:    expandNodePoolTaints(pool["taint"].(*schema.Set).List()),
+			ID:               pool["id"].(string),
+			Name:             pool["name"].(string),
+			Size:             pool["size"].(string),
+			Count:            pool["node_count"].(int),
+			AutoScale:        pool["auto_scale"].(bool),
+			MinNodes:         pool["min_nodes"].(int),
+			MaxNodes:         pool["max_nodes"].(int),
+			Tags:             tag.ExpandTags(pool["tags"].(*schema.Set).List()),
+			Labels:           expandLabels(pool["labels"].(map[string]interface{})),
+			Nodes:            expandNodes(pool["nodes"].([]interface{})),
+			Taints:           expandNodePoolTaints(pool["taint"].(*schema.Set).List()),
+			GPUPartitionMode: pool["gpu_partition_mode"].(string),
 		}
 
 		expandedNodePools = append(expandedNodePools, cr)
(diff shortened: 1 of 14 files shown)


thearyanahmed:
I could not set labels, do not merge please.

digitalocean/kubernetes/resource_kubernetes_cluster.go
```diff
@@ -379,6 +381,36 @@ func ResourceDigitalOceanKubernetesCluster() *schema.Resource {
 				},
 			},
 
+			nvidiaGpuDraDriverField: {
```
DO-rrao, line 386: These blocks are documented as **"Create-only to enable"**, but they're `Optional`/`Computed` without `ForceNew`, and they're also plumbed into the update path (the `HasChanges` list at L796 and the `KubernetesClusterUpdateRequest` at L811–L812). Net effect: changing `enabled` after create produces an in-place **update** rather than a forced replacement — the opposite of what the docs imply. Compare with `gpu_partition_mode`, which correctly uses `ForceNew: true`.

Please pick one and make it consistent:
- **If truly create-only:** add `ForceNew: true` here and drop the two fields from the update `HasChanges` list and the update request struct.
- **If updates are actually supported:** reword the doc/godo comment ("Create-only to enable") so it doesn't say create-only.

Verified by dumping the runtime schema: `nvidia_gpu_dra_driver` / `amd_gpu_dra_driver` → `ForceNew=false`, while `gpu_partition_mode` → `ForceNew=true`.

digitalocean/kubernetes/resource_kubernetes_cluster.go
```diff
@@ -379,6 +381,36 @@ func ResourceDigitalOceanKubernetesCluster() *schema.Resource {
...
+					Schema: map[string]*schema.Schema{
+						"enabled": {
+							Type:     schema.TypeBool,
+							Required: true,
+						},
+					},
+				},
+			},
+
+			amdGpuDraDriverField: {
```
DO-rrao, line 402: The docs state `nvidia_gpu_dra_driver` is mutually exclusive with `nvidia_gpu_device_plugin` (and the AMD equivalents), but nothing enforces that in the schema — `ConflictsWith` is empty on all four blocks, so a config setting both passes plan-time validation and only fails server-side. Consider adding `ConflictsWith` (e.g. `nvidia_gpu_dra_driver` ⇄ `nvidia_gpu_device_plugin`, `amd_gpu_dra_driver` ⇄ `amd_gpu_device_plugin`) so users get a clear error at plan time.

digitalocean/kubernetes/kubernetes.go
```diff
@@ -97,6 +97,16 @@ func nodePoolSchema(isResource bool) map[string]*schema.Schema {
 			Optional: true,
 			Elem:     nodePoolTaintSchema(),
 		},
+
+		"gpu_partition_mode": {
```
DO-rrao, line 101: The new `gpu_partition_mode` field has no acceptance-test coverage — a search across the package's `*_test.go` files returns no references to it. The DRA drivers each got a dedicated test, so it'd be good to add one for `gpu_partition_mode` too (and ideally a negative case exercising the DRA-vs-device-plugin mutual exclusivity). Note this will need an AMD GPU–capable size, so it may have to be gated/skipped like other size-restricted acc tests.

docs/resources/kubernetes_cluster.md
```diff
@@ -184,6 +184,10 @@ The following arguments are supported:
     - `enabled` - (Required) Boolean flag whether the component should be enabled or not.
 * `nvidia_gpu_device_plugin` - (Optional) Block containing options for the NVIDIA GPU device plugin component. If not specified, the component will be enabled by default for clusters with NVIDIA GPU nodes.
   - `enabled` - (Required) Boolean flag whether the component should be enabled or not.
+* `nvidia_gpu_dra_driver` - (Optional) Block containing options for the NVIDIA GPU DRA driver component. Mutually exclusive with `nvidia_gpu_device_plugin`. Create-only to enable.
+  - `enabled` - (Required) Boolean flag whether the component should be enabled or not.
+* `amd_gpu_dra_driver` - (Optional) Block containing options for the AMD GPU DRA driver component. Mutually exclusive with `amd_gpu_device_plugin`. Create-only to enable.
```
DO-rrao: Nit (pre-existing, but you're editing right above it): the `rdma_shared_device_plugin` line just below is missing its leading `* ` bullet, so it renders differently from the surrounding list items. Cheap to fix while you're here.

go.mod
```diff
@@ -73,3 +73,4 @@ require (
 )
 
 replace git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999
+
```
DO-rrao: Nit: stray trailing blank line added after the `replace` directive — `go mod tidy` will strip it once you re-vendor against the real godo v1.202.0 release.

DO-rrao (commented):
Thanks for this — the wiring is clean and consistent with the existing GPU plugin pattern, and it builds/vets/gofmts fine against the vendored godo. Left a few inline notes. Two are behavioral (the "create-only" DRA blocks are still editable via update, and the documented mutual exclusivity isn't enforced in-schema), plus a test-coverage gap on `gpu_partition_mode`. Non-blocking nits on `go.mod`/docs. The godo v1.202.0 dependency is of course the known "DO NOT MERGE" blocker until it's tagged.

digitalocean/kubernetes/kubernetes.go
thearyanahmed: Good catch. I've added tests `TestAccDigitalOceanKubernetesCluster_GpuPartitionMode` and also for the exclusivity `TestAccDigitalOceanKubernetesCluster_NvidiaGpuDraConflictsWithDevicePlugin` , `TestAccDigitalOceanKubernetesCluster_AmdGpuDraConflictsWithDevicePlugin` .

digitalocean/kubernetes/resource_kubernetes_cluster.go
thearyanahmed: Added `ConflictsWith` for both.

digitalocean/kubernetes/resource_kubernetes_cluster.go
thearyanahmed: Updated the docs to remove the create-only part. As this feature is still in preview, we are aiming to support update paths by the time we go GA.

digitalocean/kubernetes/resource_kubernetes_cluster.go
thearyanahmed: Good catch. And a new learning for me ( `ConflictsWith` ). thanks. 

I've updated the code.

DO-rrao:
**Blocker: standalone `digitalocean_kubernetes_node_pool` create panics on the happy path**

`digitaloceanKubernetesNodePoolCreate` now does an unchecked assertion:

```go
GPUPartitionMode: pool["gpu_partition_mode"].(string),
```

…but its only caller, `resourceDigitalOceanKubernetesNodePoolCreate`, builds `rawPool` as a hand-written map literal that was never updated to include the key:

```go
rawPool := map[string]interface{}{
    "name":       d.Get("name"),
    ...
    "taint":      d.Get("taint"),
    // no "gpu_partition_mode"
}
```

A missing map key returns `nil`, and `nil.(string)` panics. This isn't gated behind the new field — it fires for **any** `digitalocean_kubernetes_node_pool` apply (even configs that never mention `gpu_partition_mode`), before the API call is ever made.

I dry-ran the happy path (normal `name`/`size`/`node_count`, no `gpu_partition_mode`) through the resource's real schema + this exact create path:

```
CURRENT PR (no gpu_partition_mode key) PANICS before any API call:
  interface conversion: interface {} is nil, not string
WITH FIX: gpu_partition_mode = "" (no panic)
```

Note the inline `node_pool { }` block on `digitalocean_kubernetes_cluster` is unaffected (it goes through `expandNodePools`, whose map always has the key), which is why CI is green — the existing acc tests only exercise the inline path.

**Fix:**

```go
rawPool := map[string]interface{}{
    ...
    "taint":              d.Get("taint"),
    "gpu_partition_mode": d.Get("gpu_partition_mode"),
}
```

`d.Get` returns `""` for the unset `TypeString`, so the assertion is safe. Also worth making the helper defensive (`v, _ := pool["gpu_partition_mode"].(string)`) so a missing key can never panic again.

thearyanahmed:
Again, really good find. I made some changes, tests passing locally. Requesting a re-review.

DO-rrao (approved):
lgtm

PR #1585, opened 2026-08-03
doks: add support for isolated workers

varshavaradarajan:
This PR adds a new configuration field to kubernetes cluster create call for isolated workers. It has ForceNew set, which means that a cluster will be recreated upon changing the value for isolated_workers.


anup-deka (approved):
LGTM

PR #1604, opened 2026-09-18
wait for droplet IPv4 to be available

imaskm:
Some resources like domain_record, wait for a droplet to be active currently but sometimes due to replication delay, droplet can be active but NetworkInfo can have empty IP Address which wrongly sets the IP addresses in DNS records, this change fixes that, now we wait for the IPv4 address to be present


digitalocean/droplet/resource_droplet.go
```diff
@@ -450,7 +454,14 @@ func resourceDigitalOceanDropletCreate(ctx context.Context, d *schema.ResourceDa
 		return diag.Errorf("Error waiting for droplet (%s) to become ready: %s", d.Id(), err)
 	}
 
-	// waitForDropletAttribute updates the Droplet's state and calls setDropletAttributes.
+	// Status can become active before networking is populated. Wait for IPv4 so
+	// dependents (e.g. digitalocean_domain / digitalocean_record using
+	// ipv4_address) do not call the DNS API with an empty address.
+	if err := waitForDropletIPv4(ctx, d, meta, expectPublicIPv4); err != nil {
```
SSharma-10: This is the right place to wait, but we should skip this call if the last waitForDropletAttribute GET already returned the expected IPv4. Otherwise, every create will do an extra wait even when the networking info is already available.

Also, if this errors, the droplet ID is already set. `setDropletAttributes` will have stored public_networking=false while the IP was empty and that field is ForceNew, so the next apply could recreate the droplet

digitalocean/droplet/resource_droplet.go
```diff
@@ -919,6 +936,56 @@ func waitForDropletDestroy(ctx context.Context, d *schema.ResourceData, meta int
...
+			}
+
+			if FindIPv4AddrByType(droplet, addrType) != "" {
+				return droplet, "ready", nil
+			}
+
+			return droplet, "waiting", nil
+		},
+		Timeout:        d.Timeout(schema.TimeoutCreate),
+		Delay:          10 * time.Second,
```
SSharma-10: `StateChangeConf.Delay` waits before the first refresh. Create already has a 10s delay while waiting for `status=active`, so this adds another 10s even if the IP is already available.

Can we set `Delay` to 0 (or ~3s) so we check immediately and then poll with `MinTimeout`? The race we care about is a few seconds of empty `NetworkInfo` after `active`, not adding another fixed 10s wait

digitalocean/droplet/resource_droplet.go
```diff
@@ -919,6 +936,56 @@ func waitForDropletDestroy(ctx context.Context, d *schema.ResourceData, meta int
...
+				return nil, "", err
+			}
+
+			if FindIPv4AddrByType(droplet, addrType) != "" {
+				return droplet, "ready", nil
+			}
+
+			return droplet, "waiting", nil
+		},
+		Timeout:        d.Timeout(schema.TimeoutCreate),
```
SSharma-10: This adds a second full create timeout (default 60m) on top of `waitForDropletAttribute`. If the IP never shows up, create could run for almost 2 hours

A shorter dedicated timeout of a few minutes should be enough for replica lag. Failing faster is better than holding the apply for that long

digitalocean/droplet/resource_droplet.go
```diff
@@ -919,6 +936,56 @@ func waitForDropletDestroy(ctx context.Context, d *schema.ResourceData, meta int
...
+			droplet, resp, err := client.Droplets.Get(context.Background(), id)
+			if err != nil {
+				if resp != nil && resp.StatusCode == http.StatusNotFound {
+					// Still provisioning; keep retrying under NotFoundChecks.
+					return nil, "", nil
+				}
+				return nil, "", fmt.Errorf("Error retrieving droplet: %s", err)
+			}
+
+			if err := setDropletAttributes(d, droplet); err != nil {
```
SSharma-10: This is what writes `ipv4_address` into state, which is good, but it also sets `public_networking = (publicIPv4 != "")`. During the empty-IP window, that becomes `false`.

If the waiter times out after that, Terraform can keep the resource because the ID is already set. The next plan then sees `public_networking=false` vs `true`/unset and `ForceNew`-replaces the droplet

Can we avoid persisting that computed `false` while we are still waiting or reset it if the waiter fails?

SSharma-10 (approved):

